PR #9, opened 2019-04-13
fix: sort items in selected list in same order as unselected

jenniferarnesen:
When the toggler calculates which items are added, the items needed to be re-sorted to match the unselected list.


jenniferarnesen:
Superceded by another PR

PR #13, opened 2019-04-19
fix: determine the unselected items based on selected list

jenniferarnesen:
The selected items should not show up in the unselected list


edoardo (approved):
LGTM.
Tested in DV.

PR #14, opened 2019-04-22
chore: remove deploy-build from provider in travis

varl:
Deploy build is not strictly required for libraries which are published from the root, so it is safe to remove it from the Travis deployment pipeline and only use `d2-utils release` to publish the package to NPM.

Mitigates the problem outlined in https://github.com/dhis2/deploy-build/issues/11.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,13 +7,6 @@ deploy:
 - provider: script
   skip_cleanup: true
   script:
-  - npx --package @dhis2/deploy-build deploy-build
-  on:
-    tags: false
-    all_branches: true
-- provider: script
-  skip_cleanup: true
-  script:
   - npx @dhis2/cli-utils release --publish npm
   on:
     tags: false


jenniferarnesen (approved):

jenniferarnesen:
@varl Approved, but I thought the problem is actually in the publishing of d2-ui, which has d2-ui-analytics as a dependency, and it tries to set the version of d2-ui-analytics to 5.3.x (same as d2-ui).

varl:
> 
> 
> @varl Approved, but I thought the problem is actually in the publishing of d2-ui, which has d2-ui-analytics as a dependency, and it tries to set the version of d2-ui-analytics to 5.3.x (same as d2-ui).

Sorry and thanks for pointing that out. I misunderstood what the problem was.

I'll try to make the `deploy-build` script look at its own packages and only update those as per your suggestion.

PR #20, opened 2019-04-25
fix: get period-selector with latest button style fixes

jenniferarnesen:


neeilya (approved):

jenniferarnesen:
Will be superceded by a newer period-selector

PR #25, opened 2019-05-22
fix(layout module): move/add functions, constants and tests from DV app

janhenrikoverland:
Building a set of functions for easy fetching and validation of axes, dimensions and items in the layout object (columns/rows/filters part of an AO).

Also adding test resources that could be reused other places. Will export them as soon as they are needed.


amcgee (commented):
This looks great to me, but should also probably be reviewed by someone with more context.  Would be great to use these in maps @turban

jenniferarnesen (approved):

PR #29, opened 2019-06-04
feat: export orgUnitId functions for detecting and creating org unit level and group ids

jenniferarnesen:
Related to [DHIS2-6990](https://jira.dhis2.org/browse/DHIS2-6990). The orgUnitId functions handle the group and level prefixes.
The reason these functions are in the shared lib are because they are being used by both data-vis and the d2-charts-api.

Changes include:
- run tests on travis (added `yarn test`, everything else in .travis.yml is formatting)
- give consistent names to orgUnitId functions
- export only functions related to the orgUnitId prefix
- BREAKING CHANGE: several functions that were not necessary have been removed, and the name of the export has been changed.


src/index.js
```diff
@@ -35,6 +35,8 @@ export {
     FIXED_DIMENSIONS,
 } from './modules/fixedDimensions'
 
+export { orgUnitId } from './modules/orgUnitId/index'
```
amcgee: The `/index` here is redundant.  Doesn't hurt, but maybe cleaner to omit?

janhenrikoverland (approved):

PR #30, opened 2019-06-20
feat: active/inactive styling for dimension items

edoardo:
The active flag is passed from the app within the dimension items list passed to the DataDimension
component.
The item will still be draggable like the others and the selectedItems
list is handled as before (all the items are used).
DV app and d2ca have the logic for using only the first item in the list
for single value visualization type.
The feature here is basically just applying a different style to the
item component to visually differentiate the one actually used from the
rest.
By reordering the selected items (right hand side of the selector) is
possible to change the item used by the single value visualization.


edoardo:
@jenniferarnesen it would be nice if you can review this as you worked on the ItemSelector.
This PR is a proposal solution for the active/inactive state of dimension items and the info box in the selected items part of the ItemSelector.

See this [sketch](https://sketch.cloud/s/abbJq/a/2JxDdM/play) demonstrating the active/inactive styling for data dimension items and the "info box".

We want to keep this components generic, so the logic for the info message must be implemented in the app and passed to the component.
The `infoBoxMessage` prop is passed from the app to DataDimension with the text to show and if some text is defined the info box is rendered in ItemSelector.
Other components that use ItemSelector could use the prop too (ie. DynamicDimensions) if needed in the future.

src/components/ItemSelector/SelectedItems.js
```diff
@@ -17,6 +18,14 @@ const Subtitle = () => (
     </div>
 )
 
+const InfoBox = ({ message }) => (
+    <div className="info-container">
+        <InfoIcon style={{ fontSize: 16, color: '#6E7A8A' }} />
```
jenniferarnesen: Is the color something we can bring in from ui-core?

jenniferarnesen:
Do you think it's possible to add a story for this use case?

src/components/ItemSelector/SelectedItems.js
edoardo: I think so, there is a colors file with some definitions, that can probably be removed altogether and use ui-core colors instead.

edoardo:
> Do you think it's possible to add a story for this use case?

I guess so. I wanted to be sure we are going for this solution before writing tests.

2 new stories added for active/inactive and for infobox message.

jenniferarnesen (approved):

PR #45, opened 2019-06-27
fix: summarize ou level and groups in filter title

jenniferarnesen:
Part of implementation for [DHIS2-6990]

If the filter contains ou levels and/or groups, then summarize this instead of listing each individual ou (which could be a very long list)


src/visualizations/config/adapters/dhis_highcharts/getFilterTitle.js
```diff
@@ -1,44 +1,99 @@
...
+                    titleParts.push(getOuTitle(items, metaData, true))
+                }
+            } else {
+                const filterItems = metaData.dimensions[filter.dimension]
 
-            if (isArray(filterItems)) {
-                l = filterItems.length
-                let id
+                if (isArray(filterItems)) {
+                    const hasMetadata = filterItems.some(id =>
```
jenniferarnesen: I'm not sure whether this is a safe change.  Before, it checked the metadata for each item. Now it just checks that at least one item is present in the metadata.

jenniferarnesen: Restoring the "old" way

edoardo (dismissed):
Tested in DV and dashboards, works well.

edoardo (approved):

edoardo:
@jenniferarnesen we need to port this to the dhis adapter too.
Now that we have 2 adapters perhaps we should have some shared code.

jenniferarnesen:
@edoardo Could you take a second look at this, due to the refactor to support single value items?

src/visualizations/config/adapters/dhis_highcharts/getFilterText.js
```diff
@@ -1,68 +1,62 @@
...
+            titleFragments.push(getOuFilterText(filter, metaData))
+        } else {
+            const filterItems = metaData.dimensions[filter.dimension]
+
+            if (Array.isArray(filterItems)) {
+                l = filterItems.length
+                let id
+                const sectionParts = []
+
+                for (i = 0; i < l; i++) {
```
edoardo, line 36: Maybe simplify this using `filterItems.forEach`?

edoardo (approved):

src/visualizations/config/adapters/dhis_highcharts/getFilterText.js
jenniferarnesen: I left this code as is (it is the original implementation before I added anything), because it uses a `break`, which you cant do with foreach.

edoardo: Right!

edoardo (approved):

PR #72, opened 2019-08-01
fix: show "No data" instead of undefined when no data is returned

edoardo:


src/visualizations/config/index.js
```diff
@@ -5,6 +5,9 @@ import { theme1 } from '../util/colors'
 
 const DEFAULT_EXTRA_OPTIONS = {
     colors: theme1,
+    noData: {
+        text: 'No data',
```
janhenrikoverland: Change to "No data to display"?

janhenrikoverland (approved):

PR #79, opened 2019-08-13
fix: add function to replace a dimension in the layout

jenniferarnesen:
The specific use case that caused the creation of this function:
- In the past, when a visualization layout is saved and it contains org unit levels, the level number (e.g., 1, 2, 3, 4) was saved. In the new DV, we now store the uid of the org unit level, but for historical data, the layout will still contain the numeric values. To soft-migrate old data to the uid, we need to replace the ou dimension in the layout.

An additional function was added: axisHasOuDimension, to parallel the other related functions (e.g., axisHasPeriodDimension).

[DHIS2-6990]

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -47,6 +47,7 @@ export {
     layoutGetAxisNameDimensionIdsObject,
 } from './modules/layout/layoutGetAxisNameDimensionIdsObject'
 export { layoutGetDimension } from './modules/layout/layoutGetDimension'
+export { layoutReplaceDimension } from './modules/layout/layoutReplaceDimension'
 export {
     layoutGetDimensionIdItemIdsObject,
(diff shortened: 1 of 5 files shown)


jenniferarnesen:
Comments from Jan:

1) use `dimensionId` instead of `dimensionName`

Jennifer's reply: Done

2) as we cut the reference to both the layout object, the axis array and the dimension object (which is good) - it would make sense to create a `layoutReplaceAxis` function as well (which i could reuse in `layoutFilterDimensions` btw)
just to be pedantic we could even make a `layoutReplaceLayout` (or just `layoutReplace`) function so we don't repeat the `Object.assign({}, layout)` part

Jennifer's reply: I can do this as a separate PR, in order to limit the scope of this one.

janhenrikoverland (approved):

PR #80, opened 2019-08-26
feat: add layoutGetDimensionItems function

janhenrikoverland:
- Adds function "layoutGetDimensionItems" to help locate and return items for a specific dimension from a layout object + test
- Adds some basic validation to two functions so that they are safe to nest


jenniferarnesen (approved):

jenniferarnesen (changes requested):
Can we have a discussion regarding my previous comment?

janhenrikoverland:
Ref Jennifer's comments, we have discussed this a bit and agree that we should find a better solution. However we would have to follow that pattern across the board, so merging this PR now to be able to complete https://github.com/dhis2/data-visualizer-app/pull/313

PR #84, opened 2019-08-28
fix: expand user org unit names and keep relative period names in title/subtitle

edoardo:
Fixes DHIS2-7460.

The list of OUs name is shown in place of relative OU ids like USER_ORGUNIT. USER_ORGUNIT_CHILDREN...


jenniferarnesen (changes requested):
Is it possible to centralize the getFilterText code so that we don't have one adapter reaching into another?  If there is no time for this now, then go ahead and override my request for changes and we can fix that immediately afterwards.

Also, could you add the Jira issue to the PR?

edoardo:
> Is it possible to centralize the getFilterText code so that we don't have one adapter reaching into another? If there is no time for this now, then go ahead and override my request for changes and we can fix that immediately afterwards.

Yes, definitely.
That is the end goal, where should we move the shared code?
We can perhaps use `utils` under `visualization` or some new folder there for shared stuff.
I wouldn't go to `modules` outside `visualization` as this stuff is just for visualization adapters.

> Also, could you add the Jira issue to the PR?

Added in the description.

jenniferarnesen:
Sure, go ahead and put it in utils

edoardo:
I realised that with this change, the relative periods get expanded.
So for example, while before for LAST_12_MONTHS we had "Last 12 months" now it lists all the last 12 months separated by comma.

We should rather keep the short descriptive labels where available as the goal of all the work on title/subtitle is to make it more compact and readable possible.

I thought to see if we could detect relative period ids and use the associated label for it instead of expanding the items, but perhaps we can have a more generic solution where if a name is available for the particular filter id, it should be used (which is what we want for relative periods), otherwise the filter items can be expanded and their names are used (which is what we want for user org units).

edoardo:
Fixed the relative period issue.
Now if a name is available for the relative period id, it is used instead of using the list of items included in the period.

jenniferarnesen (approved):

PR #130, opened 2019-11-04
feat: add layout rules and helpers

janhenrikoverland:
- Add layout rules and helpers
- Add chart type and layout type contants and mappers
- Add util functions for layout rules
- Added rule tests


martinkrulltott (approved):
Looks great!

PR #143, opened 2019-11-13
fix: add support for PT

edoardo:
This is the minimum needed for the DV branches based on master which
implement PT stuff to work.


martinkrulltott (approved):

martinkrulltott:
> This is the minimum needed for the DV branches based on master which
> implement PT stuff to work.

Looks good, as agreed I changed the default vis type back to Column for temporary backwards compatibility.

PR #151, opened 2019-11-20
fix: clean up layout fns

martinkrulltott:


src/modules/fixedDimensions.js
```diff
@@ -27,3 +27,8 @@ export const FIXED_DIMENSIONS = {
         icon: OrgUnitIcon,
     },
 }
+
+export const filterFixedDimensions = dimensionIds =>
```
edoardo: I would rename this function so it indicates that it filters **out** the fixed dimensions.

edoardo (approved):

janhenrikoverland (approved):

PR #153, opened 2019-11-20
fix: rename axis name to axis id

janhenrikoverland:
Rename all variants of "axis name", like AXIS_NAME, axisName etc to "axis id".


martinkrulltott (approved):

src/modules/axis.js
```diff
@@ -1,25 +1,25 @@
...
 }
 
-export const getAxisDisplayName = axisName => {
-    const displayName = axisDisplayNames[axisName]
+export const getAxisName = axisId => {
+    const displayName = axisNames[axisId]
 
     if (!displayName) {
-        throw new Error(`${axisName} is not a valid axis id`)
+        throw new Error(`${axisId} is not a valid axis id`)
```
martinkrulltott, line 22: Consider updating the message?

src/modules/axis.js
```diff
@@ -1,25 +1,25 @@
...
+    [AXIS_ID_ROWS]: i18n.t('Category'),
+    [AXIS_ID_FILTERS]: i18n.t('Filter'),
+    [AXIS_ID_YEAR_OVER_YEAR_SERIES]: i18n.t('Series'),
+    [AXIS_ID_YEAR_OVER_YEAR_CATEGORY]: i18n.t('Category'),
 }
 
-export const getAxisDisplayName = axisName => {
-    const displayName = axisDisplayNames[axisName]
+export const getAxisName = axisId => {
+    const displayName = axisNames[axisId]
```
martinkrulltott: "displayName" should be "axisName"?

martinkrulltott (approved):

src/modules/axis.js
janhenrikoverland: :+1:

PR #156, opened 2019-11-21
fix: ui glitch for data dim groups DHIS2-7907

martinkrulltott:
* Fixes UI glitch for the data dimension group item selector which was misaligned with the container element (specified here: https://jira.dhis2.org/browse/DHIS2-7907)

### DIFF
--- a/src/components/DataDimension/styles/Groups.style.js
+++ b/src/components/DataDimension/styles/Groups.style.js
@@ -4,7 +4,7 @@ export const styles = {
     container: {
         border: `1px solid ${colors.greyLight}`,
         display: 'flex',
-        flexFlow: 'column',
+        flexDirection: 'column',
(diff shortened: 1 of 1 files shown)


src/components/DataDimension/styles/Groups.style.js
```diff
@@ -4,18 +4,19 @@ export const styles = {
     container: {
         border: `1px solid ${colors.greyLight}`,
         display: 'flex',
+        flexFlow: 'column',
```
janhenrikoverland: This only sets flexDirection so I guess we should say that instead.

edoardo (approved):

janhenrikoverland (changes requested):

janhenrikoverland (approved):

PR #172, opened 2019-12-03
fix: rules tests

martinkrulltott:
* Test scripts for the UI rules engine
* Named constants for rule props


src/modules/layoutUiRules/__tests__/rules.spec.js
```diff
@@ -1,40 +1,182 @@
...
+                    Object.keys(rule),
+                    Object.values(testResourceAllRuleProps)
+                )
             )
-        ).toBe(false)
+        ).toBe(true)
+    })
+})
+
+describe("verify each rule's ", () => {
```
janhenrikoverland, line 105: ?

janhenrikoverland (approved):

src/modules/layoutUiRules/__tests__/rules.spec.js
martinkrulltott: This was added to improve the readability of the test output and enables the following:
Screenshot 2019-12-05 at 10 25 19

PR #184, opened 2019-12-17
feat: always show dual axis DHIS2-8009

martinkrulltott:
[Ticket](https://jira.dhis2.org/browse/DHIS2-8009)

### DIFF
--- a/src/components/DimensionMenu.js
+++ b/src/components/DimensionMenu.js
@@ -1,4 +1,4 @@
-import React, { Component } from 'react'
+import React from 'react'
 import Divider from '@material-ui/core/Divider'
 import Menu from '@material-ui/core/Menu'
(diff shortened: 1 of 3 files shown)


src/components/DimensionMenu.js
```diff
@@ -42,97 +24,158 @@ const getRemoveMenuItem = onClick => (
...
-            getDualAxisItem(dimensionId, () => {
-                dualAxisItemHandler()
-                onClose()
-            })
-        )
+    numberOfDimensionItems
+) => {
+    let label
+
+    if (!isDualAxisType(visType)) {
```
martinkrulltott, line 34: This could be solved in a more compact / ES6 way and possibly be merged with the logic check on line 100. @janhenrikoverland thoughts?

janhenrikoverland (approved):

PR #222, opened 2020-01-14
fix: ui glitch for data elements v2(DHIS2-8044)

martinkrulltott:
* UI fix for a glitch with the data element selector, that's causing elements to overlap.


martinkrulltott:
Fixed: image

janhenrikoverland (approved):

PR #253, opened 2020-01-22
fix: helpers for full axis, locked dimension, dimension can be added to axis

edoardo:
These are used when a dimension is added/moved/dragged in the layout.
It respects the layout rules by checking the current state of the destination axis (is full, the dimension already present is locked...) to avoid incorrectly adding a dimension.

### DIFF
--- a/src/modules/layoutUiRules/index.js
+++ b/src/modules/layoutUiRules/index.js
@@ -10,8 +10,8 @@ export {
 
 export {
     hasAxisTooManyItemsByVisType as hasAxisTooManyItems,
-    isDimensionLocked,
-    isAxisFull,
-    canDimensionBeAddedToAxis,
+    isDimensionLockedByVisType as isDimensionLocked,
+    isAxisFullByVisType as isAxisFull,
+    canDimensionBeAddedToAxisByVisType as canDimensionBeAddedToAxis,
     getTransferableDimensionPerAxisByVisType as getTransferableDimension,
 } from './rulesUtils'
--- a/src/modules/layoutUiRules/index.js
+++ b/src/modules/layoutUiRules/index.js
@@ -7,4 +7,7 @@ export {
     hasAxisTooManyItemsByVisType as hasAxisTooManyItems,
     getAxisPerLockedDimByVisType as getAxisPerLockedDimension,
     getAllLockedDimIdsByVisType as getAllLockedDimensionIds,
+    isDimensionLocked,
+    isAxisFull,
+    canDimensionBeAddedToAxis,
 } from './rulesHelper'
--- a/src/modules/layoutUiRules/index.js
+++ b/src/modules/layoutUiRules/index.js
@@ -4,10 +4,14 @@ export {
     getAxisMaxNumberOfItemsByVisType as getAxisMaxNumberOfItems,
     getAxisMaxNumberOfDimsByVisType as getAxisMaxNumberOfDimensions,
     getAxisMinNumberOfDimsByVisType as getAxisMinNumberOfDimensions,
-    hasAxisTooManyItemsByVisType as hasAxisTooManyItems,
     getAxisPerLockedDimByVisType as getAxisPerLockedDimension,
     getAllLockedDimIdsByVisType as getAllLockedDimensionIds,
+} from './rulesHelper'
+
+export {
+    hasAxisTooManyItemsByVisType as hasAxisTooManyItems,
     isDimensionLocked,
     isAxisFull,
     canDimensionBeAddedToAxis,
-} from './rulesHelper'
+    getTransferableDimensionPerAxisByVisType as getTransferableDimension,
+} from './rulesUtils'
--- a/src/modules/layoutUiRules/rulesHelper.js
+++ b/src/modules/layoutUiRules/rulesHelper.js
@@ -38,3 +38,25 @@ export const getAxisPerLockedDimByVisType = (visType, dimensionId) =>
 
 export const getAllLockedDimIdsByVisType = visType =>
     Object.keys(getLockedDimsByVisType(visType))
+
+export const isDimensionLocked = (visType, dimensionId) =>
+    getAllLockedDimIdsByVisType(visType).find(id => id === dimensionId)
+
+export const isAxisFull = (visType, axisId, axisDimensionsCount) =>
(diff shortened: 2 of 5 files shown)


src/modules/layoutUiRules/rulesHelper.js
```diff
@@ -38,3 +38,25 @@ export const getAxisPerLockedDimByVisType = (visType, dimensionId) =>
...
+    axisDimensionsCount === getAxisMaxNumberOfDimsByVisType(visType, axisId)
+
+export const canDimensionBeAddedToAxis = (visType, layout, axisId) => {
+    const axisIsFull = isAxisFull(visType, axisId, layout[axisId].length)
+    const dimensionIsLocked = isDimensionLocked(visType, layout[axisId][0])
+
+    // 1 dimension allowed in axis
+    // 1 dimension is already present and not locked
+    // the dragged one can be added and will cause the old one to be moved to filters
+    if (axisIsFull && !dimensionIsLocked) {
```
martinkrulltott: shorten to `return (axisIsFull && !dimensionIsLocked) || (!axisIsFull)` ?

src/modules/layoutUiRules/rulesHelper.js
```diff
@@ -38,3 +38,25 @@ export const getAxisPerLockedDimByVisType = (visType, dimensionId) =>
 
 export const getAllLockedDimIdsByVisType = visType =>
     Object.keys(getLockedDimsByVisType(visType))
+
+export const isDimensionLocked = (visType, dimensionId) =>
+    getAllLockedDimIdsByVisType(visType).find(id => id === dimensionId)
+
+export const isAxisFull = (visType, axisId, axisDimensionsCount) =>
+    axisDimensionsCount === getAxisMaxNumberOfDimsByVisType(visType, axisId)
```
martinkrulltott: should check >= instead of === ?

src/modules/layoutUiRules/rulesHelper.js
```diff
@@ -38,3 +38,25 @@ export const getAxisPerLockedDimByVisType = (visType, dimensionId) =>
...
+
+export const canDimensionBeAddedToAxis = (visType, layout, axisId) => {
```
martinkrulltott: why pass the whole `layout` when only `layout[axisId]` is used?

martinkrulltott (commented):
Generally I'm not sure if I agree with adding these three new methods to `rulesHelper`.
Internally they are sound, but should they really live here? `isDimensionLocked` is only an extension of `isDimensionLocked`, `isAxisFull` extends `getAxisMaxNumberOfDimsByVisType` and `canDimensionBeAddedToAxis` extends the other two. Even though these methods highly depend on the rules, what they're doing for `rulesHelper` though is just adding layout _logic_ to the _rules_ engine. Is this our preferred pattern? 🤔 @janhenrikoverland

src/modules/layoutUiRules/rulesHelper.js
janhenrikoverland: Or just `return !(axisIsFull && dimensionIsLocked)`

src/modules/layoutUiRules/rulesHelper.js
janhenrikoverland: I guess it won't hurt

edoardo: Having the count greater than the max limit indicates that somewhere the rule has been violated.
Could that perhaps happen when loading an existing visualization?

martinkrulltott: Not sure when/if this would happen, but for future proofing I would check `>=`, since greater than (even though there might be an error somewhere) still counts as "full" right?

src/modules/layoutUiRules/rulesHelper.js
edoardo: This one changed a bit now since `dimensionIsLocked` is not used anymore in the check.

src/modules/layoutUiRules/rulesHelper.js
edoardo: Fixed.

src/modules/layoutUiRules/rulesHelper.js
edoardo: Fixed

src/modules/layoutUiRules/rulesUtils.js
```diff
@@ -0,0 +1,59 @@
...
+    // what happens with max limit > 1, axis full and 1 or more locked dimensions?
+    return !(axisIsFull && !transferableDimension)
+}
+
+export const getTransferableDimensionPerAxisByVisType = (
+    visType,
+    axisId,
+    axis
+) => {
+    const lockedDimsByVisType = getLockedDimsByVisType(visType)
```
martinkrulltott, line 50: can you replace `const lockedDimsByVisType = getLockedDimsByVisType(visType)` and `Object.keys(lockedDimsByVisType)` with `getAllLockedDimIdsByVisType` from rules helper and skip the`'./rules'` import?

edoardo: I need the `axisId` information in order to do the filtering...
I can use `getAllLockedDimIdsByVisType` but I still need `getLockedDimsByVisType` for the link between dimension and axis.

src/modules/layoutUiRules/index.js
```diff
@@ -4,7 +4,14 @@ export {
...
     getAxisMaxNumberOfDimsByVisType as getAxisMaxNumberOfDimensions,
     getAxisMinNumberOfDimsByVisType as getAxisMinNumberOfDimensions,
-    hasAxisTooManyItemsByVisType as hasAxisTooManyItems,
     getAxisPerLockedDimByVisType as getAxisPerLockedDimension,
     getAllLockedDimIdsByVisType as getAllLockedDimensionIds,
 } from './rulesHelper'
+
+export {
+    hasAxisTooManyItemsByVisType as hasAxisTooManyItems,
+    isDimensionLocked,
```
martinkrulltott: Technically, to follow the pattern, the internal names for `isDimensionLocked`, `isAxisFull` and `canDimensionBeAddedToAxis` should be suffixed by `...ByVisType`

martinkrulltott (approved):

src/modules/layoutUiRules/index.js
edoardo: Fixed

PR #262, opened 2020-01-28
fix: fixed and dynamic dimensions

martinkrulltott:
* This separates code based dimensions in to `fixed` (dx, pe, oa) and `dynamic` (assigned categories), which are both `predefined`. All other dimensions (i.e. instance dependant dimensions) are `nonpredefined`.
* Breaking changes for DV, Dashboards and any other dependants.


edoardo (approved):

janhenrikoverland (approved):

janhenrikoverland:
Maybe make a constant of the "noItems" prop and export it. Then use it in DV instead of:
`getPredefinedDimensionProp(dimensionId, 'noItems')` so that the constant's value can easily be changed later without breaking apps.

janhenrikoverland (approved):

PR #264, opened 2020-01-28
refactor: move pivot table to a multi-component layout

amcgee:
This is a potential refactor into multiple components, for feedback on improved cleanliness and readability.  Does this make the logic more clear or less clear?

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@dhis2/analytics",
-    "version": "3.0.0-alpha.1",
+    "version": "3.0.0-alpha.2",
     "main": "./build/index.js",
     "sideEffects": false,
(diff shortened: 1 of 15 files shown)


amcgee:
Previous version here, now split into multiple component files
https://github.com/dhis2/analytics/blob/feat/pivot-table/src/components/PivotTable/PivotTable.js

amcgee:
Would recommend looking at the [source](https://github.com/dhis2/analytics/tree/refactor/pivot-table-multicomponent/src/components/PivotTable) rather than the diff

edoardo:
I prefer the split version.
We'll probably need to expand PivotTable.js further for supporting various features, so it can quickly become too big and cumbersome to work with.

edoardo (approved):
LGTM

PR #269, opened 2020-01-30
fix: dimensions panel redesign

martinkrulltott:
* Divides the Dimension Panel in to sections in accordance to the new design, with "Main" and "Your" dimensions.
* Style update for the dimensions search at the top
* Since DV now uses it's own implementation of the dimensions panel this change is only targeted to Dashboards


janhenrikoverland (approved):
LGTM

PR #284, opened 2020-02-05
fix: dynamic axis names based on vis type

martinkrulltott:
* Dynamically fetching axis names based on vis type or layout type

Note: `getAxisName` is deprecated but kept for backwards compatibility. It's replaced by `getAxisNameByLayoutType`

### DIFF
--- a/src/components/DimensionMenu.js
+++ b/src/components/DimensionMenu.js
@@ -12,10 +12,13 @@ import { getAvailableAxes } from '../modules/layoutUiRules'
 import { AXIS_ID_COLUMNS, AXIS_ID_FILTERS } from '../modules/layout/axis'
 import { DIMENSION_ID_DATA } from '../modules/predefinedDimensions'
 import { isDualAxisType, getDisplayNameByVisType } from '../modules/visTypes'
-import { getAxisNameByVisType } from '../modules/axis'
+import { getAxisNameByLayoutType } from '../modules/axis'
+import { getLayoutTypeByVisType } from '../modules/visTypeToLayoutType'
 
-const getAxisItemLabelPrefix = isDimensionInLayout =>
-    isDimensionInLayout ? 'Move to' : 'Add to'
+const getAxisItemLabel = (axisName, isDimensionInLayout) =>
+    isDimensionInLayout
+        ? i18n.t('Move to {{axisName}}', { axisName })
+        : i18n.t('Add to {{axisName}}', { axisName })
 
 const getRemoveMenuItem = onClick => (
     <MenuItem key="remove-menu-item" onClick={onClick}>
@@ -185,10 +188,12 @@ export class DimensionMenu extends Component {
                                             closeWholeMenu()
                                         }}
                                     >
-                                        {`${getAxisItemLabelPrefix()} ${getAxisNameByVisType(
-                                            destination,
-                                            visType
-                                        )}`}
+                                        {getAxisItemLabel(
+                                            getAxisNameByLayoutType(
+                                                destination,
+                                                getLayoutTypeByVisType(visType)
+                                            )
+                                        )}
                                     </MenuItem>
                                 )
                             )}
@@ -244,10 +249,12 @@ export class DimensionMenu extends Component {
                         closeWholeMenu()
                     }}
                 >
-                    {i18n.t(
-                        `${getAxisItemLabelPrefix(
-                            isDimensionInLayout
-                        )} ${getAxisNameByVisType(axisId, visType)}`
+                    {getAxisItemLabel(
+                        getAxisNameByLayoutType(
+                            axisId,
+                            getLayoutTypeByVisType(visType)
+                        ),
+                        isDimensionInLayout
                     )}
                 </MenuItem>
             ))
--- a/src/components/DimensionMenu.js
+++ b/src/components/DimensionMenu.js
@@ -26,8 +26,8 @@ const getRemoveMenuItem = onClick => (
 const getDividerItem = key => <Divider light key={key} />
 
 const getUnavailableLabel = visType =>
-    i18n.t('Not available for {{visType}}', {
-        visType: getDisplayNameByVisType(visType),
+    i18n.t('Not available for {{visualizationType}}', {
+        visualizationType: getDisplayNameByVisType(visType),
     })
 
 const getDualAxisMenuItemLabel = (
@@ -185,12 +185,10 @@ export class DimensionMenu extends Component {
                                             closeWholeMenu()
                                         }}
                                     >
-                                        {i18n.t(
-                                            `${getAxisItemLabelPrefix()} ${getAxisNameByVisType(
-                                                destination,
-                                                visType
-                                            )}`
-                                        )}
+                                        {`${getAxisItemLabelPrefix()} ${getAxisNameByVisType(
+                                            destination,
+                                            visType
+                                        )}`}
                                     </MenuItem>
                                 )
                             )}
--- a/src/components/DimensionMenu.js
+++ b/src/components/DimensionMenu.js
@@ -12,7 +12,7 @@ import { getAvailableAxes } from '../modules/layoutUiRules'
 import { AXIS_ID_COLUMNS, AXIS_ID_FILTERS } from '../modules/layout/axis'
 import { DIMENSION_ID_DATA } from '../modules/predefinedDimensions'
 import { isDualAxisType, getDisplayNameByVisType } from '../modules/visTypes'
-import { getAxisName } from '../modules/axis'
+import { getAxisNameByVisType } from '../modules/axis'
 
 const getAxisItemLabelPrefix = isDimensionInLayout =>
     isDimensionInLayout ? 'Move to' : 'Add to'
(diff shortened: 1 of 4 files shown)


src/modules/axis.js
```diff
@@ -1,26 +1,41 @@
...
-    [AXIS_ID_YEAR_OVER_YEAR_SERIES]: i18n.t('Series'),
-    [AXIS_ID_YEAR_OVER_YEAR_CATEGORY]: i18n.t('Category'),
+const getAxisNamesByLayoutType = layoutType => {
+    switch (layoutType) {
+        case LAYOUT_TYPE_DEFAULT:
+        case LAYOUT_TYPE_PIE:
+        case LAYOUT_TYPE_YEAR_OVER_YEAR:
+        default:
+            return {
+                [AXIS_ID_COLUMNS]: i18n.t('Series'),
```
amcgee, line 17: So this is actually a much bigger problem in this library... we **don't actually support i18n extraction and generation in @dhis2/analytics currently!**

We're using `d2-i18n` several places, even before this change, but since we don't actually look for those translations in the build step or pull them out into a `pot` file, let alone load them at runtime, they will never actually be translated.  CC @janhenrikoverland @edoardo @jenniferarnesen

src/modules/axis.js
```diff
@@ -1,26 +1,41 @@
...
+export const getAxisNameByLayoutType = (axisId, layoutType) => {
+    const name = getAxisNamesByLayoutType(layoutType)[axisId]
     if (!name) {
         throw new Error(`${axisId} is not a valid axis id`)
     }
 
     return name
 }
+
+export const getAxisNameByVisType = (axisId, visType) =>
```
amcgee: Do we really need two new exported functions or can we limit the API surface area and expose just one (the existing one)?

amcgee (changes requested):
Can we make this NOT a breaking change by just adding an (optional) second parameter to `getAxisName`?  So like this:

```js
const getAxisName = (axisId, layoutType = LAYOUT_TYPE_DEFAULT) => (
  getAxisNameByVisType(axisId, layoutType)
)
```

We can expose the new functions, but I would prefer to incorporate them into the existing `getAxisName` function.  This will help a lot with backports as well, if they're ever necessary.

src/modules/axis.js
martinkrulltott: So the idea is that since `layoutType` is whats actually used for the lookup but in most places we only have access to `visType`, the `getAxisNameByVisType` removes one step of the process (getting `layoutType` from `visType`). But in some cases (mostly regarding `YearOverYear`), we need to explicitly get the name by `layoutType`, hence the two exported functions.

src/modules/axis.js
martinkrulltott: Good catch! This is a major issue. Continuing the discussion on Slack.

martinkrulltott: Definitely something we need to address, but since it's out of scope for this specific PR I'm marking this as resolved.

src/components/DimensionMenu.js
```diff
@@ -248,7 +249,7 @@ export class DimensionMenu extends Component {
                     {i18n.t(
                         `${getAxisItemLabelPrefix(
                             isDimensionInLayout
-                        )} ${getAxisName(axisId)}`
```
amcgee, line 251: This should use `interpolation` instead of string template literals, even though we're not currently extracting the strings it's a good idea to fix it now.

src/components/DimensionMenu.js
```diff
@@ -186,8 +186,9 @@ export class DimensionMenu extends Component {
                                         }}
                                     >
                                         {i18n.t(
-                                            `${getAxisItemLabelPrefix()} ${getAxisName(
```
amcgee, line 189: This should also use [`interpolation`](https://www.i18next.com/translation-function/interpolation) instead of a string template literal.

amcgee: And actually there's no point calling `i18n.t` here at all since there's no static text to translate

amcgee (changes requested):
> Can we make this NOT a breaking change by just adding an (optional) second parameter to getAxisName? So like this:
> ```
> const getAxisName = (axisId, layoutType = LAYOUT_TYPE_DEFAULT) => (
>   getAxisNameByVisType(axisId, layoutType)
> )
> ```
> We can expose the new functions, but I would prefer to incorporate them into the existing getAxisName function. This will help a lot with backports as well, if they're ever necessary.

@martinkrulltott see the above comment about un-breaking this change.  I'm also not convinced `getAxisNameByVisType` is necessary when we could just call `getAxisName(axisId, getLayoutTypeByVisType(visType))`

src/components/DimensionMenu.js
amcgee: This `i18n.t` is also not going to do anything, so it should be removed.  Just note for the future that using a string template literal here will ask translators to translate your code, rather than a string with a variable, so [interpolation](https://www.i18next.com/translation-function/interpolation) is required when including variables in translated strings.

martinkrulltott:
> I'm also not convinced `getAxisNameByVisType` is necessary when we could just call `getAxisName(axisId, getLayoutTypeByVisType(visType))`

Yes, it's possible to call `getLayoutTypeByVisType` bundled every time you need `axisName` but only have access to `visType`, but as this accounts for almost all of the cases and probably 99% of real-world usage, I thought the extra Analytics call was unnecessary and being able to just fetch `axisName` by `visType` directly would be more convenient? Thoughts @janhenrikoverland ?

src/components/DimensionMenu.js
martinkrulltott: Valid point, i18n removed

src/components/DimensionMenu.js
martinkrulltott: Valid point. Removed i18n here

src/modules/axis.js
martinkrulltott: `getAxisNameByVisType` is now removed

src/components/DimensionMenu.js
```diff
@@ -12,7 +12,7 @@ import { getAvailableAxes } from '../modules/layoutUiRules'
 import { AXIS_ID_COLUMNS, AXIS_ID_FILTERS } from '../modules/layout/axis'
 import { DIMENSION_ID_DATA } from '../modules/predefinedDimensions'
 import { isDualAxisType, getDisplayNameByVisType } from '../modules/visTypes'
-import { getAxisName } from '../modules/axis'
+import { getAxisNameByVisType } from '../modules/axis'
 
 const getAxisItemLabelPrefix = isDimensionInLayout =>
     isDimensionInLayout ? 'Move to' : 'Add to'
```
amcgee: These will (1) never be translated and (2) lack the context required to translate them correctly.  Since these are only used 2 places, this function should instead look like this:

```js
const getAxisItemLabel = (axisName, isDimensionInLayout) => 
    isDimensionInLayout 
      ? i18n.t('Move to {axisName}', { axisName })
      : i18n.t('Add to {axisName}', { axisName })
```

And the two references to this function below should be updated.

src/modules/axis.js
```diff
@@ -1,26 +1,40 @@
...
+    }
 }
 
-export const getAxisName = axisId => {
-    const name = axisNames[axisId]
-
+export const getAxisNameByLayoutType = (axisId, layoutType) => {
+    const name = getAxisNamesByLayoutType(layoutType)[axisId]
     if (!name) {
         throw new Error(`${axisId} is not a valid axis id`)
```
amcgee, line 33: This error should be updated to include the `layoutType`, since it might be a valid `axisId` in a different context:

```js
throw new Error(`${axisId} is not a valid axis id in layout ${layoutType}`) // Assuming this has a meaningful string value here...
```

src/components/DimensionMenu.js
```diff
@@ -248,7 +247,7 @@ export class DimensionMenu extends Component {
                     {i18n.t(
```
amcgee: this `i18n.t` call should be removed as it does nothing (`getAxisItemLabel` should be used instead, see comment above)

amcgee (changes requested):
Looking good!  Sorry for all the nitpicks - found another translation bug.  I realize most of this was already existing, but might as well fix it since we're touching it now!

janhenrikoverland (approved):
Approved when the remaining comments have been resolved.

src/components/DimensionMenu.js
martinkrulltott: Indeed, same issue as found in https://github.com/dhis2/data-visualizer-app/pull/623, fixed.

src/components/DimensionMenu.js
martinkrulltott: Fixed

src/modules/axis.js
martinkrulltott: Keep in mind that `layoutType` doesn't have a display name so it would show up as e.g. `X is not a valid axis id in layout LAYOUT_TYPE_PIVOT_TABLE`. This error is really just for developers, if you for whatever reason send in something thats not an actual `axisId`. All axes are represented with a name in all layouts, so the case of an axis being part of one layout but not another isn't valid currently. But I don't see any harm in this 🙂 Fixed

amcgee (approved):
LGTM

PR #288, opened 2020-02-11
feat: gauge plot lines and range values

martinkrulltott:
Adds support for plot lines and custom range values for Gauge charts
image

Highcharts demo: https://jsfiddle.net/6dqpuax7/

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,7 +67,7 @@
         "classnames": "^2.2.6",
         "d2-utilizr": "^0.2.16",
         "d3-color": "^1.2.3",
-        "highcharts": "^7.1.2",
+        "highcharts": "^7.2.1",
         "lodash": "^4.17.13",
         "react-beautiful-dnd": "^10.1.1",
         "resize-observer-polyfill": "^1.5.1",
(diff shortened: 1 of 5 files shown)


janhenrikoverland:
Maybe base/target should also say what the actual value is? @cooper-joe

package.json
```diff
@@ -67,7 +67,7 @@
         "classnames": "^2.2.6",
         "d2-utilizr": "^0.2.16",
         "d3-color": "^1.2.3",
-        "highcharts": "^7.1.2",
+        "highcharts": "^7.2.1",
```
martinkrulltott, line 70: Upgraded to 7.2.1 (latest minor, to avoid 8.0.0) to fix an issue with the plot lines being drawn all the way to the center.

cooper-joe:
> Maybe base/target should also say what the actual value is? @cooper-joe

Yeah I think that would make the target/base lines even more useful. Formatting could be: `Target: 120` and `Base: 30`. If the user changes the label for that line it would still include the value after the user-entered label, e.g. `UserLabel: 300`

This could be a setting, but I think it's useful enough to just include them by default now.

martinkrulltott:
Updated to reflect the comments from @cooper-joe and @janhenrikoverland

edoardo (approved):
LGTM

PR #303, opened 2020-02-13
fix: weird pie chart graphics for selections with no data (DHIS2-8302)

janhenrikoverland:
Fixes the weird looking pie chart issue. Caused by series with NaN data. Fixed by leaving out data items with no data.

"Before" screenshot in the jira ticket: https://jira.dhis2.org/browse/DHIS2-8302

"After":
Screenshot from 2020-02-13 14-51-09

### DIFF
--- a/src/visualizations/store/adapters/dhis_highcharts/pie.js
+++ b/src/visualizations/store/adapters/dhis_highcharts/pie.js
@@ -5,7 +5,7 @@ export default function(acc, seriesIds, categoryIds, idValueMap, metaData) {
         if (value) {
             acc.push({
                 name: metaData.items[seriesId].name,
-                y: parseFloat(idValueMap.get(seriesId)),
+                y: parseFloat(value),
             })
         }
     })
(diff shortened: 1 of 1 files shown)


src/visualizations/store/adapters/dhis_highcharts/pie.js
```diff
@@ -1,8 +1,12 @@
...
-        }))
-    )
-}
+    seriesIds.forEach(seriesId => {
+        const value = idValueMap.get(seriesId)
+
+        if (value) {
+            acc.push({
+                name: metaData.items[seriesId].name,
+                y: parseFloat(idValueMap.get(seriesId)),
```
martinkrulltott: Shouldn't this be `y: parseFloat(value),` ?

martinkrulltott: Or maybe even move the `parseFloat` to `const value` to make sure it parsed okay in the `if (value)` check?

janhenrikoverland: But then there is a bigger problem somewhere higher up and I don't want to camouflage that.

martinkrulltott (approved):

PR #304, opened 2020-02-13
fix: correct string number parsing and rendering

amcgee:
@martinkrulltott this fixes the number parsing + rendering error you saw, thanks for raising the issue!

This also removes client-side float parsing (except when calculating totals) and precision enforcement, leaving that to the server.

Also removed some `console.log` invaders


martinkrulltott (approved):
Code LGTM, haven't tested locally though

janhenrikoverland (approved):
Seems to work well

PR #310, opened 2020-02-18
feat: support fixed legends and fix floating-point totals bug

amcgee:
This PR adds support for a new `legendSets` prop which can be passed to the `PivotTable` component.  This is an array of legendSets which have been fetched from the server - the caller (DV plugin) should fetch all required legendSets before rendering the PivotTable component.

This includes support for only FIXED legends, not BY_DATA_ITEM legends (coming soon) and also supports FILL/TEXT legend styling and automatic text color contrast for dark background colors.

Finally, this fixes a bug which would often cause a crash (trying to render a number when expecting a string) and another issue where floating-point total values were rendered without regard to precision (i.e. `15.000000003`).  I've re-introduced fixed-point conversion to address this.

There are several TODOs remaining (in comments), but I'd like to get this feature in to increase feature coverage incrementally rather than in a large PR.

### DIFF
--- a/src/modules/pivotTable/applyLegendSet.js
+++ b/src/modules/pivotTable/applyLegendSet.js
@@ -1,0 +1,68 @@
+import { isColorBright } from './isColorBright'
+
+const LEGEND_DISPLAY_STRATEGY_BY_DATA_ITEM = 'BY_DATA_ITEM'
+const LEGEND_DISPLAY_STRATEGY_FIXED = 'FIXED'
+
+const LEGEND_DISPLAY_STYLE_FILL = 'FILL'
+const LEGEND_DISPLAY_STYLE_TEXT = 'TEXT'
+
+const valueFitsLegend = (value, legend) =>
+    legend.startValue <= value && legend.endValue > value // TODO: Confirm inclusive/exclusive bounds
+
+const getLegendSet = engine => {
+    let legendSetId
+    switch (engine.visualization.legendDisplayStrategy) {
+        case LEGEND_DISPLAY_STRATEGY_BY_DATA_ITEM:
+            legendSetId = undefined // TODO: ByDXID LegendSet
+            //engine.rawData.metaData.items[dxId].legendSet
+            break
+        case LEGEND_DISPLAY_STRATEGY_FIXED:
+        default:
+            legendSetId = engine.visualization.legendSet
+                ? engine.visualization.legendSet.id
+                : undefined
+            break
+    }
+
+    return engine.legendSets[legendSetId]
+}
+
+const buildStyleObject = (legend, engine) => {
+    const style = {}
+    switch (engine.visualization.legendDisplayStyle) {
+        case LEGEND_DISPLAY_STYLE_TEXT:
+            style.color = legend.color
+            // TODO: Adapt background color for light text colors
+            break
+        case LEGEND_DISPLAY_STYLE_FILL:
+        default:
+            style.backgroundColor = legend.color
+            if (isColorBright(legend.color)) {
+                style.color = '#000000' // TODO: Don't hard-code color here!
+            } else {
+                style.color = '#FFFFFF' // TODO: Don't hard-code color here!
+            }
+            break
+    }
+    return style
+}
+
+export const applyLegendSet = (value, engine) => {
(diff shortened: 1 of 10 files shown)


amcgee:
@janhenrikoverland regarding [this line](https://github.com/dhis2/analytics/pull/310/files#diff-ddc3f7ea059b434e4ffa962a79e8f993R10) - the previous behavior was a bit ambiguous, because it used inclusive bounds on BOTH sides of the range and just matched whichever one it hit LAST (see [d2-analysis](https://github.com/dhis2/d2-analysis/blob/master/src/table/PivotTable.js#L1842) and [d2-utilizr](https://github.com/dhis2/d2-utilizr/blob/master/src/numberConstrain.js)).

I'm not sure what the intended behavior is here - if a legendSet has two legends with the ranges `0 to 100` and `100 to 1000` then which range should match the value `100.0`?  We probably want to clearly define which end of the range is **exclusive**, as I've done with the `endValue` currently, so that there isn't overlap.

src/modules/pivotTable/applyLegendSet.js
```diff
@@ -0,0 +1,68 @@
...
+
+const valueFitsLegend = (value, legend) =>
+    legend.startValue <= value && legend.endValue > value // TODO: Confirm inclusive/exclusive bounds
+
+const getLegendSet = engine => {
+    let legendSetId
+    switch (engine.visualization.legendDisplayStrategy) {
+        case LEGEND_DISPLAY_STRATEGY_BY_DATA_ITEM:
+            legendSetId = undefined // TODO: ByDXID LegendSet
+            //engine.rawData.metaData.items[dxId].legendSet
```
edoardo, line 18: Can we have multiple different legendSets at the same time?
If we put the fetch of the legendSet by dx id in the wrapper plugin, here we would just need to use whatever legend set is passed in.

src/modules/pivotTable/applyLegendSet.js
```diff
@@ -0,0 +1,68 @@
...
+    switch (engine.visualization.legendDisplayStyle) {
+        case LEGEND_DISPLAY_STYLE_TEXT:
+            style.color = legend.color
+            // TODO: Adapt background color for light text colors
+            break
+        case LEGEND_DISPLAY_STYLE_FILL:
+        default:
+            style.backgroundColor = legend.color
+            if (isColorBright(legend.color)) {
+                style.color = '#000000' // TODO: Don't hard-code color here!
```
edoardo: Do we have some colors for this in `ui-core` colors/theme?

edoardo (approved):
LGTM.
A couple of comments.

src/modules/pivotTable/applyLegendSet.js
amcgee: Yeah there can be multiple different legendSets - there might be multiple DX items which each has its own legendSet assigned, so we need to look up the relevant legendSet based on the dx of whichever cell we're in.

src/modules/pivotTable/applyLegendSet.js
amcgee: @cooper-joe I think this is a question for you - what should be text color be for both (a) dark and (b) light background colors in a legend-styled pivot table?  Right now the default is #000 text for light backgrounds and #FFF text for dark backgrounds, but maybe we want something slightly different?

amcgee: Another @cooper-joe question - what should happen on hover for cells with a background-color set by the legend?  Currently nothing does, but non-legend cells have a hover bg-color of [#F3F3F3](https://github.com/dhis2/analytics/blob/master/src/components/PivotTable/styles/PivotTable.style.js#L77)

cooper-joe: 1. Light background color can use `grey900` and dark backgrounds can use `white` from ui-core.colors.

2. I don't think we should apply the standard hover color on legend cells, it would be confusing. However, trying to compute a slightly darker shade of the legend-cell-color also sounds quite intensive and is actually changing the data (modifying the legend color modifies its meaning). Therefore, let's keep it that legend cells do not have a hover state.

amcgee: 👍 thanks @cooper-joe !

PR #315, opened 2020-02-20
fix: legends for gauge

martinkrulltott:
Enables Gauge to display custom colors based on legends, either for the fill color or the text color.

Text color:
image

Fill color:
image

### DIFF
--- a/src/modules/legends.js
+++ b/src/modules/legends.js
@@ -15,5 +15,5 @@ export const getColorByValueFromLegendSet = (legendSet, value) => {
     const legend = legendSet.legends.find(
         legend => value >= legend.startValue && value < legend.endValue // TODO: Confirm inclusive/exclusive bounds
     )
-    return legend ? legend.color : undefined
+    return legend && legend.color
 }
--- a/src/modules/legends.js
+++ b/src/modules/legends.js
@@ -4,6 +4,13 @@ export const LEGEND_DISPLAY_STRATEGY_FIXED = 'FIXED'
 export const LEGEND_DISPLAY_STYLE_FILL = 'FILL'
 export const LEGEND_DISPLAY_STYLE_TEXT = 'TEXT'
 
+export const getLegendSetById = (legendSets, id) => {
+    if (!legendSets || !id) {
+        return {}
+    }
+    return legendSets.find(legendSet => legendSet.id === id)
+}
+
 export const getColorByValueFromLegendSet = (legendSet, value) => {
     const legend = legendSet.legends.find(
         legend => value >= legend.startValue && value < legend.endValue // TODO: Confirm inclusive/exclusive bounds
--- a/src/modules/legends.js
+++ b/src/modules/legends.js
@@ -1,3 +1,9 @@
+export const LEGEND_DISPLAY_STRATEGY_BY_DATA_ITEM = 'BY_DATA_ITEM'
+export const LEGEND_DISPLAY_STRATEGY_FIXED = 'FIXED'
+
+export const LEGEND_DISPLAY_STYLE_FILL = 'FILL'
+export const LEGEND_DISPLAY_STYLE_TEXT = 'TEXT'
+
 export const getColorByValueFromLegendSet = (legendSet, value) => {
     const legend = legendSet.legends.find(
         legend => value >= legend.startValue && value < legend.endValue // TODO: Confirm inclusive/exclusive bounds
--- a/src/modules/legends.js
+++ b/src/modules/legends.js
@@ -1,0 +1,6 @@
+export const getColorByValueFromLegendSet = (legendSet, value) => {
+    const legend = legendSet.legends.find(
+        legend => value >= legend.startValue && value < legend.endValue // TODO: Confirm inclusive/exclusive bounds
+    )
+    return legend ? legend.color : undefined
+}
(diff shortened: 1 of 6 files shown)


src/modules/pivotTable/applyLegendSet.js
```diff
@@ -28,17 +26,17 @@ const getLegendSet = engine => {
...
-            style.color = legend.color
+            style.color = legendColor
             // TODO: Adapt background color for light text colors
             break
         case LEGEND_DISPLAY_STYLE_FILL:
         default:
-            style.backgroundColor = legend.color
-            if (isColorBright(legend.color)) {
+            style.backgroundColor = legendColor
+            if (isColorBright(legendColor)) {
```
amcgee, line 38: @martinkrulltott this will break when `legendColor` is undefined, because `isColorBright` tries to parse it as a string.

src/modules/pivotTable/applyLegendSet.js
```diff
@@ -58,12 +56,7 @@ export const applyLegendSet = (value, engine) => {
         return {}
     }
 
-    const legend = legendSet.legends.find(legend =>
-        valueFitsLegend(value, legend)
-    )
-    if (!legend) {
```
amcgee, line 64: This short-circuit needs to stay here

amcgee (changes requested):
Haven't done a full review but found a bug in the PT legend processing change, might also need to be addressed in Gauge

src/modules/pivotTable/applyLegendSet.js
martinkrulltott: Yes, it was removed by accident, fixed.

src/modules/pivotTable/applyLegendSet.js
martinkrulltott: Should solved by re-adding the short-circuit from your other comment, but `getColorByValueFromLegendSet` could return an empty string if you prefer?

amcgee: I think that would also break unless we update the `isColorBright` implementation to accept non-color strings, so the contract for this function should be that it always accepts a color string.

martinkrulltott:
@janhenrikoverland  / @cooper-joe  did you want to change the default (blue #87B5E7) color now that we have custom colors as well?

cooper-joe:
> @janhenrikoverland / @cooper-joe did you want to change the default (blue #87B5E7) color now that we have custom colors as well?

No, I don't think we should modify the defaults until we give some robust options for all users to edit those defaults (not just those using legends).

src/modules/pivotTable/applyLegendSet.js
martinkrulltott: Ok, still with the short-circuit re-added it will never currently never reach `isColorBright`, right?

martinkrulltott:
_As per https://github.com/dhis2/data-visualizer-app/pull/703/files this needs to be updated so that the _first_ legendSet is picked regardless of if it matches the `layout.legendSet.id`or not._

⬆️ addressed by https://github.com/dhis2/analytics/pull/315/commits/4dc7e88047a9ea9fd3710948142e88e6187710af

src/modules/legends.js
```diff
@@ -0,0 +1,19 @@
+export const LEGEND_DISPLAY_STRATEGY_BY_DATA_ITEM = 'BY_DATA_ITEM'
```
edoardo, line 1: The first place where I added these is the `LegendDisplayStrategy` and `LegendDisplayStyle` options in DV app.
Since they are used in many places now it makes sense to import them from here.
Once this is merged I can update the option components in DV app.

src/visualizations/config/adapters/dhis_highcharts/series/gauge.js
```diff
@@ -1,9 +1,9 @@
+import { getColorByValueFromLegendSet, LEGEND_DISPLAY_STYLE_TEXT } from "../../../../../modules/legends"
+
 const DEFAULT_FONT_SIZE = '34px'
 const DASHBOARD_FONT_SIZE = '24px'
 
-export default function(series, dashboard) {
-    const fontSize = dashboard ? DASHBOARD_FONT_SIZE : DEFAULT_FONT_SIZE
-
+export default function(series, layout, legendSet, dashboard) {
```
janhenrikoverland: Change the arg order to avoid breaking change.

src/modules/legends.js
```diff
@@ -0,0 +1,19 @@
...
+    }
+    return legendSets.find(legendSet => legendSet.id === id)
+}
+
+export const getColorByValueFromLegendSet = (legendSet, value) => {
+    const legend = legendSet.legends.find(
+        legend => value >= legend.startValue && value < legend.endValue // TODO: Confirm inclusive/exclusive bounds
+    )
+    return legend ? legend.color : undefined
+}
```
janhenrikoverland, line 12: return legend && legend.color

janhenrikoverland (approved):

src/modules/pivotTable/applyLegendSet.js
```diff
@@ -28,17 +26,17 @@ const getLegendSet = engine => {
...
 }
 
-const buildStyleObject = (legend, engine) => {
+const buildStyleObject = (legendColor, engine) => {
     const style = {}
     switch (engine.visualization.legendDisplayStyle) {
         case LEGEND_DISPLAY_STYLE_TEXT:
-            style.color = legend.color
+            style.color = legendColor
             // TODO: Adapt background color for light text colors
```
amcgee: Mind removing this comment for me here?  We aren't going to do this

amcgee (approved):
LGTM

PR #318, opened 2020-02-24
feat: support by-dx legends, non-sum aggtypes, sortable columns, and ou hierarchies

amcgee:
More testing is needed for aggtypes like `count`, `variance`, `avg by org unit`, `std dev`, etc. (I *think* they should work) but I've tried to match the previous engine's behavior from `d2-analysis` for now, so it should at least be consistent.

`PivotTableEngine.js` is getting pretty large (800 lines) and unwieldy, will take a pass at re-factoring it soon.


src/modules/pivotTable/PivotTableEngine.js
```diff
@@ -103,6 +119,23 @@ const buildDimensionLookup = (visualization, metadata, headers) => {
         return out
     }, {})
 
+    const primaryDimensions = {
+        data: getPrimaryDimension(metadata, 'dx'),
+        period: getPrimaryDimension(metadata, 'pe'),
+        orgUnit: getPrimaryDimension(metadata, 'ou'),
```
jenniferarnesen: Is it worth using the constants for 'dx', 'pe', 'ou' that are defined here: modules/predefinedDimensions.js?

jenniferarnesen:
This PR is quite big and thus difficult to say much about. I guess at this point it is hands on testing and exposure that is needed, so I would say - get it merged ASAP to get people trying it out.

When there is time to go back and revisit after the release, my suggestions would be refactoring the PivotTableEngine (as Austin has already mentioned), add unit tests at least for the modules, and possibly some code documentation of PivotTableEngine.

src/modules/pivotTable/PivotTableEngine.js
amcgee: Sure, will do

amcgee: Done

amcgee:
Thanks @jenniferarnesen !  I completely agree we should get it in for testing (and we can iterate if there are any remaining bugs to track down).  If @janhenrikoverland is OK with that, can we get it in tonight so we can look at #326  #327 and #328 which depend on this?

jenniferarnesen (approved):

janhenrikoverland:
+1

PR #319, opened 2020-02-24
fix: loading spinner in ou dialog DHIS2-8360

edoardo:
https://jira.dhis2.org/browse/DHIS2-8360

Screen recording before:
spinner-before

and after:
spinner-after

Note: the buttons are different because I took the screen recording from both dashboards and DV.


janhenrikoverland (approved):

martinkrulltott (approved):
LGTM. Great with the screenshots btw.

PR #332, opened 2020-03-02
fix: reflow pivot table when interpretations panel is toggled

edoardo:
Probably not the prettiest solution, but when I noticed that disabling
the width via DevTools caused the table to reflow and the panel to show,
I thought to set the width to 0 when the interpretations panel toggles
and the resize observer will then trigger and get the correct size
presumably because the panel has then a chance to show and the flex to
work.

### DIFF
--- a/src/components/PivotTable/PivotTable.js
+++ b/src/components/PivotTable/PivotTable.js
@@ -16,9 +16,9 @@ import { PivotTableTitleRow } from './PivotTableTitleRow'
 import { PivotTableEmptyRow } from './PivotTableEmptyRow'
 import getFilterText from '../../visualizations/util/getFilterText'
 
-const PivotTable = ({ visualization, data, legendSets, renderId }) => {
+const PivotTable = ({ visualization, data, legendSets, renderCounter }) => {
     const containerRef = useRef(undefined)
-    const { width, height } = useParentSize(containerRef, renderId)
+    const { width, height } = useParentSize(containerRef, renderCounter)
     const [sortBy, setSortBy] = useState(null)
 
     const engine = useMemo(
@@ -123,7 +123,7 @@ PivotTable.propTypes = {
     data: PropTypes.object.isRequired,
     visualization: PropTypes.object.isRequired,
     legendSets: PropTypes.arrayOf(PropTypes.object),
-    renderId: PropTypes.string,
+    renderCounter: PropTypes.number,
 }
 
 export default PivotTable
--- a/src/components/PivotTable/PivotTable.js
+++ b/src/components/PivotTable/PivotTable.js
@@ -16,9 +16,9 @@ import { PivotTableTitleRow } from './PivotTableTitleRow'
 import { PivotTableEmptyRow } from './PivotTableEmptyRow'
 import getFilterText from '../../visualizations/util/getFilterText'
 
-const PivotTable = ({ visualization, data, legendSets }) => {
+const PivotTable = ({ visualization, data, legendSets, renderId }) => {
     const containerRef = useRef(undefined)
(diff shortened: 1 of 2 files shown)


src/components/PivotTable/PivotTable.js
```diff
@@ -123,6 +123,7 @@ PivotTable.propTypes = {
     data: PropTypes.object.isRequired,
     visualization: PropTypes.object.isRequired,
     legendSets: PropTypes.arrayOf(PropTypes.object),
+    renderId: PropTypes.string,
```
amcgee: Isn't `renderId` a number?

src/modules/pivotTable/useParentSize.js
```diff
@@ -20,13 +21,18 @@ export const useParentSize = (
                 height: el.clientHeight,
             })
         }
+
         onResize(el)
 
+        if (renderId) {
+            setSize({ width: 0 })
```
amcgee: Could we set width to undefined instead?

src/modules/pivotTable/useParentSize.js
```diff
@@ -20,13 +21,18 @@ export const useParentSize = (
                 height: el.clientHeight,
             })
         }
+
         onResize(el)
 
+        if (renderId) {
```
amcgee: Do we need to check `renderId` at all here?  It should also work to do this on the first `effect` call, so long as `renderId` is kept in the dependency array

amcgee (commented):
Nice work!  Agree there's probably an easier way to do this, but it seems reasonable for the moment (and minimally intrusive).  Left a few comments.

src/components/PivotTable/PivotTable.js
edoardo: Yes, forgot to fix that one.

src/modules/pivotTable/useParentSize.js
edoardo: `renderId` is `null` in the first render when the interpretations button has not been clicked yet; this was to avoid resetting the width in that case.

src/modules/pivotTable/useParentSize.js
edoardo: Sure. Tested and it works fine.

src/modules/pivotTable/useParentSize.js
amcgee: If it is `null` on first render is it `0` or `1` on first update?  If `0` this won't trigger an update until the third render...

src/modules/pivotTable/useParentSize.js
```diff
@@ -20,13 +21,18 @@ export const useParentSize = (
                 height: el.clientHeight,
             })
         }
+
         onResize(el)
 
+        if (renderCounter) {
+            setSize({ width: undefined })
```
amcgee: Hmm... actually, since this replaces the entire `state` object why are we overriding only `width` explicitly?  (`height` is `undefined` by default).  A bit more explicit might be `setSize({})` or `setSize({ width: undefined, height: undefined })` or `setSize({ width: 0, height: 0 })` (I missed before that we're checking for `!width || !height` below, so `0` has the same effect as `undefined` and it's nice to match the `initialSize`. default, probably?)

amcgee (approved):
LGTM - one optional comment

src/modules/pivotTable/useParentSize.js
edoardo: Starts with `1`.

src/modules/pivotTable/useParentSize.js
edoardo: Yes, I'll go with the initial state for both width and height.

PR #335, opened 2020-03-03
fix: wrong export for default vis type

edoardo:
`defaultChartType` was the old export name.

`defaultVisType` is already in use in DV, but without the correct export it was `undefined`.


martinkrulltott (approved):
LGTM

PR #336, opened 2020-03-03
fix: adaptive clipping

amcgee:
This implements adaptive clipping (for columns only, currently) which supports fit-to-content column widths even for huge tables.  All tables are clipped.  This also allows us to better support different font sizes and display densities in clipped tables.

I've set the max-width to `360px` at @cooper-joe 's suggestion, but haven't yet implemented text-wrapping (adaptive row heights) so for now really really long text cells will overflow with a `...` overflow indicator.

This is an alternative to #331 and I think a superior solution, though less simple.

I'm leaving this as a *draft* for now since ~it relies on #330 (merged) and~ I'm going to refactor some of the newly-introduced Engine code, but it's currently ready for a first pass of reviews and testing!


janhenrikoverland:
https://jira.dhis2.org/browse/DHIS2-8397

Seems to be fixed in the previous commit.

janhenrikoverland (approved):

PR #337, opened 2020-03-04
fix: bump @dhis2/d2-ui-org-unit-dialog@6.5.9 to fix DHIS2-6562

edoardo:
In the data visualizer app, for the org unit dimension selector dialog, for the org unit tree. Currently there is a function for selecting all org unit children which is accessed by right-clicking on an org unit. Ticket: https://jira.dhis2.org/browse/DHIS2-6562

This function is labelled **"Select children"**, but should rather be: **"Select all org units below"**.

- Bumps d2-ui-org-unit-dialog to v6.5.9, which contains https://github.com/dhis2/d2-ui/pull/573 that changes the name of the label.

Screenshot 2020-03-04 at 10 32 22

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,7 @@
     },
     "dependencies": {
         "@dhis2/d2-i18n": "^1.0.4",
-        "@dhis2/d2-ui-org-unit-dialog": "^6.5.6",
+        "@dhis2/d2-ui-org-unit-dialog": "^6.5.9",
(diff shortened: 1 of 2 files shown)


martinkrulltott (changes requested):
Not approved, needs a better description

martinkrulltott (approved):
Tested locally, LGTM!

image

PR #341, opened 2020-03-05
fix: don't assume data dimension is a column or row, could be filter

amcgee:
This fixes support for Pivot Table Visualization configurations in which the `dx` dimension appears in `filters`, rather than `rows` or `columns`.

It also fixes cell max-widths for narrative (text) values, and adds several Storybook stories for testing these features.

I have also added here 3 stories which intentionally break, to demonstrate and validate other bugs which will soon be fixed (single-axis tables and Data Element Group Sets)

Finally, this factors out several constant strings to the `pivotTableConstants.js` file


janhenrikoverland (approved):
LGTM

PR #345, opened 2020-03-09
fix: make the gauge use more viewport space (DHIS2-8412)

edoardo:
The value for `pane.size` passed to Highcharts controls the amount of viewport that the gauge uses.
Increased to 140% as seen on Highcharts demo examples for DV and 100% (was 90%) for dashboards.

Also increase the font size used for the value.
It's not possible to find a fixed font-size that works well with all lengths of values.
Depending on the number of digits and the size of the container, there is a limit on which values can fit under the arch.

Fixes: https://jira.dhis2.org/browse/DHIS2-8412

Before DV:
Screenshot 2020-03-09 at 14 34 06

After DV:
Screenshot 2020-03-10 at 13 53 10

Before dashboards:
Screenshot 2020-03-10 at 13 59 48

After dashboards:
Screenshot 2020-03-10 at 14 00 13

### DIFF
--- a/src/visualizations/config/adapters/dhis_highcharts/index.js
+++ b/src/visualizations/config/adapters/dhis_highcharts/index.js
@@ -88,7 +88,7 @@ export default function({ store, layout, el, extraConfig, extraOptions }) {
         legend: getLegend(_layout, _extraOptions.dashboard),
 
         // pane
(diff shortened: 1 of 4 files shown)


janhenrikoverland (approved):
LGTM

PR #346, opened 2020-03-09
fix: correct support for data element group sets as dx dimension

amcgee:
Previously, `valueType` and `totalAggregationType` were not correctly collected from Data Element Group Set dimensions, which can act as  the `dx` dimension.  This corrects that lookup.


src/modules/predefinedDimensions.js
```diff
@@ -10,6 +10,8 @@ export const DIMENSION_ID_ORGUNIT = 'ou'
 export const DIMENSION_ID_ASSIGNED_CATEGORIES = 'co'
 export const DIMENSION_PROP_NO_ITEMS = 'noItems'
 
+export const DIMENSION_TYPE_DATA_ELEMENT_GROUP_SET = 'DATA_ELEMENT_GROUP_SET'
```
martinkrulltott: This should probably be placed somewhere else as this file only handles _actual dimensions_ whereas `DIMENSION_TYPE_DATA_ELEMENT_GROUP_SET` is a _dimension type_.

martinkrulltott: https://github.com/dhis2/analytics/blob/7c8658f5b2da1611561616a1c75bd6a87d526c24/src/modules/layout/dimension.js ?
Or a new file in `src/modules` called `dimensionTypes.js`?

martinkrulltott (approved):
LGTM, just a tiny comment.

src/modules/predefinedDimensions.js
amcgee: It's right below another constant that's tangentially related to dimensions, but could move it to `pivotTableConstants` if this is important.

amcgee: @martinkrulltott @janhenrikoverland I still think this is OK here, let me know if I should move it to `pivotTableConstants`

martinkrulltott: `predefinedDimensions.js` contains predefined dimensions. This is neither predefined nor a dimension. It's a dimension type. Should it be there? No. Can it still be there? Sure 🤷‍♂ 🙂

janhenrikoverland: We don't have constants for any other dynamic dimension types (orgunit group sets, categories etc), so `pivotTableConstants` makes sense for now I think.

amcgee: OK, done

PR #347, opened 2020-03-09
fix: support single-axis tables

amcgee:
This allows table without either a row or column axis to be rendered, and should support visualization type-switching from gauge or single value (specifically on the dashboard)

### DIFF
--- a/src/modules/pivotTable/PivotTableEngine.js
+++ b/src/modules/pivotTable/PivotTableEngine.js
@@ -501,12 +501,15 @@ export class PivotTableEngine {
 
     getDependantTotalCells({ row, column }) {
         const rowSubtotalSize = this.dimensionLookup.columns[0]?.size + 1
-        const rowSubtotal = this.doRowSubtotals && {
-            row,
-            column:
-                Math.ceil((column + 1) / rowSubtotalSize) * rowSubtotalSize - 1,
-            size: rowSubtotalSize - 1,
-        }
+        const rowSubtotal = rowSubtotalSize &&
+            this.doRowSubtotals && {
+                row,
+                column:
+                    Math.ceil((column + 1) / rowSubtotalSize) *
+                        rowSubtotalSize -
+                    1,
+                size: rowSubtotalSize - 1,
+            }
         const rowSubtotalColumnTotal = this.doColumnSubtotals &&
             this.doRowTotals && {
                 row: this.dataHeight - 1,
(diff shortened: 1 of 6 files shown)


jenniferarnesen (approved):
Works well.

Note that dashboards currently has removed the "View as" menu for SV, Pie, Gauge and YoY. So that will need to be changed (maybe need to talk to Jan or Scott?) in order for this commit to have any effect.

amcgee:
@jennifer thanks - true that you can’t switch from pivot or gauge any more, however there could be single-axis tables saved in the old PT app that would have crashed in the new plugin before this change, so it’s important regardless

src/modules/pivotTable/PivotTableEngine.js
```diff
@@ -498,21 +500,21 @@ export class PivotTableEngine {
     }
 
     getDependantTotalCells({ row, column }) {
-        const rowSubtotalSize = this.dimensionLookup.columns[0].size + 1
+        const rowSubtotalSize = this.dimensionLookup.columns[0]?.size + 1
```
martinkrulltott, line 503: Just curious, how does the `?` work here (and what's `size` btw)?

edoardo: `?.` is this operator: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

amcgee: The `?` is [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining).  if `columns[0]` is undefined the expression will return undefined, otherwise it will de-reference the `.size` property.  This prevents a runtime exception.  It's similar to `(columns[0] !== null && columns[0] !== undefined) ? columns[0].size : undefined`

Size is the span in columns or rows of **each item** in the dimension.  Basically, if it's the "outermost" dimension it's large and if it's the "innermost" it's always 1.  There's a function called `addSize` which calculates this.

janhenrikoverland: Check the output here: https://babeljs.io/en/repl#?browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=MYewdgzgLgBCBGArGBeGBvGBLAJgLhgEYYBfAKDNEljFTiQH4A6XAbiA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=es2015%2Creact%2Cstage-2&prettier=false&targets=&version=7.8.7

martinkrulltott: Thanks! Thats awesome, been waiting for something like this for ages 🚀

martinkrulltott:
Once merged, please add the new Analytics version to https://github.com/dhis2/data-visualizer-app/pull/789 and test the full flow for creating a new AO without row or col.

martinkrulltott (approved):

PR #348, opened 2020-03-09
fix: sort OU hierarchy items

amcgee:
Sort OrgUnit dimension items alphabetically


jenniferarnesen:
Is there a Jira issue for this?

amcgee:
Not that I know of, but this was the behavior of the previous PT app.  The full solution would be to render hierarchy as cascading rows or columns, by ran out of time to implement that before the freeze

src/modules/pivotTable/PivotTableEngine.js
```diff
@@ -168,7 +158,6 @@ const buildDimensionLookup = (visualization, metadata, headers) => {
         rowHeaders,
         columnHeaders,
         dataHeaders,
-        primaryDimensions,
```
martinkrulltott: Would it be safe to remove the other return values as well? Except for `columns` and `rows` which seems to be used here https://github.com/dhis2/analytics/blob/4417a91e1debc496c2c3bc26c03839901316d213/src/modules/pivotTable/PivotTableEngine.js#L265

jenniferarnesen (approved):

PR #349, opened 2020-03-10
fix: correct calculation of row subtotal column totals

amcgee:
There was a minor bug in column total calculations which skipped row subtotal columns - this fixes that issue


martinkrulltott (approved):

martinkrulltott:
This release can be added to https://github.com/dhis2/data-visualizer-app/pull/789 too 🎉

PR #353, opened 2020-03-10
fix: full title row span in single-axis, refactor header counts

amcgee:
This fixes an issue @martinkrulltott noticed with title rows having an off-by-one error in single-axis (column only) configurations.  It also refactors things so that we're not duplicating the `columns.length || 1` logic all over the place.

Nice side-effect - no-column pivot tables can be sorted now!


janhenrikoverland:
@amcgee RSA tables (row single axis) have a row of empty dimension cells when dimension labels are turned off. Could be removed I guess.
Screenshot from 2020-03-10 22-09-01

CSA tables look fine.

janhenrikoverland:
I got this on a CSA or RSA table. And I selected more than 1 item in that dimension. That's all I remember right now. And I'm not able to reproduce it at the moment.

Screenshot from 2020-03-10 22-13-08

amcgee:
@janhenrikoverland @jenniferarnesen I see the RSA empty header issue, fixing now.  I couldn't reproduce the error despite trying quite a few different things - multiple axis levels seem to work great.

amcgee:
@janhenrikoverland approved this via Slack, merging now and hopefully he can give his official approval tomorrow

jenniferarnesen (approved):

PR #357, opened 2020-03-12
fix: correctly finalize combined subtotals and totals

amcgee:
There was a bug in the finalization (average and column width calculation) for combined subtotals - this fixes that issue

Before:
Screen Shot 2020-03-12 at 1 39 25 PM

After:
Screen Shot 2020-03-12 at 1 37 51 PM

### DIFF
--- a/src/components/PivotTable/PivotTableValueCell.js
+++ b/src/components/PivotTable/PivotTableValueCell.js
@@ -18,7 +18,7 @@ export const PivotTableValueCell = ({ row, column }) => {
     })
 
     if (!cellContent || cellContent.empty) {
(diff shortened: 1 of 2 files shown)


src/components/PivotTable/PivotTableValueCell.js
```diff
@@ -18,7 +18,7 @@ export const PivotTableValueCell = ({ row, column }) => {
     })
 
     if (!cellContent || cellContent.empty) {
-        return <PivotTableEmptyCell classes={cellContent?.cellType} />
```
amcgee, line 21: This is technically unrelated, but was causing a propTypes warning so I went ahead and fixed it

janhenrikoverland (approved):
LGTM

PR #358, opened 2020-03-13
fix: fix gauge size to avoid clipping (DHIS2-8412)

edoardo:
Cannot go over 100% as it causes a bad clipping when the container size is smaller than the minimum width of the chart.
Other vis types resize only to a certain point, after that no resize happens.
With 100% size for the pane the same happens.

JIRA: https://jira.dhis2.org/browse/DHIS2-8412

Example of bad clipping.
Before:
Screenshot 2020-03-13 at 14 51 07

After:
Screenshot 2020-03-13 at 14 52 16

### DIFF
--- a/src/visualizations/config/adapters/dhis_highcharts/pane/gauge.js
+++ b/src/visualizations/config/adapters/dhis_highcharts/pane/gauge.js
@@ -1,10 +1,9 @@
-const DEFAULT_PANE_SIZE = '140%'
-const DASHBOARD_PANE_SIZE = '100%'
+const DEFAULT_PANE_SIZE = '100%'
 
-export default function(dashboard) {
+export default function() {
     return {
         center: ['50%', '85%'],
-        size: dashboard ? DASHBOARD_PANE_SIZE : DEFAULT_PANE_SIZE,
+        size: DEFAULT_PANE_SIZE,
         startAngle: -90,
         endAngle: 90,
         background: {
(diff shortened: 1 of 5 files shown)


src/visualizations/config/adapters/dhis_highcharts/pane/gauge.js
```diff
@@ -1,10 +1,9 @@
-const DEFAULT_PANE_SIZE = '140%'
-const DASHBOARD_PANE_SIZE = '100%'
+const DEFAULT_PANE_SIZE = '100%'
 
-export default function(dashboard) {
```
martinkrulltott, line 4: To reflect the removed param this should be updated as well: https://github.com/dhis2/analytics/blob/a1e3fe5cdedd0e4f4d73528f7c5b461b033fb6a9/src/visualizations/config/adapters/dhis_highcharts/pane/index.js#L7

src/visualizations/config/adapters/dhis_highcharts/series/gauge.js
```diff
@@ -17,9 +16,7 @@ export default function(series, dashboard, layout, legendSet) {
                 borderWidth: 0,
                 verticalAlign: 'bottom',
                 style: {
-                    fontSize: dashboard
```
martinkrulltott, line 20: As this is the only place `dashboard` is used the param should be removed from the function and this should stop passing it https://github.com/dhis2/analytics/blob/a1e3fe5cdedd0e4f4d73528f7c5b461b033fb6a9/src/visualizations/config/adapters/dhis_highcharts/series/index.js#L135

martinkrulltott (approved):
Approved once the 2 comments are addressed.

PR #368, opened 2020-03-16
fix: left-align org-unit hierarchies in row headers

amcgee:
This left-aligns the OU hierarchy labels when they appear in the **row axis** - they are still centered when in the column axis.

Before
Screen Shot 2020-03-16 at 3 57 31 PM

After
Screen Shot 2020-03-16 at 3 56 56 PM


cooper-joe (approved):
LGTM

janhenrikoverland (approved):

PR #369, opened 2020-03-16
fix: DV chart title should not use 'in' when no ou selected (DHIS2-8473)

jenniferarnesen:
fixes https://jira.dhis2.org/browse/DHIS2-8473

Case: ou is in filter, and group and/or level is selected, but no ou from the hierarchy is selected
Expected result: the title should use correct language

With ou selected:
image

Without ou selected: 
image


janhenrikoverland (approved):
LGTM

PR #370, opened 2020-03-18
fix: sort data when sorting org unit hierarchy headers (DHIS2-8486)

amcgee:
This fixes a critical issue @janhenrikoverland found in Pivot Tables with the `Show Org Unit Hierarchy` option.  The row or column headers would be sorted alphabetically, but the data would not.

No hierarchy shown (unsorted)
Screen Shot 2020-03-18 at 11 51 40 AM

Before
Screen Shot 2020-03-18 at 11 52 20 AM

After
Screen Shot 2020-03-18 at 11 51 35 AM


janhenrikoverland (approved):
LGTM

PR #371, opened 2020-03-19
fix: skip superfluous data rows which don't match table headers (DHIS2-8497)

amcgee:
This fixes a crash which would occur if the server responded with data not matching the row or column headers of the current table.  Thanks @janhenrikoverland for finding the bug, which can be reproduced here - https://debug.dhis2.org/2.34dev/dhis-web-data-visualizer/index.html#/AOphCIgmjeN


amcgee:
This fixes [DHIS2-8497](https://jira.dhis2.org/browse/DHIS2-8497)

janhenrikoverland (approved):
LGTM

PR #372, opened 2020-03-19
fix: use interpolation to give context for translators [DHIS2-8473]

jenniferarnesen:
Use i18n interpolation to provide context for translators.

No functional changes

### DIFF
--- a/src/modules/getOuLevelAndGroupText.js
+++ b/src/modules/getOuLevelAndGroupText.js
@@ -46,7 +46,11 @@ const getLevelAndGroupText = (items, metaData, isLevel) => {
         const lastOuName = getNameFromMetadata(
             ouIdHelper.removePrefix(lastItem.id)
         )
-        allDynamicOuNames = `${dynamicOuNames} ${i18n.t('and')} ${lastOuName}`
+        // allDynamicOuNames = `${dynamicOuNames} ${i18n.t('and')} ${lastOuName}`
+        allDynamicOuNames = i18n.t('{{dynamicOuNames}} and {{lastOuName}}', {
+            dynamicOuNames,
+            lastOuName,
+        })
     } else {
         allDynamicOuNames = dynamicOuNames
     }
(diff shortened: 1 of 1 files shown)


src/modules/getOuLevelAndGroupText.js
```diff
@@ -46,7 +46,11 @@ const getLevelAndGroupText = (items, metaData, isLevel) => {
         const lastOuName = getNameFromMetadata(
             ouIdHelper.removePrefix(lastItem.id)
         )
-        allDynamicOuNames = `${dynamicOuNames} ${i18n.t('and')} ${lastOuName}`
+        // allDynamicOuNames = `${dynamicOuNames} ${i18n.t('and')} ${lastOuName}`
```
martinkrulltott: Remove?

src/modules/getOuLevelAndGroupText.js
```diff
@@ -61,12 +65,26 @@ const getLevelAndGroupText = (items, metaData, isLevel) => {
...
+            })
+        }
+
+        return i18n.t('{{allDynamicOuNames}} groups', {
+            allDynamicOuNames,
+        })
+    }
+
+    if (isLevel) {
+        return i18n.t('{{allDynamicOuNames}} levels in {{staticOuNames}}', {
```
martinkrulltott: Great update with the interpolations. Could we possibly use variable names that are easier to understand for the translators?

src/modules/getOuLevelAndGroupText.js
```diff
@@ -61,12 +65,26 @@ const getLevelAndGroupText = (items, metaData, isLevel) => {
...
+            allDynamicOuNames,
+            staticOuNames,
+        })
     }
 
-    return `${allDynamicOuNames} ${
-        isLevel ? i18n.t('levels in') : i18n.t('groups in')
-    } ${staticOuNames}`
+    return i18n.t('{{allDynamicOuNames}} groups in {{staticOuNames}}', {
```
martinkrulltott: Not sure if we have a best practice for this, but generally I try to avoid multiple return statements. Now there's 4, and they all return the same thing, so might be better to change to a `let` that's returned in the end for readability?

martinkrulltott (approved):
Approved, minor comments

PR #375, opened 2020-03-25
feat: move period selector to analytics [TECH-333]

jenniferarnesen:


jenniferarnesen:
It might not be necessary to move the period-selector after all. Will wait for ui-core/widgets to finish up some components. Then reevaluate.

PR #376, opened 2020-03-25
fix: build library with DHIS2 App Platform, fix static analysis bugs

amcgee:
This is a largely non-functional change to build with the app platform instead of a custom babel config.  Storybook, tests, and production builds should all work as before.

The bulk of the platform port is in 15bcbe5.

In 9a0aeda I applied the dhis2 eslint and prettier rules to the entire codebase, notably fixing a few functional bugs like [a missing import](https://github.com/dhis2/analytics/compare/chore/app-platform?expand=1#diff-55bdeaa520f3ccaf6abd00a25e8a16f2R1) and lots of `==` instead of `===` that eslint found.

Previously `eslint` was disabled for the `/src/visualizations` directory (presumably because it was legacy code) but this was actually hiding real issues.  I turned `eslint` back on for that directory and fixed all the errors it raised, but turned off the `max-params` rule only for that directory to avoid a large refactor of function interfaces.

A few notes:
- Stories moved from `/stories` to `/src/__demo__` (required for our babel configs)
- i18n strings are extracted to `/i18n/en.pot` and code is generated in `/src/locales`, but the strings aren't loaded yet (need to figure out the best place to do that...).  I did add `/src/locales/index.js` to the package.json sideEffect array since it operates on a singleton (unfortunately...)
- The build process now generates both `cjs` and `es` output, and both use `rollup` so everything ends up in one file.  Tree shaking should work, but we'll want to confirm that.


src/visualizations/config/adapters/dhis_highcharts/addTrendLines.js
```diff
@@ -77,10 +73,11 @@ function getRegressionObj(data, regressionType) {
     // - decimalPlaces (default = 2)
 
     let regression
-    let regressionTypeOptions = {}
+    const regressionTypeOptions = {}
 
     switch (regressionType) {
         case 'LINEAR':
```
jenniferarnesen: I don't think I've seen `default` other than last in a switch. Could it be moved to the bottom, just for convention's sake? Actually, I am a little nervous about some of these functional changes (this one, and the == to === changes) and any effects they might have in the apps. I would almost prefer to make individual eslint exceptions (output warning instead of error), and then address them individually.

.eslintignore
```diff
@@ -1,8 +1,7 @@
...
 /config/*
 /docs/*
-/stories/*
+/src/__demo__/data/*
 /storybook-static/*
 /build/*
-/coverage/*
-/src/visualizations/*
\ No newline at end of file
+/coverage/*
```
jenniferarnesen: Quite a few files are missing a newline.

src/visualizations/config/adapters/dhis_highcharts/addTrendLines.js
```diff
@@ -212,114 +209,114 @@ function linear(data, decimalPlaces = 2) {
...
-        Math.round(B * 100) / 100 +
-        ' ln(x)'
-
-    return {
-        equation: [A, B],
-        points: results,
-        string: string,
-    }
-}
+// function logarithmic(data) {
```
jenniferarnesen: Maybe add a comment stating that these functions are currently not in use but may be needed in the future. @janhenrikoverland @edoardo Is there a good reason to hang on to this code?

src/visualizations/config/adapters/dhis_highcharts/addTrendLines.js
amcgee: This is a functional change, true, but the current behavior in the default case (which is hopefully never hit) is to cause an access violation (cannot read property x of undefined) a few lines later, so this is definitely a safe improvement I think.

Which == to === changes do you have issues with?  I think they were mostly incorrect before this change.

src/visualizations/config/adapters/dhis_highcharts/addTrendLines.js
amcgee: The code referenced undefined variables, so it has to be removed, fixed, or commented.  I would be happy to remove it.

src/visualizations/config/adapters/dhis_highcharts/addTrendLines.js
jenniferarnesen: I was just thinking generally about risk of going to a stricter equality check. Might mean that other functions using this were being lazy about types, and suddenly those functions get unexpected results.

jenniferarnesen:
I got this when yarn linking analytics in data-visualizer: 
image

src/visualizations/config/adapters/dhis_highcharts/addTrendLines.js
edoardo: Not in use so I would remove it.
@janhenrikoverland was it used in the past at some point?
This would need support in the backend for saving the correct value for the option, which does not seem to be there currently: https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/RegressionType.java.
Safe to remove then?

edoardo: Same thing for `power` below.

edoardo (approved):
LGTM.

I understand the concern about the undefined checks.
This library is currently only used in DV, where we are currently working a lot, so hopefully if there is a problem with one of the checks, we would notice.

src/visualizations/config/adapters/dhis_highcharts/addTrendLines.js
janhenrikoverland: @amcgee `logarithmic` and `power` are okay to remove.

amcgee:
I added some missing EOF newlines, removed the commented-out code, and rearranged the select case order so the default (linear regression) comes last.

src/visualizations/config/adapters/dhis_highcharts/addTrendLines.js
amcgee: Removed

.eslintignore
amcgee: Fixed

src/visualizations/config/adapters/dhis_highcharts/addTrendLines.js
amcgee: Rearranged the default case

amcgee: Discussed with @edoardo and in slack - the `=== undefined` changes should be OK

janhenrikoverland (approved):

jenniferarnesen (approved):

PR #388, opened 2020-04-17
fix: move d2-i18n to peer dependency [WIP]

jenniferarnesen:
BREAKING CHANGE: apps that use components having i18n must have d2-i18n as a dependency

Part of fix for https://jira.dhis2.org/browse/DHIS2-8638

WIP:
* discuss: if an app is using the analytics library but not using any components that require d2-i18n, then the app does not actually need d2-i18n
* waiting on d2-ui components to be published

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,12 +37,12 @@
         "typeface-roboto": "^0.0.54"
     },
     "peerDependencies": {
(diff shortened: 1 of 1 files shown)


jenniferarnesen:
Not sure this is necessary

PR #389, opened 2020-04-19
chore: move period selector to analytics

jenniferarnesen:
Due to continuing issues with the circular dependency between d2-ui and analytics, move PeriodSelector to analytics.

No functional changes. Only a few code changes, related to moving from css to css modules.

Note: we should discuss whether it makes sense to do this now. Looks like a new `transfer` component is nearing completion (would replace the ItemSelector that is the cause of the circular dependency). This circular dependency is blocking finishing up the DV translations. So, what is the timeline for getting the translation functionality into DV v33? If it is urgent, then we should go forward with this. Otherwise, it can probably wait until we rebuild the Period Selector with the new component.


jenniferarnesen:
Will redo this a bit differently.

PR #394, opened 2020-04-22
chore: move PeriodSelector component to analytics [v2.4.10] [DHIS2-8638]

jenniferarnesen:
Part of fix for [DHIS2-8638](https://jira.dhis2.org/browse/DHIS2-8638)

The PeriodSelector component has been part of a circular dependency between d2-ui and analytics. This has created a lot of headaches with upgrading and other at-times strange errors and behaviour. In the future, PeriodSelector will be part of ui-widgets, but until that can happen, Period Selector is being moved to analytics to resolve the circular dependency issues

* No functional changes
* Component can (and should) be viewed and tested in Storybook (yarn start)

Code is nearly identical to what was in d2-ui. Only changes were related to converting the css file to styled-jsx. So for review, I recommend only checking the style files and trying the component in storybook.

### DIFF
--- a/src/components/PeriodSelector/styles/InputLabel.style.js
+++ b/src/components/PeriodSelector/styles/InputLabel.style.js
@@ -1,0 +1,8 @@
+import css from 'styled-jsx/css'
+
+export default css`
+    .input-label {
+        color: #595959;
+        white-space: nowrap;
+    }
+`
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@
 # production
 /build
 /dist
+/src/locales
 
 # misc
 .DS_Store
--- a/package.json
+++ b/package.json
@@ -61,7 +61,6 @@
     "dependencies": {
         "@dhis2/d2-i18n": "^1.0.4",
         "@dhis2/d2-ui-org-unit-dialog": "^6.3.0",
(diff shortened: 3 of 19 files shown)


src/components/PeriodSelector/styles/InputLabel.style.js
```diff
@@ -0,0 +1,8 @@
+import css from 'styled-jsx/css'
+
+export default css`
+    .input-label {
+        color: #595959;
```
martinkrulltott, line 5: `ui-core` colors?

martinkrulltott (approved):
As per the instructions, I only briefly looked over the `PeriodSelector/*.js` files, as I expect these to just be copied in (plus this will most likely all be replaced for 2.35 anyway (However for the backports I suggest giving the PS some extra testing attention)). CSS and UI seems to LGTM 🚀 

-----
Before:
image

After: 
image

src/components/PeriodSelector/styles/InputLabel.style.js
jenniferarnesen: It is not a ui-core color, but maybe it could be.  (The ui-core greys have a blue-ish tone, while this one has a red-ish tone).

jenniferarnesen: I'll address it when PeriodSelector is refactored.

PR #400, opened 2020-04-27
fix: generate translations

jenniferarnesen:
v33 of our apps need working translations from analytics.

Changes include:
- run i18n extract and generate during the build
- import src/locales in the root index file so the translations get loaded by the apps
- make @dhis2/d2-i18n a peer and dev dependency: peer for the apps, and dev for storybook

NOTE: Before this PR can be merged and published as v2.4.10, a d2-i18n-generate PR needs to be merged and published: https://github.com/dhis2/d2-i18n-generate/pull/7. This is also the reason the build on this PR currently fails.

### DIFF
--- a/i18n/en.pot
+++ b/i18n/en.pot
@@ -1,0 +1,120 @@
+msgid ""
+msgstr ""
+"Project-Id-Version: i18next-conv\n"
+"MIME-Version: 1.0\n"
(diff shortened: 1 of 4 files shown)


amcgee (approved):
LGTM, but use `1.2.0` instead of `1.1.1` for the published  `d2-i18n-generate`

PR #408, opened 2020-05-04
fix: dimension panel search field is missing the 'x' to clear (DHIS2-8790)

martinkrulltott:
Fixes [DHIS2-8790](https://jira.dhis2.org/browse/DHIS2-8790).
Adds the browser default 'x' by passing through the `type` prop (which can be set to `search` to get the 'x'.

Also updates the `@dhis2/ui-core` dep to latest, v4.19.1

image

Relates to: https://github.com/dhis2/data-visualizer-app/pull/948

### DIFF
--- a/src/components/Filter/Filter.js
+++ b/src/components/Filter/Filter.js
@@ -14,10 +14,14 @@ export const Filter = ({ text, onChange, onClear, placeholder, type }) => (
 
 Filter.propTypes = {
     placeholder: PropTypes.string.isRequired,
-    text: PropTypes.string.isRequired,
+    text: PropTypes.string,
     type: PropTypes.string.isRequired,
     onChange: PropTypes.func.isRequired,
     onClear: PropTypes.func.isRequired,
+}
+
+Filter.defaultProps = {
+    type: 'text',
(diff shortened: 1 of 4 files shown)


src/components/Filter/Filter.js
```diff
@@ -2,18 +2,20 @@ import React from 'react'
...
         value={text}
         dense
+        type={type}
     />
 )
 
 Filter.propTypes = {
     placeholder: PropTypes.string.isRequired,
     text: PropTypes.string.isRequired,
+    type: PropTypes.string.isRequired,
```
janhenrikoverland, line 17: Would it make sense to just let it fall back to the html default (text) if not provided instead of making it required?

martinkrulltott: Yes, I believe `ui-core` does this already (since we didn't pass `type` before). But we can do it too. Nevertheless, it shouldn't be `isRequired`. Fixed

edoardo (approved):
LGTM.
As agreed, simplify by removing the `Filter` "wrapper" component and use `ui-core` component directly.

PR #414, opened 2020-05-06
fix: add translation fixes for v2.4

jenniferarnesen:
Part of fix for: https://jira.dhis2.org/browse/DHIS2-8638

All i18n.t statements need to be executed at render time for the translations to be correct.

With Spanish (and Spanish-ish) translation:

Data Dimension:

image

image

image

image

Period dimension:
image

image

Chart title:
image

PR #418, opened 2020-05-08
fix: period selector translations for v2.4

jenniferarnesen:
Part of fix for: https://jira.dhis2.org/browse/DHIS2-8638

Fixes primarily consist of using functions when calling i18n.t().

Currently this branch is based off of another fix branch (for PR https://github.com/dhis2/analytics/pull/414). As soon as that PR is approved and merged to v2.4, then I will update this PR to be towards v2.4

### DIFF
--- a/src/components/PeriodSelector/modules/FixedPeriodsGenerator.js
+++ b/src/components/PeriodSelector/modules/FixedPeriodsGenerator.js
@@ -2,7 +2,6 @@ import i18n from '@dhis2/d2-i18n'
 // generatePeriods config object: { boolean offset, boolean filterFuturePeriods, boolean reversePeriods }
 
 export const MONTHLY = 'Monthly'
-const NUM_MONTHS = 12
 
 const getMonthName = key => {
     const monthNames = [
@@ -286,9 +285,7 @@ function SixMonthlyPeriodType(fnFilter) {
         let period = {}
         period.startDate = `${year}-01-01`
         period.endDate = `${year}-06-30`
-        const monthStart = getMonthName(0)
-        const monthEnd = getMonthName(5)
-        period.name = () => `${monthStart} - ${monthEnd} ${year}`
+        period.name = () => `${getMonthName(0)} - ${getMonthName(5)} ${year}`
         period.iso = `${year}S1`
         period.id = period.iso
         periods.push(period)
@@ -296,9 +293,7 @@ function SixMonthlyPeriodType(fnFilter) {
         period = {}
         period.startDate = `${year}-07-01`
         period.endDate = `${year}-12-31`
-        const monthStart2 = getMonthName(6)
-        const monthEnd2 = getMonthName(11)
-        period.name = () => `${monthStart2} - ${monthEnd2} ${year}`
+        period.name = () => `${getMonthName(6)} - ${getMonthName(11)} ${year}`
         period.iso = `${year}S2`
         period.id = period.iso
         periods.push(period)
@@ -321,9 +316,7 @@ function SixMonthlyAprilPeriodType(fnFilter) {
         let period = {}
         period.startDate = `${year}-04-01`
         period.endDate = `${year}-09-30`
-        const monthStart = getMonthName(3)
-        const monthEnd = getMonthName(8)
-        period.name = () => `${monthStart} - ${monthEnd} ${year}`
+        period.name = () => `${getMonthName(3)} - ${getMonthName(8)} ${year}`
         period.iso = `${year}AprilS1`
         period.id = period.iso
         periods.push(period)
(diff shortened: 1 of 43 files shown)


src/components/PeriodSelector/modules/FixedPeriodsGenerator.js
```diff
@@ -62,7 +84,12 @@ function WeeklyPeriodType(formatYyyyMmDd, weekObj, fnFilter) {
             period.id = period.iso
             date.setDate(date.getDate() + 6)
             period.endDate = formatYyyyMmDd(date)
-            period.name = `Week ${week} - ${period.startDate} - ${period.endDate}`
+
+            const weekNumber = week
```
martinkrulltott, line 87: `const` needed? same for biWeekNumber, monthName etc below

src/components/PeriodSelector/modules/FixedPeriodsGenerator.js
```diff
@@ -251,15 +286,19 @@ function SixMonthlyPeriodType(monthNames, fnFilter) {
...
+        period.name = () => `${monthStart} - ${monthEnd} ${year}`
         period.iso = `${year}S1`
         period.id = period.iso
         periods.push(period)
 
         period = {}
         period.startDate = `${year}-07-01`
         period.endDate = `${year}-12-31`
-        period.name = `${monthNames[6]} - ${monthNames[11]} ${year}`
+        const monthStart2 = getMonthName(6)
```
martinkrulltott: `monthStart2`? 🤷‍♂️

martinkrulltott (commented):
Generally looks good. There's a lot of `const`'s being introduced for single-use purposes. IMHO this makes the code bloated. 

There's also a lot of repetition in the generators (since a lot of the date output formats are the same). Could we use global helper functions to generate these strings? 
For example the format `${monthStart} - ${monthEnd} ${year}` (another example `${monthStart} ${yearStart} - ${monthEnd} ${yearEnd}` ) is used in multiple places. I suggest there's a function `someFnName = (monthStart, monthEnd, year) => '${monthStart} - ${monthEnd} ${year}'`. To keep things DRY and reduce the risk of inconsistency because of a typo or a future update to the formatting.

src/components/PeriodSelector/modules/FixedPeriodsGenerator.js
jenniferarnesen: Yes, it is needed. A new const (`weekNumber`) is needed for each iteration in the loop, otherwise, the name function will use the latest value of `week` when the name function is later executed, which would look like this:

image

Same goes for most of the other `const` vars that I've introduced. I was able to remove some that were just being set to a constant month (e.g., `getMonthName(6)`)

jenniferarnesen: If I don't have a `const monthName`:

image

src/components/PeriodSelector/modules/FixedPeriodsGenerator.js
jenniferarnesen: I was able to remove these constants. So `monthStart2` is gone (yes, I was getting lazy, and tired of this code 🙄 )

jenniferarnesen:
> Generally looks good. There's a lot of `const`'s being introduced for single-use purposes. IMHO this makes the code bloated.

Addressed in a separate comment (TLDR; I never wanted all these constants, but most of them were necessary due to name() function being a closure).

> 
> There's also a lot of repetition in the generators (since a lot of the date output formats are the same). Could we use global helper functions to generate these strings?
> For example the format `${monthStart} - ${monthEnd} ${year}` (another example `${monthStart} ${yearStart} - ${monthEnd} ${yearEnd}` ) is used in multiple places. I suggest there's a function `someFnName = (monthStart, monthEnd, year) => '${monthStart} - ${monthEnd} ${year}'`. To keep things DRY and reduce the risk of inconsistency because of a typo or a future update to the formatting.

I would really prefer not to mess with this file any more than is absolutely necessary (refactoring, future-proofing). Since this Period Selector component is going to be replaced soon, any refactoring seems like not the best use of time. But I'm willing to discuss.

src/components/PeriodSelector/modules/FixedPeriodsGenerator.js
martinkrulltott: What I mean for these constants (`weekNumber`, `biWeekNumber` `monthName`) is that they're only used in one single place, like this example
`${i18n.t('Week {{weekNumber}}', { weekNumber })} - ${period.startDate} - ${period.endDate}` (https://github.com/dhis2/analytics/blob/3f5060970e26044fd2bdb0e89c1d59dbdbae202d/src/components/PeriodSelector/modules/FixedPeriodsGenerator.js#L90).

Can this be replaced with just 
`${i18n.t('Week {{weekNumber}}', { weekNumber: week })} - ${period.startDate} - ${period.endDate}` 
to avoid creating a single-use `const`?

martinkrulltott: regarding `monthName`, you do indeed have a const named `monthName`, it's here: https://github.com/dhis2/analytics/blob/3f5060970e26044fd2bdb0e89c1d59dbdbae202d/src/components/PeriodSelector/modules/FixedPeriodsGenerator.js#L188

janhenrikoverland (approved):
Refactors can be done later due to 2.33 time restrictions.

PR #423, opened 2020-05-13
fix: more period selector translations for v2.4

martinkrulltott:
Part of fix for: https://jira.dhis2.org/browse/DHIS2-8638

More i18n translation fixes based on https://github.com/dhis2/analytics/pull/418


martinkrulltott:
Turns out this was already fixed

PR #432, opened 2020-05-15
feat: drill down/up support on table cells

edoardo:
Part of the work for adding drilling feature to PT: https://jira.dhis2.org/browse/DHIS2-8769

Each value cell has now an onClick handler which is used to trigger a callback passed from the plugin and which receives a reference for positioning the context menu, and an object with the cell info.

This object is extended to contain also the ids of the org unit and the period which define the cell value.
(This would probably change with the refactor of the PT engine)

These parameters are then used by the `ContextualMenu` component in DV's plugin (for now, it can be moved out from the plugin and in `analytics`) for configuring the menu items.

### DIFF
--- a/src/components/PivotTable/PivotTableValueCell.js
+++ b/src/components/PivotTable/PivotTableValueCell.js
@@ -1,5 +1,6 @@
-import React from 'react'
+import React, { useRef } from 'react'
 import PropTypes from 'prop-types'
+
 import { applyLegendSet } from '../../modules/pivotTable/applyLegendSet'
 import { PivotTableCell } from './PivotTableCell'
 import { usePivotTableEngine } from './PivotTableEngineContext'
@@ -9,7 +10,11 @@ import {
 } from '../../modules/pivotTable/pivotTableConstants'
 import { PivotTableEmptyCell } from './PivotTableEmptyCell'
 
-export const PivotTableValueCell = ({ row, column }) => {
+export const PivotTableValueCell = ({
+    row,
+    column,
+    onToggleContextualMenu,
+}) => {
     const engine = usePivotTableEngine()
 
     const cellContent = engine.get({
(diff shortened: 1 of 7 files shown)


src/modules/pivotTable/PivotTableEngine.js
```diff
@@ -320,6 +323,19 @@ export class PivotTableEngine {
...
+                    header => header.dimension === DIMENSION_ID_ORGUNIT
+                )
+            ]
+        const peId =
+            dataRow[
+                this.dimensionLookup.headerDimensions.findIndex(
+                    header => header.dimension === DIMENSION_ID_PERIOD
+                )
+            ]
+
```
amcgee: Is it possible that an `ou` or a `pe` header won't exist?  I think they should always exist, but if not this will return `-1` which will return the last column in the dataRow, and that's not what we want.

amcgee (approved):
Looks good - one clarifying comment though!

src/components/PivotTable/PivotTableValueCell.js
```diff
@@ -40,19 +45,32 @@ export const PivotTableValueCell = ({ row, column }) => {
         maxWidth: width,
     }
 
+    const cellRef = useRef(undefined)
```
amcgee: This needs to move above the early `return` on line 26

https://travis-ci.com/github/dhis2/analytics/builds/166674837#L4203

amcgee (changes requested):

src/components/PivotTable/PivotTableValueCell.js
edoardo: Yes, I figured and fixed locally.

src/modules/pivotTable/PivotTableEngine.js
edoardo: Good point, not sure it's possible to not have `ou` and `pe` headers, but I can fix the code to avoid the wrong data returned.

edoardo: Fixed in a9c85c5065d2ae28c8f9dac0741b9dcf96296e26

amcgee (approved):
Looks good!

yarn.lock
```diff
@@ -1347,7 +1347,18 @@
   dependencies:
     prop-types "^15"
 
-"@dhis2/ui-core@^4.19.1", "@dhis2/ui-core@^4.6.1", "@dhis2/ui-core@^4.9.1":
```
martinkrulltott, line 1361: Silly `yarn` 🤷‍♂️

edoardo: Fixed.

PR #437, opened 2020-05-19
chore: replace travis with actions

janhenrikoverland:
Remove travis config, add gh actions config. Will work for the 2.4.x and 4.x branches as well.


.github/workflows/node-publish.yml
```diff
@@ -0,0 +1,47 @@
...
+            - '[0-9]+.x.x'
+            - '[0-9]+.[0-9]+.x'
+
+env:
+    GIT_AUTHOR_NAME: '@dhis2-bot'
+    GIT_AUTHOR_EMAIL: 'apps@dhis2.org'
+    GIT_COMMITTER_NAME: '@dhis2-bot'
+    GIT_COMMITTER_EMAIL: 'apps@dhis2.org'
+    NPM_TOKEN: ${{secrets.NPM_TOKEN}}
+    GH_TOKEN: ${{secrets.GH_TOKEN}}
```
amcgee: To use the org secrets let's use `secrets.DHIS2_BOT_NPM_TOKEN` and `secrets.DHIS2_BOT_GITHUB_TOKEN`

.github/workflows/node-test.yml
```diff
@@ -0,0 +1,21 @@
...
+            - uses: actions/checkout@v1
+            - uses: actions/setup-node@v1
+              with:
+                  node-version: 12.x
+
+            - name: Install
+              run: yarn install
+
+            - name: Test
+              run: yarn test
```
amcgee: Might be a good idea to add a `yarn build` step here, so that we make sure the build doesn't break in PRs?

amcgee (approved):
Looks good!  Maybe we should remove the `no merge commits` option from [semantic.yml](https://github.com/dhis2/analytics/blob/master/.github/semantic.yml#L4)?

PR #438, opened 2020-05-19
fix: complete translations [v4.x]

jenniferarnesen:
Part of fix for https://jira.dhis2.org/browse/DHIS2-8894

This is nearly the same as what went into version 2.4 (multiple PRs for 2.4)
* moved PeriodSelector to analytics from d2-ui, due to circular dependency causing issues with multiple instances of d2-i18n (commits were cherry-picked from v2.4 so it shouldn't be necessary to review PeriodSelector code in any detail)
* Put i18n.t into functions

### DIFF
--- a/src/modules/dataTypes.js
+++ b/src/modules/dataTypes.js
@@ -23,55 +23,51 @@ const DATA_SETS = 'dataSets'
 const EVENT_DATA_ITEMS = 'eventDataItems'
 const PROGRAM_INDICATORS = 'programIndicators'
 
-const TOTALS = 'totals'
-const DETAIL = 'detail'
+export const TOTALS = 'totals'
+export const DETAIL = 'detail'
 
-const programText = i18n.t('Program')
-const selectProgramText = i18n.t('Select a program')
+const getProgramText = () => i18n.t('Program')
 
 export const dataTypes = {
     [INDICATORS]: {
         id: INDICATORS,
-        name: i18n.t('Indicators'),
-        groupLabel: i18n.t('Select indicator group'),
-        defaultGroup: { id: ALL_ID, name: i18n.t('[ All groups ]') },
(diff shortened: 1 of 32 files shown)


src/modules/dataTypes.js
```diff
@@ -23,55 +23,51 @@ const DATA_SETS = 'dataSets'
 const EVENT_DATA_ITEMS = 'eventDataItems'
 const PROGRAM_INDICATORS = 'programIndicators'
 
-const TOTALS = 'totals'
-const DETAIL = 'detail'
+export const TOTALS = 'totals'
+export const DETAIL = 'detail'
 
-const programText = i18n.t('Program')
-const selectProgramText = i18n.t('Select a program')
```
martinkrulltott, line 30: why not use `getSelectProgramText` instead of duplicating `i18n.t('Select a program')` below? (to follow the pattern used by `getProgramText`)

src/visualizations/config/index.js
```diff
@@ -42,6 +36,13 @@ export default function({
     }
 
     this.getConfig = () => {
+        const DEFAULT_EXTRA_OPTIONS = {
+            colors: theme1,
+            noData: {
+                text: i18n.t('No data'),
```
martinkrulltott, line 42: Is the text itself supposed to be changed? `No data to display` -> `No data`

martinkrulltott (approved):
LGTM. Haven't tested locally. Assume all `/PeriodSelector/*` files were cherry-picked correctly.  I'm obviously not going to go through a 10000 line `yarn.lock` diff, but please make sure that it's been deduplicated (maybe thats actually the reason for the large diff?)

src/visualizations/config/index.js
jenniferarnesen: Yes, that was in discussion with Jan.

PR #439, opened 2020-05-20
fix: use pointer cursor style on clickable value cells

edoardo:
Part of the work for adding drilling feature to PT: https://jira.dhis2.org/browse/DHIS2-8769

Use pointer style for the cursor when the value cell is clickable for drilling.

Always enable drilling also on empty value cells.


amcgee (approved):
LGTM

PR #443, opened 2020-05-20
fix: support empty cell drill by looking up headers instead of data rows

amcgee:
This also correctly excludes period and orgunit dimension items when the dimension is in filter


edoardo (approved):
LGTM

PR #444, opened 2020-05-20
fix: fix drilling on empty value cells

edoardo:
- pass along the ref for the ContextualMenu positioning
- pass onClick handler on empty cells
- return ouId and peId in cellContent also for empty cells to enable
drilling

### DIFF
--- a/src/components/PivotTable/PivotTableEmptyCell.js
+++ b/src/components/PivotTable/PivotTableEmptyCell.js
@@ -3,9 +3,9 @@ import PropTypes from 'prop-types'
 import { cell as cellStyle } from './styles/PivotTable.style'
 import { PivotTableCell } from './PivotTableCell'
 
-export const PivotTableEmptyCell = ({ type, ...props }) => {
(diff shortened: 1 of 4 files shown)


src/components/PivotTable/PivotTableEmptyCell.js
```diff
@@ -3,14 +3,15 @@ import PropTypes from 'prop-types'
 import { cell as cellStyle } from './styles/PivotTable.style'
 import { PivotTableCell } from './PivotTableCell'
 
-export const PivotTableEmptyCell = ({ type, ...props }) => {
+export const PivotTableEmptyCell = ({ type, forwardedRef, ...props }) => {
```
amcgee: Should this use `React.forwardRef` as well?

amcgee (approved):
One minor suggestion, but generally looks good!

src/components/PivotTable/PivotTableEmptyCell.js
edoardo: Changed.

PR #451, opened 2020-05-25
fix: disable onClick on value cells if no ouId

edoardo:
Right now we don't have support for period drill up/down, so in case ou
dimension is in filters, the contextual menu opens empty (peId is passed
if pe dimension is in rows/columns).

### DIFF
--- a/src/components/PivotTable/PivotTableValueCell.js
+++ b/src/components/PivotTable/PivotTableValueCell.js
@@ -33,7 +33,7 @@ export const PivotTableValueCell = ({
         return (
             <PivotTableEmptyCell
                 type={cellContent?.cellType}
-                onClick={onClick}
+                onClick={onToggleContextualMenu ? onClick : undefined}
                 ref={cellRef}
             />
         )
@@ -60,7 +60,7 @@ export const PivotTableValueCell = ({
 
     const classes = [cellContent.cellType, cellContent.valueType]
 
-    if (cellContent.ouId) {
+    if (onToggleContextualMenu && cellContent.ouId) {
         classes.push('clickable')
     }
     return (
@@ -69,7 +69,7 @@ export const PivotTableValueCell = ({
             classes={classes}
             title={cellContent.renderedValue}
             style={style}
-            onClick={onClick}
+            onClick={onToggleContextualMenu ? onClick : undefined}
             ref={cellRef}
         >
             {cellContent.renderedValue ?? null}
--- a/src/components/PivotTable/PivotTableValueCell.js
+++ b/src/components/PivotTable/PivotTableValueCell.js
@@ -66,7 +66,7 @@ export const PivotTableValueCell = ({
     return (
         <PivotTableCell
             key={column}
-            classes={classes} //[cellContent.cellType, cellContent.valueType, isClickable]}
+            classes={classes}
             title={cellContent.renderedValue}
             style={style}
             onClick={onClick}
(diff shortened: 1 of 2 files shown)


src/components/PivotTable/PivotTableValueCell.js
```diff
@@ -58,10 +58,15 @@ export const PivotTableValueCell = ({
...
+    const classes = [cellContent.cellType, cellContent.valueType]
+
+    if (cellContent.ouId) {
+        classes.push('clickable')
+    }
     return (
         <PivotTableCell
             key={column}
-            classes={[cellContent.cellType, cellContent.valueType]}
+            classes={classes} //[cellContent.cellType, cellContent.valueType, isClickable]}
```
janhenrikoverland: I guess the comment is not needed

edoardo: Yep, leftover code. Cleaned up.

src/components/PivotTable/PivotTableValueCell.js
```diff
@@ -58,13 +58,18 @@ export const PivotTableValueCell = ({
...
+    }
     return (
         <PivotTableCell
             key={column}
-            classes={[cellContent.cellType, cellContent.valueType]}
+            classes={classes}
             title={cellContent.renderedValue}
             style={style}
-            onClick={onClick}
+            onClick={onToggleContextualMenu ? onClick : undefined}
```
janhenrikoverland: Maybe just do `onToggleContextualMenu && onClick`

janhenrikoverland (approved):

src/components/PivotTable/PivotTableValueCell.js
```diff
@@ -58,13 +58,18 @@ export const PivotTableValueCell = ({
         maxWidth: width,
     }
 
+    const classes = [cellContent.cellType, cellContent.valueType]
```
amcgee: Logic looks fine to me, might be a little cleaner like this:

```js
const isClickable = cellContent.cellType === CELL_TYPE_VALUE && cellContent.ouId
const onClick = () => { onToggleContextualMenu(cellRef, cellContent) }
const classes = [cellContent.cellType, cellContent.valueType, isClickable && 'clickable']

//...

return <PivotTableCell
            key={column}
            classes={classes}
            title={cellContent.renderedValue}
            style={style}
            onClick={isClickable ? onClick : undefined}
            ref={cellRef}
/>
```

amcgee (approved):
Minor suggestion, may or may not be cleaner

src/components/PivotTable/PivotTableValueCell.js
edoardo: Can't do:
> index.js:1 Warning: Expected `onClick` listener to be a function, instead got `false`.

> If you used to conditionally omit it with onClick={condition && value}, pass onClick={condition ? value : undefined} instead.

src/components/PivotTable/PivotTableValueCell.js
edoardo: Cleaned up.

src/components/PivotTable/PivotTableValueCell.js
```diff
@@ -33,7 +27,7 @@ export const PivotTableValueCell = ({
         return (
             <PivotTableEmptyCell
                 type={cellContent?.cellType}
-                onClick={onClick}
+                onClick={onToggleContextualMenu ? onClick : undefined}
```
amcgee, line 30: @edoardo I think `onClick` isn’t defined yet here so this will break on empty cells?

edoardo: True, fixing in a new PR.

PR #452, opened 2020-05-25
feat: PeriodDimension using ui-core components (DHIS2-8807)

martinkrulltott:
**TL;DR;**

1. Moves the period selector from `@dhis2/d2-ui` to Analytics
2. Material UI replaced by `ui-core`
3. Adds i18n changes from v2.4.x
4. BREAKING CHANGE and known issues / missing features

--------
**Description** 
Implements [DHIS2-8807](https://jira.dhis2.org/browse/DHIS2-8807). The `PeriodDimension` (aka _"period selector"_) used to depend on [@dhis2/d2-ui-period-selector-dialog](https://www.npmjs.com/package/@dhis2/d2-ui-period-selector-dialog). This has now been incorporated in to Analytics directly and refactored to use [@dhis2/ui-core Transfer component](https://ui-core.dhis2.nu/#/api?id=transferproptypes-object), to no longer depend on Material UI.

More details and full spec in the JIRA ticket.
Note: `Year` is now a number input instead of a dropdown.

At the same time a lot of the i18n changes made by @jenniferarnesen for [v2.4.x](https://github.com/dhis2/analytics/tree/2.4.x) (https://github.com/dhis2/analytics/pull/418, https://github.com/dhis2/analytics/pull/438) have been included as well.

As `Transfer` takes care of all the item handling (select, deselect, reorder) internally, the `onDeselect` and `onReorder` props are no longer in use, as this functionality is now bundled within the `onSelect` prop, hence the breaking change.

--------
**BREAKING CHANGE** 
`PeriodDimension` props `onDeselect` and `onReorder` deprecated

--------
**Known issues**
- Selected items missing with changing filter. https://github.com/dhis2/ui/issues/50
- DV specific: Selected items missing when reopening the modal after an _Update_ to the API. Same as above (https://github.com/dhis2/ui/issues/50)
- Filter elements misaligned. https://github.com/dhis2/ui/issues/89

--------
**Missing features**

- Filter state (i.e. what's been selected in the dropdowns / input field) isn't preserved when the modal is closed/reopened. _Solution pending discussion_
- Right side title (_Selected Periods_ in spec) missing. https://github.com/dhis2/ui/issues/90
- Right side styles (right side items are green in the spec) missing. https://github.com/dhis2/ui/issues/105

--------
**Screenshots**

_Default state, the period selector is now based on ui-core_ `Transfer` 
image

`Year` _is now an input field_
image

_Filter elements misaligned. The ui-core issue causes an incorrect margin/spacing between elements (noted in red below) because the label element is too wide (highlighted in blue below)._ 
image

_Selected items missing with changing filter_
filter clears selection

_DV specific: Selected items missing when reopening the modal after an Update to the API_
selection issue in dv


jenniferarnesen:
The main plan for now is to move the Period Selector to analytics and use the Transfer component, saving the refactoring of the PeriodSelector code for another time, right? (just confirming, otherwise I will look at the PR with different goals).

LG, just a few things:
FixedPeriodsGenerator has tests (they caught some issues when I was doing the i18n work):  https://github.com/dhis2/analytics/blob/4.x/src/components/PeriodSelector/modules/__tests__/FixedPeriodsGenerator.spec.js

PeriodDimension has stories in the other branches:  https://github.com/dhis2/analytics/blob/4.x/stories/PeriodDimension.stories.js

I got a console warning when opening the Period Dimension (but it could be due to my DV code not being entirely compliant with the Period component - I didn't check that)
image

src/components/PeriodDimension/utils/RelativePeriodsGenerator.js
```diff
@@ -0,0 +1,140 @@
+import i18n from '@dhis2/d2-i18n'
```
jenniferarnesen: Since this file now exports functions instead of a class, by convention it should start with lowercase. Also, maybe it should just be called "relativePeriods"?  (same goes for the FixedPeriodGenerator)

src/assets/DimensionItemIcons/PeriodIcon.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react'
```
jenniferarnesen, line 1: In the src/assets directory there is a file called DynamicDimensionIcon.js. It makes me wonder how this folder should be organized - keep all dimension icons together in a single folder? Or keep all dimension icons directly under src/assets (e.g. PeriodDimensionIcon.js)

jenniferarnesen:
You referenced a few ui issues in the description (e.g. dhis2/ui#50), but I couldn't find where it describes the issue about the selected items disappearing when switching the filter.

src/assets/DimensionItemIcons/PeriodIcon.js
martinkrulltott: The reason for the new folder was that there's a bunch of other icons (https://github.com/dhis2/design-system/tree/master/resources/icons/data-type) coming soon for the Data dimension modal (next in line), so I thought I'd prepare for those. They are very similar to the existing icons (like the already existing [PeriodIcon](https://github.com/dhis2/analytics/blob/0fea8feb0e1972202f2f480155de8ec142e3bc83/src/assets/PeriodIcon.js)) but we need both. However all of these will move til the ui-icons library in the near future regardless.

src/components/PeriodDimension/PeriodSelector.js
```diff
@@ -0,0 +1,187 @@
...
+                />
+            ))}
+        </Transfer>
+    )
+}
+
+PeriodSelector.propTypes = {
+    onSelect: PropTypes.func.isRequired,
+    rightFooter: PropTypes.node,
+    selectedItems: PropTypes.arrayOf(PropTypes.object),
```
jenniferarnesen: One thing that always confused me when I'd been away from this code and then came back to it again, was having both `selectedItems` and `selectedPeriods`.  I thought about renaming `selectedItems` to `initialSelectedPeriods`, to clarify its purpose. What do you think?

src/components/PeriodDimension/utils/RelativePeriodsGenerator.js
martinkrulltott: Fixed

src/components/PeriodDimension/__tests__/FixedPeriodsGenerator.spec.js
```diff
@@ -0,0 +1,675 @@
+import {
```
jenniferarnesen: Could you rename this file to `fixedPeriods.spec.js` ?

jenniferarnesen (approved):

PR #454, opened 2020-05-26
fix: onClick and classes props on PivotTableEmptyCell

edoardo:
Correctly pass onClick handler to PivotTableEmptyCell.

Also refactored the component to use `classes` instead of `className` to allow for multiple classes.
For empty value cells, `clickable` class should be passed alongside the cell `type` one.


amcgee (approved):
LGTM!

PR #455, opened 2020-05-26
chore: replace travis with GH actions [4.x]

jenniferarnesen:
This is to enable semantic release on the 4.x branch.


README.md
```diff
@@ -1,7 +1,19 @@
...
+
+[Module layout documentation](./docs/module-layout.md)
+
+## Publishing
+
+The analytics package is published to npm as @dhis2/analytics.
+
+To publish, simply mark the commit as a patch or minor change using semantic versioning (see note below). Once committed, github actions will take care of publishing the new version to npm.
+
+**Note**: All changes to the 4.x branch must marked as minor or patch (using e.g. "fix:" or "feat:"), even if it is technically a breaking change. Attempting to publish a breaking change on this branch will fail since 5.0.0 has already been published.
```
amcgee: Minor:
> must **be** marked

Probably fine to leave out the second half of the first sentence and all of the second one (which is branch-specific so will diverge from master), we shouldn't be pushing breaking changes for backports.  Fine to leave in if you want, just make it generic.

amcgee (approved):

README.md
jenniferarnesen: The translations fix for 2.4 and 4.x had a breaking change (name as a function rather than string).

PR #458, opened 2020-05-26
chore: replace travis with GH actions [2.4.x]

jenniferarnesen:
This is to enable semantic release on the 4.x branch.

Includes:
* Branch-agnostic publishing info is also added to the README.
* Remove style pre-commit hook as there are several style issues reported in stable code, and the 2.4.x branch is expected to be stable now, with few changes.


.github/workflows/node-publish.yml
```diff
@@ -0,0 +1,47 @@
...
+            - uses: actions/checkout@v2
+            - uses: actions/setup-node@v1
+              with:
+                  node-version: 12.x
+
+            - name: Install
+              run: yarn install --frozen-lockfile
+
+            - name: Lint
+              run: yarn lint
```
amcgee, line 39: If style checks fail, the publish will fail here.  Probably needs to be removed from the action unless we can add eslint-disable comments

amcgee (approved):
Approved after the lint issue is fixed

.github/workflows/node-publish.yml
jenniferarnesen: I could add eslint-disable to the affected files.

PR #462, opened 2020-05-28
fix: complete translations (DHIS2-8894)

jenniferarnesen:
Part of fix for https://jira.dhis2.org/browse/DHIS2-8894

Same translation fixes as in v34 branch: https://github.com/dhis2/analytics/pull/438 except:
*  does not include period selector as that is being handled separately.
* no need for old d2-i18n-extract and localize npm script as the cli scripts take care of the string extract/generating


src/components/DataDimension/Groups.js
```diff
@@ -14,10 +14,11 @@ export const Groups = ({
...
     onGroupChange,
 }) => {
-    const defaultGroup = dataTypes[dataType].defaultGroup
     let optionItems = groups
 
-    if (defaultGroup) {
+    if (dataTypes[dataType].defaultGroup) {
+        const { id, getName } = dataTypes[dataType].defaultGroup
+        const defaultGroup = { id, name: getName() }
         optionItems = [defaultGroup, ...optionItems]
```
martinkrulltott: change to `optionItems = [{ id, name: getName() }, ...optionItems]`? or is `defaultGroup` used somewhere else?

martinkrulltott (dismissed):
LGTM. @jenniferarnesen have you tested this locally with DV and Dashboards?

martinkrulltott:
There's a bug when clicking the Data dimension in DV, caused by updates in this PR. WIP... dont merge!

src/modules/predefinedDimensions.js
```diff
@@ -13,19 +13,19 @@ export const DIMENSION_PROP_NO_ITEMS = 'noItems'
 const FIXED_DIMENSIONS = {
     [DIMENSION_ID_DATA]: {
         id: DIMENSION_ID_DATA,
-        name: i18n.t('Data'),
+        getName: () => i18n.t('Data'),
```
martinkrulltott: Changes like these are actually breaking, since they're subsequently exported publicly through the likes of `getDimensionById` et al. Either mark this PR as a breaking change or make sure that the exposed API signature remains the same (by e.g. change the `getName()` prop back to `name` for the exposed functions.

jenniferarnesen: Yes, I'll mark it as a BREAKING CHANGE.

jenniferarnesen:
> @jenniferarnesen have you tested this locally with DV and Dashboards?

Yes of course.

src/components/DataDimension/Groups.js
jenniferarnesen: Changed, though I did have the extra const there to make the code somewhat more self explanatory.

src/modules/predefinedDimensions.js
jenniferarnesen: After some tweaks, I was able to redo the implementation to avoid the breaking change (thanks for the tip!)

src/modules/predefinedDimensions.js
```diff
@@ -10,7 +10,7 @@ export const DIMENSION_ID_ORGUNIT = 'ou'
 export const DIMENSION_ID_ASSIGNED_CATEGORIES = 'co'
 export const DIMENSION_PROP_NO_ITEMS = 'noItems'
 
-const FIXED_DIMENSIONS = {
+const getFixedDimensionsInternal = () => ({
```
martinkrulltott: Nice that you managed to solve the breaking change issue! 👍  I assume that just exporting this function directly (instead of routing through an "Internal" function didn't work? E.g. 
`export const getFixedDimensions = () => ({
[DIMENSION_ID_DATA]: { ..... } 
})`

martinkrulltott (approved):

src/modules/predefinedDimensions.js
jenniferarnesen: all the "internal" functions are used more than once

martinkrulltott: yes, but the "external" functions are just pointing to the "internal" directly, so couldn't the "internal" ones just be exported as well? (and omit the "internal" name)

PR #476, opened 2020-06-08
feat: upgrade to ui@5 (TECH-387)

martinkrulltott:
Implements [TECH-387](https://jira.dhis2.org/browse/TECH-387)
Upgrades `ui-core` dep to `ui@5` (which has a lot of breaking changes)

Relates to https://github.com/dhis2/data-visualizer-app/pull/1068 and https://github.com/dhis2/dashboards-app/pull/891

**Major topics:**
- PeriodSelector refactored to use `name` again
- DynamicDimension refactored, note: won't work with pagination!
- New stories for Filter, PeriodDimension and DynamicDimension
- Any component using `Transfer`, `SingleSelect` or `MultiSelect` should've been updated

-----------------

**TODO**

- [x] Reimplement TransferOption icons

PR #504, opened 2020-06-18
fix: data dimension spacing and crash (NO JIRA)

martinkrulltott:
`NO JIRA` as these issues should've been part of https://github.com/dhis2/analytics/pull/476

The Data dimension was still using the legacy way of handling selected items in `SingleSelect` from `ui`, causing selection to crash. This has been addressed by removing the `.value` from the selected option.

This fix also addresses a minor ui glitch with spacing in the select field area.
Correct spacing:

image

-----------------

image

-----------------

image


janhenrikoverland (approved):
LG

PR #508, opened 2020-06-29
feat: 2 category charts (DHIS2-7876)

edoardo:
**WIP: @dhis2/analytics@8.2.0-alpha**

This PR is for adding support for 2 category charts as per https://jira.dhis2.org/browse/DHIS2-7876.
The feature is enabled only for column and stacked column chart types for now.
For these types the DV layout allows to have 2 dimensions in the category area.
If only one dimension is used, the chart should work exactly as before.
When the 2nd category is used, various things need to be adjusted in the Highcharts config to make it render something like this:
Screenshot 2020-06-29 at 10 22 22

Stacked values for the same chart look like this:
Screenshot 2020-06-29 at 10 25 10

Sorting works a bit differently, it's only possible to sort on the 1st serie, and the "groups" of values are moved around within the 1st category dimension items:
Screenshot 2020-06-29 at 10 28 26

More screenshots for different vis types:
Screenshot 2020-07-31 at 14 16 01
Screenshot 2020-07-31 at 14 18 04
Screenshot 2020-07-31 at 14 18 49
Screenshot 2020-07-31 at 14 19 13

DV options still to be worked out:
- ~Cumulative values~
- ~Hide empty categories~
- ~Trend line~

Other things to check/fix:
- ~Save a 2 category does not work. The 2nd category is moved to filters~

Issues to solve:
- ~in Area, when using the `Stacked values add to 100% option` (`stacking: "percent"` in Highcharts), the 2 sections of the chart are not separated but connected with a 0 value in between.~
Screenshot 2020-07-31 at 14 22 29


src/modules/layoutUiRules/rules.js
```diff
@@ -79,8 +95,8 @@ const pivotTableRules = {
 }
 
 const visTypeToRules = {
-    [VIS_TYPE_COLUMN]: defaultRules,
-    [VIS_TYPE_STACKED_COLUMN]: defaultRules,
+    [VIS_TYPE_COLUMN]: dualCategoryRules,
```
martinkrulltott: Breaking change? @janhenrikoverland

martinkrulltott: With this update, the current version of DV will be able to add 2 items to Category but won’t be able to process the implications properly.
image

martinkrulltott: Turns out there was an issue with the backend (debug/dev). Testing @dhis2/analytics@8.2.0-alpha.1 in DV master against debug/2.34dev works! So seems like there's no breaking changes.

image

src/modules/layoutUiRules/rules.js
```diff
@@ -39,6 +39,22 @@ const defaultRules = {
     },
 }
 
+const dualCategoryRules = {
```
janhenrikoverland: Let's stick to the "Two category" term, so `twoCategoryRules`

janhenrikoverland: Maybe the two category rules should be something that is applied on top of the vis type rules. So we could do e.g. `{ ...rulesByType, ...extraRulesByFeature }` so that we avoid e.g. `[VIS_TYPE_LINE]: dualCategoryRules`. Thoughts?

src/visualizations/config/adapters/dhis_highcharts/addTrendLines.js
```diff
@@ -20,15 +24,26 @@ const DEFAULT_TRENDLINE = {
 export const isRegressionIneligible = type =>
     arrayContains([VIS_TYPE_GAUGE, VIS_TYPE_PIE], type)
 
-export default function(regressionType, series, isStacked) {
+export default function(layout, series, isStacked) {
+    if (isDualCategoryChartType(layout.type) && layout.rows.length > 1) {
```
janhenrikoverland: @edoardo Check the rules instead of the hardcoded `> 1` ?

src/visualizations/config/adapters/dhis_highcharts/getCumulativeData.js
```diff
@@ -32,3 +34,54 @@ export default function(series) {
...
+            )
+
+            decimals = 0
+        })
+
+    return series
+}
+
+export default function(series, layout) {
+    if (isDualCategoryChartType(layout.type) && layout.rows.length > 1) {
```
janhenrikoverland: Same as above

src/visualizations/config/adapters/dhis_highcharts/getSortedConfig.js
```diff
@@ -36,8 +98,10 @@ function getSortedConfig(config, isStacked, direction) {
     return sortedConfig
 }
 
-export default function(config, layout, isStacked) {
-    const direction = sortOrderMap.get(parseInt(layout.sortOrder))
-
-    return getSortedConfig(config, isStacked, direction)
+export default function(config, layout, stacked) {
+    if (isDualCategoryChartType(layout.type) && layout.rows.length > 1) {
```
janhenrikoverland: Same

src/visualizations/config/adapters/dhis_highcharts/getStackedData.js
```diff
@@ -1,4 +1,14 @@
-export default function(series, isZeroAsNull) {
+import { isDualCategoryChartType } from '../../../../modules/visTypes'
+
+export default function(series, layout) {
+    if (isDualCategoryChartType(layout.type) && layout.rows.length > 1) {
```
janhenrikoverland: Same

src/visualizations/config/adapters/dhis_highcharts/getTrimmedConfig.js
```diff
@@ -89,73 +93,156 @@ function cleanData(
...
-            categories: cleanData(
-                xAxis.categories,
-                emptySeriesIndexes,
-                firstValueIndex,
-                lastValueIndex,
-                hideEmptyRowItems
-            ),
-        },
+export default function(config, layout) {
+    if (isDualCategoryChartType(layout.type) && layout.rows.length > 1) {
```
janhenrikoverland: Same

src/visualizations/config/adapters/dhis_highcharts/index.js
```diff
@@ -127,14 +127,19 @@ export default function({ store, layout, el, extraConfig, extraOptions }) {
...
+    if (config.xAxis?.length) {
+        config.xAxis = config.xAxis.map(xAxis => ({
+            ...xAxis,
+            categories: xAxis.categories.flat(),
+        }))
     }
 
     // force apply extra config
     Object.assign(config, extraConfig)
+    console.log('chart config', config)
```
janhenrikoverland: Remove log

src/visualizations/store/adapters/dhis_highcharts/index.js
```diff
@@ -95,33 +107,33 @@ function getSeriesFunction(type) {
     }
 }
 
-export default function({ type, data, seriesId, categoryId }) {
-    const seriesFunction = getSeriesFunction(type)
+export default function({ type, data, seriesId, categoryIds }) {
+    console.log('cat ids', categoryIds)
```
janhenrikoverland: Remove log

janhenrikoverland (approved):
LGTM

src/modules/layoutUiRules/rules.js
edoardo: Solved for now with the simplest solution.
As suggested by @martinkrulltott only Radar differs from the other vis types that used the default layout rules, so the new default rules are now for 2 category, and Radar got its own rules to only allow 1 dimension for categories.

src/visualizations/config/adapters/dhis_highcharts/index.js
edoardo: Done

src/visualizations/store/adapters/dhis_highcharts/index.js
edoardo: Done

src/modules/layoutUiRules/rules.js
edoardo: Renamed all functions related to two category.

PR #509, opened 2020-07-02
feat: introduce fixed-period-single-select component

HendrikThePendric:
I needed a slightly different version of the `PeriodDimension` / `PeriodSelector` for the WHO DQ app. I discussed with @edoardo that adding it here would be OK. The component mainly uses helpers that were already there, and I've also added two new ones. Both the new component and those two new helper function have test coverage.

I also made a few changes to the Storybook configuration, so the correct font is added and the page has some very basic styling.

There are two changes I think I need to draw your attention to:
1. I couldn't run the project because of a failing import of `./locales` in `src/index.js`. So I commented that line out. I was expecting to be able to run some sort of script to generate the `./locales` but couldn't find one. It's probably wrong that this line is commented out, but I'm not sure what to do with that....
2. I've changed the period IDs from human-readable strings to codes (uppercase, no spaces). The reason I did this, is that my new component needs to be able to restrict the period-types that may be selected. For this I have added a `allowedPeriodTypes` prop (Array.<String>). I thought that to avoid typos when consuming the component, it'd be better to switch to these codes... After all, writing `['YEARLY', 'FNOV']` seems less error prone than `['Yearly', 'Financial year (Start November)']`

I'll place some comments regarding these two points in the code as well.

Looking forward to receiving your review.

----

### Update:
All comments should have been addressed. Ready for re-review


src/components/PeriodDimension/__tests__/fixedPeriods.spec.js
```diff
@@ -14,23 +16,23 @@ describe('fixedPeriods utils', () => {
...
+                'MONTHLY',
+                'BIMONTHLY',
+                'QUARTERLY',
+                'SIXMONTHLY',
+                'SIXMONTHLYAPR',
+                'YEARLY',
+                'FYNOV',
+                'FYOCT',
+                'FYJUL',
+                'FYAPR',
```
HendrikThePendric, line 35: Here's the part where I changed the period-type IDs as I mentioned in the PR description.

src/index.js
```diff
@@ -1,9 +1,10 @@
-import './locales'
+// import './locales'
```
HendrikThePendric: This line should probably be uncommented, but I'm not sure how to generate the `./locales` directory.

amcgee: @HendrikThePendric `yarn build` will generate the directory, it's part of the i18n extract/generate build steps

amcgee: Without this line there will be no translations

HendrikThePendric: Yup, I can confirm running `yarn build` did the trick. Uncommented this line in 3323118.

src/__demo__/PeriodDimension.stories.js
```diff
@@ -1,14 +1,26 @@
...
 import PeriodDimension from '../components/PeriodDimension/PeriodDimension'
+import FixedPeriodSingleSelect from '../components/PeriodDimension/FixedPeriodSingleSelect'
 
 const selectedPeriods = [{ id: 'LAST_12_MONTHS', name: 'Last 12 months' }]
 
 storiesOf('PeriodDimension', module).add('None selected', () => {
     return <PeriodDimension onSelect={selected => console.log(selected)} />
 })
 
+storiesOf('PeriodDimension', module).add('FixedPeriodSingleSelect', () => {
```
martinkrulltott: image
I think there's a need to rename the other stories, as the names (see screenshot) don't indicate that 2 stories are for the period selector and one is for the new FixedPeriodSingleSelect. Alternatively, the FixedPeriodSingleSelect could be its own story module.

src/index.js
```diff
@@ -4,6 +4,7 @@ import './locales'
 export { default as DataDimension } from './components/DataDimension/DataDimension'
 
 export { default as PeriodDimension } from './components/PeriodDimension/PeriodDimension'
+export { default as FixedPeriodSingleSelect } from './components/PeriodDimension/FixedPeriodSingleSelect'
```
martinkrulltott: I think there's a need to rename the components internally (keeping the same names externally to not create breaking changes), since "PeriodDimension/PeriodDimension" and "PeriodDimension/FixedPeriodSingleSelect" might not be easy to understand. Previously, "PeriodDimension" (along with "DataDimension" etc) was a container for everything needed to select the respective dimension (i.e. as of now - a wrapper for the Transfer component). However now that PeriodDimension also houses another period selector, maybe they should be called something like "PeriodTransfer" and "FixedPeriodSelector" to separate the two?

src/components/PeriodDimension/FixedPeriodSingleSelect.js
```diff
@@ -0,0 +1,95 @@
...
+    const [options, setOptions] = useState()
+
+    useEffect(() => {
+        const period = parsePeriodCode(value, allowedPeriodTypes)
+        if (period) {
+            setPeriodType(period.id)
+            setYear(period.year)
+            setOptions(period.options)
+        }
+        if (!period && periodType && year) {
```
martinkrulltott: `else if` ?

src/__demo__/PeriodDimension.stories.js
martinkrulltott: If moved to a separate module, could you please add a case that shows the `allowedPeriodTypes` in action as well?

src/components/PeriodDimension/FixedPeriodSingleSelect.js
```diff
@@ -0,0 +1,95 @@
...
+                    {options.map(option => (
+                        <SingleSelectOption
+                            key={option.id}
+                            value={option.id}
+                            label={option.name}
+                        />
+                    ))}
+                </SingleSelectField>
+            )}
+            <style jsx>{`
```
martinkrulltott: Maybe nitpicking since it's such tiny bit, but it's preferred to keep the jsx in a separate `styles/[componentName].style.js` file.

src/components/PeriodDimension/FixedPeriodFilter.js
```diff
@@ -50,6 +62,7 @@ FixedPeriodFilter.propTypes = {
     currentYear: PropTypes.string.isRequired,
     onSelectPeriodType: PropTypes.func.isRequired,
     onSelectYear: PropTypes.func.isRequired,
+    allowedPeriodTypes: PropTypes.arrayOf(PropTypes.string),
```
martinkrulltott, line 70: One could argue that as the code disregards the filtering of periodTypes when `allowedPeriodTypes` isn't passed, the invert of `allowedPeriodTypes` should be passed ("disallowedPeriodTypes"?) instead. As if nothing is passed all are allowed. 

I.e. not passing the prop would result in `"no disallowedPeriodTypes" = "all periodTypes"` instead of the semantically incorrect case of `"no allowedPeriodTypes" = "all periodTypes"`, as one might assume that `"no allowedPeriodTypes" = "no periodTypes at all"`.

But I assume that the use-case is to pass the allowed ones, not the opposite, so then this makes sense I guess 🤷‍♂️

src/index.js
martinkrulltott: Also, calling it `FixedPeriodSingleSelect` when it actually consists of two single selects and a number input field might not be the most semantic as well 🙂

src/components/PeriodDimension/FixedPeriodFilter.js
HendrikThePendric: I agree with the reasoning above, it would be more logical to end up with an exclude-list, i.e. `disallowedPeriodTypes`. But as you guessed, the use-case is usually to only allow a small handful of periodTypes.

There is another reason for going with the allow-list as well: In the couple of years I've worked here, I've seen a few new period-types emerge. Once a new one is added this component will need to be updated and that's fine. However, if we were to go with an exclude-list approach, all apps would have to update their `disallowedPeriodTypes` array whenever this happens. Using an allow-list approach would mean the new periodType wouldn't magically start appearing in these dropdowns after a version upgrade.

So yeah, I think sticking to the current implementation with `allowedPeriodTypes` seems best, even though the "ignore when undefined" part is a bit special.

I'll resolve this conversation and will keep this part as it is.

src/components/PeriodDimension/__tests__/fixedPeriods.spec.js
HendrikThePendric: I'm guessing you don't object to these changes, so will resolve this conversation

src/components/PeriodDimension/FixedPeriodFilter.js
```diff
@@ -19,15 +20,26 @@ const FixedPeriodFilter = ({
                 onChange={({ selected }) => onSelectPeriodType(selected)}
                 dense
                 selected={currentPeriodType}
+                disabled={
+                    Array.isArray(allowedPeriodTypes) &&
+                    allowedPeriodTypes.length === 1 &&
+                    allowedPeriodTypes[0] === currentPeriodType
+                }
```
amcgee: There's probably a good reason for this, but it's not obvious to me... why is it only disabled when the length is 1?  Maybe it would be good to pull this out into a variable which is descriptively-named?

src/components/PeriodDimension/FixedPeriodSelect.js
```diff
@@ -0,0 +1,89 @@
...
+        if (period) {
+            setPeriodType(period.id)
+            setYear(period.year)
+            setOptions(period.options)
+        } else if (periodType && year) {
+            const offset = getYearOffsetFromNow(year)
+            const options = getPeriodById(periodType).getPeriods({ offset })
+            setOptions(options)
+        }
+    }, [allowedPeriodTypes, periodType, value, year])
```
amcgee: I believe this will flash a single empty render before parsing and rendering the period, since `useEffect` happens after the first commit phase, is that correct?

amcgee: If so, is there a way to avoid this?

src/components/PeriodDimension/FixedPeriodSelect.js
```diff
@@ -0,0 +1,89 @@
...
+    return (
+        <div data-test={dataTest} className={className}>
+            <div className="row">
+                <FixedPeriodFilter
+                    allowedPeriodTypes={allowedPeriodTypes}
+                    currentPeriodType={periodType}
+                    currentYear={year}
+                    onSelectPeriodType={type => {
+                        setPeriodType(type)
+                        onChange('')
```
amcgee: Should this be `null` or `undefined` instead of an empty string...?

src/components/PeriodDimension/__tests__/fixedPeriods.spec.js
amcgee: @janhenrikoverland @martinkrulltott @edoardo just to confirm... this doesn't ever get sent to the server right?  Shouldn't we match the server-defined period IDs regardless?

amcgee (commented):
Added a few questions, but someone more familiar with this component should review

src/components/PeriodDimension/__tests__/fixedPeriods.spec.js
martinkrulltott: I had the same reflection, but as I understand it these IDs are only used to control which period _**type**_ to display and has nothing to do with the periods themselves. Highlighted in green below. This is only used for filtering and shouldn't be passed to the server.
image

src/components/PeriodDimension/FixedPeriodFilter.js
HendrikThePendric: This is a bit of an edge case. If only one period type is allowed, and that one also happens to be selected, there is no point allowing the user to interact with the dropdown since thus user would only be able to chose the option that is already selected. I'll pull this into a variable named `onlyAllowedTypeSelected`

src/components/PeriodDimension/FixedPeriodSelect.js
HendrikThePendric: I believe you're right. The first time `useEffect` will run is when the component mounts and this is the second render. Apart from changing this to a class component and also running this logic in the constructor, I wouldn't know any way of avoiding this.

I'm not sure if this is something we need to worry about though. It shouldn't be visible to the human eye, and I also think the current way is quite a common approach (i.e. either using `useEffect` or `componentDidMount` + `componentDidUpdate` (and not executing the same logic in the constructor).

src/components/PeriodDimension/__tests__/fixedPeriods.spec.js
HendrikThePendric: > Shouldn't we match the server-defined period IDs regardless?

I've had a quick search in the dhis2-core codebase, but couldn't really find any period type IDs. Perhaps when you mention server-defined period IDs, do you mean these enums?
https://github.com/dhis2/dhis2-core/blob/e1c12681de3f9a7b28ee29865f7afd25ad3e8f83/dhis-2/dhis-api/src/main/java/org/hisp/dhis/calendar/DateUnitType.java#L38-L55

src/components/PeriodDimension/FixedPeriodSelect.js
HendrikThePendric: I could change this, but have done this intentionally. Currently I have defined a default value for `value` to be an empty string in `FixedPeriodSelect.defaultProps`, and when resetting I also chose that value. This way the `value` prop is **always** of type `string`. This has two advantages:
- When I pass the value to `parsePeriodCode` I can safely assume it's a string so I can unconditionally call the `.match()` method.
- _[less relevant]_ I envisage using this component with a form library and typically these will translate `null` or `undefined` to empty strings as well.

Let me know what you prefer.

src/components/PeriodDimension/__tests__/fixedPeriods.spec.js
HendrikThePendric: Discussed this with @martinkrulltott on Slack and we can keep the current IDs. Resolving this conversation now.

src/components/PeriodDimension/FixedPeriodSelect.js
HendrikThePendric: I've discussed this with @martinkrulltott and we decided to go with an implicit `undefined`, which makes the component fall back on its `value` from `defaultProps`. https://github.com/dhis2/analytics/pull/509/commits/c32bf54b1b410a1a6c33ae6dcdff41995344d333

src/components/PeriodDimension/FixedPeriodSelect.js
HendrikThePendric: I've discussed this with @martinkrulltott and also looked into it a bit. When using CPU throttling x 6 I could visually observe the content flash. I guess since we're targeting low-end devices this is relevant. The initial solution looked especially bad, because the first render produces a partial UI. Martin suggested that if it's not possible to completely avoid the flash, it would be better to just return null instead of a partial UI on the first render. I agreed with that....

So I looked into two solutions:
1. Return `null` instead of a partial UI. So we just don't show anything on the first render, and on the second render everything is as it should be. This doesn't really solve the issue, but it does make it less bad. And on the plus side: we're still using a functional component and have not increased the line-count too much. See https://github.com/dhis2/analytics/pull/509/commits/ee5dd24e8c8c3f505c249cf33e3ce881354b7a67
2. Refactor to a class based component and use `getDerivedStateFromProps` to make sure the state is correct **before** the (first) render. It seems to me that this is actually working perfectly: I can't observe a content flash anymore, even with my CPU throttled x 6. The line-count has increased a quite a bit, but I would probably argue that the code is slightly more readable in its current form. See https://github.com/dhis2/analytics/pull/509/commits/66dd411131b8a273cc5045afa998c4a66bc5f2b6

I guess my personal preference would be to go with the class based component.

HendrikThePendric:
I've just added a small commit that changes the `onChange` callback payload. There were two reasons for this:
- This way we can access some of the period metadata, like `startDate`, `endDate` and most notably `name`. This will come in handy for me in the WHO DQ app.
- We're now following the `(payloadObject, event)` pattern we are using everywhere in our ui libs.

src/components/PeriodDimension/FixedPeriodSelect.js
amcgee: @HendrikThePendric the state of this component is derived exclusively from the props, so I don't think this should use a `useEffect` at all and it should actually remain a simple, pure functional component.  Something like this:

```js
const FixedPeriodSelect = ({
    onChange,
    allowedPeriodTypes,
    className,
    dataTest,
    value,
}) => {
    const period = parsePeriodCode(value, allowedPeriodTypes) || {}
    const [periodType, setPeriodType] = useState(period.id)
    const [year, setYear] = useState(period.year)

    let options = period.options
    if (periodType && year) {
        const offset = getYearOffsetFromNow(year)
        options = getPeriodById(periodType).getPeriods({ offset })
    }
```

I'm not sure about the logic underlying the overwrite of `options`, but kept the existing behavior.  This only accepts the `value` and `allowedPeriodTypes` to be specified once on component mount, but I believe that's the intent anyway (otherwise the state will get overwritten).  It would be possible to add a `useEffect` for the update here as well, if needed.

Finally, this could be cleaned up quite a bit I think with a helper function (maybe as a singular state variable) but I'll leave that as a bonus here.

src/components/PeriodDimension/FixedPeriodSelect.js
```diff
@@ -0,0 +1,89 @@
...
+    const [year, setYear] = useState('')
+    const [options, setOptions] = useState()
+
+    useEffect(() => {
+        const period = parsePeriodCode(value, allowedPeriodTypes)
+        if (period) {
+            setPeriodType(period.id)
+            setYear(period.year)
+            setOptions(period.options)
+        } else if (periodType && year) {
```
amcgee: I actually think this case will NEVER run if `value` and `allowedPeriodTypes` are passed, since `periodType` and `year` are derived from those?

src/components/PeriodDimension/FixedPeriodSelect.js
amcgee: If the intent is to allow this too be controlled or uncontrolled, then a `useEffect` with an `initialRender` state might work as well

amcgee: Here's an alternative approach:

```js
const parseOptions = period => {
    if (period.id && period.year) {
        const offset = getYearOffsetFromNow(year)
        return getPeriodById(periodType).getPeriods({ offset })
    }
    return period.options
}
const FixedPeriodSelect = ({
    onChange,
    allowedPeriodTypes,
    className,
    dataTest,
    value,
}) => {
    const [period, setPeriod] = useState(parsePeriodCode(value, allowedPeriodTypes))
    const options = parseOptions(period)

    // I would actually recommend AGAINST this, I don't think we should support both controlled and uncontrolled components...  not sure what the intent was here.  See https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html
    const useEffect(() => {
        const newPeriod = parsePeriodCode(value, allowedPeriodTypes)
        if (!shallowEqual(period, newPeriod)) {
            setPeriod(newPeriod)
        }
    }, [value, allowedPeriodTypes, period])
    
    // ..., use `setPeriod({ ...period, id: periodType })` etc. below
```

HendrikThePendric: >  the state of this component is derived exclusively from the props, so I don't think this should use a useEffect at all and it should actually remain a simple, pure functional component.

This is not entirely correct. There are two paths to consider:
1. There is a value: now the state period-type and year are fully derived
2. There is no value: now the state for period-type and year are not derived, and more importantly, user-interaction is required to set the period-type and year state. Only once these state properties both have a value, the actual period can be selected.

As such, I believe we do need state, and I think that the current implementation, which implements `getDerivedStateFromProps`, is a good fit.

amcgee: Nice, yeah the class Component version with `getDerivedStateFromProps` looks pretty clean to me

amcgee: 🎉

amcgee (approved):
Code LGTM, but would be good for someone more familiar with the usage to review and test too!

amcgee (approved):
Code LGTM, but would be good for someone more familiar with the usage to review and test too!

amcgee:
CC @martinkrulltott @dhis2/team-analytics

HendrikThePendric:
Some screenshots to show the component in action:
Screenshot 2020-08-12 at 09 18 02
Screenshot 2020-08-12 at 09 18 22
Screenshot 2020-08-12 at 09 18 46

martinkrulltott (approved):
Test the storybook locally, LGTM

PR #531, opened 2020-08-07
feat: add support for DV's colorSet option (DHIS2-670)

edoardo:
Part of the implementation for https://jira.dhis2.org/browse/DHIS2-670

The colorSet option in DV allows to choose a color set to use in the
chart.
A typical color set is a list of colors to use for distinguishing the
various series in a chart.
One color set is using patterns instead of colors, the patterns used are
the 10 default patterns provided by Highcharts.


janhenrikoverland (approved):
LG

PR #534, opened 2020-08-11
feat: support font style options (DHIS2-8426)

martinkrulltott:
**WIP: @dhis2/analytics@8.8.0-alpha**

Implements [DHIS2-8426](https://jira.dhis2.org/browse/DHIS2-8426)
Spec: [DHIS2-9061](https://jira.dhis2.org/browse/DHIS2-9061)
Relates to https://github.com/dhis2/data-visualizer-app/pull/1184

**Summary**
Adds support for font styles to all texts in DV by providing default values for the option modal which are then sent through the  highcharts config to be applied in the rendered visualizations. 

_Note 1:_ Even though this revolves heavily around a new prop called `fontStyle` this is not a breaking change since if no `fontStyle` is provided will result in the defaults being used instead. https://github.com/dhis2/analytics/blob/f3a5e52ece7cb567ffa2ef9017f311ab5c6723c3/src/visualizations/index.js#L31

--------------
**Main topics**
- Adds font style options and default values to be used in DV for selecting options
- Implements said options in the visualization generation process (as part of the Highcharts config) based on the [spec](https://jira.dhis2.org/browse/DHIS2-9061)
- Bumps `ui` to 5.4.1 for various bug fixes + new toggle button

- As of now, the available font style options are:
  - Font size
  - Text align
  - Color
  - Bold
  - Italic
- The available text elements are:
  - Visualization title
  - Visualization subtitle
  - Horizontal axis title
  - Vertical axis title
  - Legend
  - Series axis label
  - Category axis label
  - Base line label
  - Target line label

--------------

_Note 2_: There's a bit of complexity regarding the align options, as different elements require different values.
- Visualization title, subtitle, legend, base line label, target line label: left, center, right
- Series axis title, category axis title: start, middle, end - _Highcharts config: `low`, `middle`, `high`_
- Axis labels: no align options (hence doesn't include `FONT_STYLE_OPTION_TEXT_ALIGN` in the `defaultFontStyle`)

--------------

_Note 3_:  Two helper functions are available for handling the `fontStyle` object:
- `mergeFontStyleWithDefault`: Since DV only handles the fontStyles that are explicitly set (i.e. the result object is minimised), this function merges the default font style with the provided font style, to guarantee that all props are always available to the visualization generation process.
- `deleteFontStyleOption` - To facilitate the minimised strategy above, this function is used to delete a prop and all nodes that as a result are subsequently empty. E.g. `{ A: { A1: 'value' }, B: { B1: 'value' } }`, if `A1` is removed the whole parent node of `A` will also be removed. If `B1` is later removed the whole object will subsequently be empty as `{}`

--------------

_Note 4_: 
https://github.com/dhis2/analytics/blob/0e443906feaad78319136bf7ecc76c11dfc2b31a/src/visualizations/config/adapters/dhis_highcharts/chart.js#L18
A load function in `events` was added to the Highcharts config to dynamically calculate the position of the legend icon, as this doesn't natively position itself vertically centered if a custom font size is used for the legend text. Based on a ([solution](https://jsfiddle.net/BlackLabel/hoepjvmc/) from Highcharts support.


edoardo (approved):
LGTM

PR #515, opened 2020-07-08
feat: implement new series options and 4 axes (DHIS2-9103)

martinkrulltott:
Implements the new series options, which holds `axis` and `type`, to support custom types and 4 axes. [DHIS2-9103](https://jira.dhis2.org/browse/DHIS2-9103)

- **BREAKING CHANGE**: `dualAxisItemHandler` prop removed from `DimensionMenu`
- Renames `optionalAxes` to the more appropriate name `customAxes`
- Z-index logic to make sure _LINE_ items are always on top of _COLUMN_
- New axis colors
- `DimensionMenu` no longer uses `dualAxisItemHandler`, as this has been moved in https://github.com/dhis2/data-visualizer-app/pull/1143
- `VIS_TYPE_AREA` no longer part of `dualAxisTypes` - fixes [DHIS2-9011](https://jira.dhis2.org/browse/DHIS2-9011)
- Only `isDualAxisType` types will now display multiple axes - fixes [DHIS2-9012](https://jira.dhis2.org/browse/DHIS2-9012)

Note: As the new `series` array can contain all, some or none of the dimension items (depending on if/when the options dialog has been used), this uncertainty has to be taken in to consideration everywhere which increases the complexity.

-----------

**Known issues / bugs** _(todo: unchecked, fixed: checked)_

- [x] Switching from a multi axes series dimension to a single axis series dimension by dragging and dropping (e.g. replace Data with Period) will display the single axis series as "Axis 1" (in blue, instead of multi-coloured)

PR #535, opened 2020-08-11
fix: prevent saved AOs without series from crashing (DHIS2-6385)

martinkrulltott:
A previous PR (https://github.com/dhis2/analytics/pull/515) introduced a bug which caused saved AOs to crash, since they don't always have the `series` prop. This fix prevents this crash by using an optional chaining operator whenever `layout.series` is used.


janhenrikoverland (approved):
LG

PR #550, opened 2020-08-14
fix: enable multi axis for Area (DHIS2-9011)

edoardo:
Fixes [DHIS2-9011](https://jira.dhis2.org/browse/DHIS2-9011).

Required by [@dhis2/data-visualizer-app PR 1192](https://github.com/dhis2/data-visualizer-app/pull/1192).

Since we now differentiate between `Area` and `Stacked area`, multi axis can be enabled for `Area` (no stacking).


martinkrulltott (approved):
LGTM

PR #557, opened 2020-08-18
feat: add more daily relative periods (DHIS2-8438)

janhenrikoverland:
Add last 30/60/90/180 days as relative periods.

Screenshot from 2020-08-18 16-03-31


martinkrulltott (approved):
LGTM. Keep in mind that the change to [PeriodSelector.spec.js.snap](https://github.com/dhis2/analytics/pull/557/files#diff-135e765dd597fb1fb13e521fb2a50457) will cause the tests to fail. Make sure you have the latest from master and nuke your node modules before running the `yarn test --update` command. @edoardo just had the same issue 🙂

PR #571, opened 2020-08-26
fix: implement digit group separator in SV

edoardo:
Fixes [DHIS2-9340](https://jira.dhis2.org/browse/DHIS2-9340).

It uses the same functions for formatting the numeric value used in PT.
For this to work, both `skipRounding` and `digitGroupSeparator` DV options are used.

For screenshots and an example see https://github.com/dhis2/data-visualizer-app/pull/1228.


janhenrikoverland (approved):
LG

PR #574, opened 2020-08-27
feat: allow multi dx (DHIS2-8808)

janhenrikoverland:
BREAKING CHANGE: no series/columns for YOY charts

Year over year charts should no longer receive `dx` as columns, all other dimensions than `pe` should be passed as filter.


edoardo (approved):
LG

PR #578, opened 2020-09-01
fix: dimension menu updates requested by DV (DHIS2-9365)

martinkrulltott:
**WIP v9.0.1-alpha**

Fixes [DHIS2-9365](https://jira.dhis2.org/browse/DHIS2-9365)
Relates to https://github.com/dhis2/data-visualizer-app/pull/1237

- `DimensionMenu` no longer uses `anchorEl`, as the anchoring should now be handled by the parent component (requires a change in DV to function properly) 
- Refactors `DimensionMenu` to a functional component that uses `@dhis2/ui` instead of `MUI`
- `DimensionItem` now passed the ref of the option button in the `onOptionsClick` (to be used in DV to display the dimension menu in the right location)

-----------

**TODO**
- [x] Is this a breaking change?

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@dhis2/analytics",
-    "version": "9.0.1-alpha.1",
+    "version": "9.0.1-alpha.2",
     "main": "./build/cjs/lib.js",
     "module": "./build/es/lib.js",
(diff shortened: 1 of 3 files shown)


janhenrikoverland (approved):
LG

PR #583, opened 2020-09-02
feat: paging of dynamic dimension items (DHIS2-9274)

martinkrulltott:
Implements [DHIS2-9274](https://jira.dhis2.org/browse/DHIS2-9274)

Relates to https://github.com/dhis2/data-visualizer-app/pull/1239

Based on [POC](https://codesandbox.io/s/ui5-transfer-paging-forked-5t6s5)

This change adds infinite scrolling to dynamic dimensions, which enables the user to scroll down to fetch more items.

- Paginated requests to the server
- Fetches new items when scrolling to the bottom of the list
- Displays a loading spinner while fetching
- Searching is now done server-side and retains a paginated result
- Debounced searching to prevent server requests on every single letter that is typed
- Server request using the new app-platform `useDataEngine` instead of the old `d2` method

The `ItemSelector` now handles following states
- Initial request with a proper response
- Initial request with empty response (i.e. no items on the server at all) - displays `noItemsMessage`
- Subsequent request (after scrolling)
- Initial and subsequent requests with a search param
- Subsequent requests with empty response (i.e. no more items available on the server / all items already fetched)

---

### BREAKING CHANGE

- `DynamicDimension` no longer requires `context` and instead requires `dimensionTitle`

---

###  TODO
- [x] **DRAFT:** While waiting for the `dimensions` endpoint to support pagination, the `dataElements` endpoint is used for testing. This should be updated once [DHIS2-9272](https://jira.dhis2.org/browse/DHIS2-9272) is done

---

###  ISSUES
- [ ] Scrolling down, then applying/changing the search filter causes the scroll position to remain the same https://github.com/dhis2/ui/issues/253
- [x] Clicking any item in the list prevents the infinite scrolling from working https://github.com/dhis2/ui/issues/257
- [ ] Various buttons aren't disabled while loading [TECH-431](https://jira.dhis2.org/browse/TECH-431)

------------
With dummy data (for testing purposes)

Initial state
image

After scrolling down a few pages
image

After searching
image

------------

With real data

Initial state
image


src/api/dimensions.js
```diff
@@ -99,6 +100,28 @@ export const dataElementGroupsQuery = {
...
+    engine,
+    dimensionId,
+    searchTerm,
+    pageSize,
+    page,
+}) => {
+    const query = {
+        dimensions: {
+            resource: `dimensions/${dimensionId}/items`,
+            params: objectClean({
```
edoardo: Is `objectClean` really needed?
For the `filters` in `api/dimensions` I used an empty array on which various filters are pushed only when needed.

src/api/dimensions.js
```diff
@@ -99,6 +100,28 @@ export const dataElementGroupsQuery = {
     }),
 }
 
+export const apiFetchItemsByDimension = ({
```
edoardo: This is ok for now, but I would like it to follow the same format as the other queries here, if the query definition and fetch function stays here.
The other fetch functions return the data instead of the query object, so the consumer doesn't have to know anything about the query definition or the namespace used.

src/components/DynamicDimension/DynamicDimension.js
```diff
@@ -1,72 +1,57 @@
-import React, { useEffect, useState } from 'react'
+import React from 'react'
 import PropTypes from 'prop-types'
+import i18n from '@dhis2/d2-i18n'
+import { useDataEngine } from '@dhis2/app-runtime'
 
-import { useDataQuery } from '@dhis2/app-runtime'
 import ItemSelector from './ItemSelector'
-
-const itemsByDimensionQuery = {
```
edoardo, line 7: Alternatively this can be kept in the components just adding the paging (and filters) parameters.

src/components/DynamicDimension/ItemSelector.js
```diff
@@ -12,20 +12,105 @@ import {
...
+        }
+    }, [value, delay])
+
+    return debouncedValue
+}
+
 const ItemSelector = ({
-    allItems,
+    initialSelected,
+    moItemsMessage,
```
edoardo: Typo!
The param passed is `noItemsMessage`

edoardo (approved):
Some comments are about things we can fix later if preferred.
There is a typo in a parameter name tho.

edoardo:
One more comment.
The dialog's buttons in the screenshot looks quite wide apart, is that by design?

martinkrulltott:
> One more comment.
The dialog's buttons in the screenshot looks quite wide apart, is that by design?
Good find @edoardo . Previously `ui` didn't provide a margin in the `ButtonStrip` so I added a temporary fix, but now `ui` does so the margin is doubled. I removed it in the DV PR.

src/api/dimensions.js
martinkrulltott: Done

edoardo (approved):

edoardo (approved):

PR #587, opened 2020-09-04
fix: optional chaining for legend items (DHIS2-9413)

martinkrulltott:
Fixes [DHIS2-9413](https://jira.dhis2.org/browse/DHIS2-9413)

**Background**
https://github.com/dhis2/analytics/pull/534 introduced font styling for various text based options. As the font size of the legend could be changed, the corresponding icon (aka `legendSymbol`) had to be repositioned to be vertically aligned with the (now larger or smaller) text. However not all legend items has that object (namely trend lines), which caused legend items without the `legendSymbol` to crash the app.

**Fix**
By using optional chaining, items without the `legendSymbol` prop can be ignored.

**Proof** 
Locally opening the AO that caused this crash in the smoke test (`O52HFRvCjIi`) using the smoke test backend (`https://smoke.dhis2.org/dev_smoke/`).
image

### DIFF
--- a/src/visualizations/config/adapters/dhis_highcharts/chart.js
+++ b/src/visualizations/config/adapters/dhis_highcharts/chart.js
@@ -18,7 +18,7 @@ const getEvents = () => ({
     events: {
         load: function() { // Align legend icon with legend text
             this.legend.allItems.forEach((item) => {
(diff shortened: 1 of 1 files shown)


janhenrikoverland (approved):
LG

PR #589, opened 2020-09-07
fix: resolve previously introduced build error (DHIS2-9413)

martinkrulltott:
https://github.com/dhis2/analytics/pull/587 introduced a build error which didn't get caught by our build process. 
`eslint` doesn't support optional chaining within array functions, so the rule was disabled for this specific case.

EDIT: There's a mismatch between the eslint validation in the lint process and the build process. Instead of disabling the rule, the optional chaining was replaced by an if statement.


janhenrikoverland (approved):
LG

PR #591, opened 2020-09-07
fix: subtract initial empty columns from partition calculation [DHIS2-9381]

amcgee:
This fixes an issue with hidden empty columns (also empty rows) with at least one empty column/row occurring **before any column or row with data**.  This situation caused an invalid partition calculation.  The fix is a simple one-line change, to subtract the starting position of the first partition from the calculated starting position of the current clipping box.

This also adds two Storybook stories which exhibited the following behavior before this change:

1. empty columns - the `Cannot read property 'pre' of undefined` crash error @janhenrikoverland found
2. empty columns (weekly) - the extraneous empty column reported in [DHIS2-9381](https://jira.dhis2.org/browse/DHIS2-9381)

Both should be fixed by this change.


amcgee:
The only functional change here is [this line](https://github.com/dhis2/analytics/pull/591/files#diff-cb329c08ea9f1d6dfecab6d435884dffL20)

janhenrikoverland (approved):

PR #595, opened 2020-09-10
fix: text styling not applied when switching to 2 category

edoardo:
The font styling was not applied when switching to 2 category.
See screencast below:
ezgif com-video-to-gif (1)

The fix makes sure the styling is applied also in 2 category charts:
Screenshot 2020-09-10 at 11 09 24


jenniferarnesen (approved):

martinkrulltott:
What about the text align (position) prop?

edoardo:
> What about the text align (position) prop?

Clarified with @martinkrulltott the misunderstanding.
Position is not available for the labels, only the title.

martinkrulltott (approved):

PR #598, opened 2020-09-14
fix: exclude negative sign from digit group separation

amcgee:
This fixes an issue which caused `-,123` instead of `-123` when using a digit group separator for negative numbers with 3n digits


janhenrikoverland (approved):
LGTM

PR #604, opened 2020-09-17
fix: debounce Data Dimension's filterText change

KaiVandivier:
Calling `updateAlternatives` in the filterText onChange handler wasn't getting debounced properly, i.e. typing 10 characters in a row caused 10 network requests instead of 1.

It seems this happened because a _new_ anonymous debounced function was getting called on each event, instead of a single, persistent function that can track its debouncing status, so I made a named debounced function to call on each event.

Thanks for taking a look!


martinkrulltott (approved):
Great find! The fix looks good to me, tested locally 👍

PR #613, opened 2020-09-21
fix: setting org unit as series cause crash

martinkrulltott:
Issue: Setting org unit as series caused a crash. 
_Error message:_
image

Cause: `itemIds` passed as an array of objects `[{id: '12345'}]`  instead of an array of ids `['12345']`

Solution: Map the `id` prop to `dimension.items` in `dimensionHasRelativeItems`

_Proof: Org unit can be used as Series_
image


janhenrikoverland:
Now using existing dimension functions + added a test.

janhenrikoverland (approved):

PR #619, opened 2020-09-23
fix: extract color from pattern for trendlines

edoardo:
When the colorSet option is set to patterns, there isn't a color passed
to highcharts, but the index of the pattern.
The trendline tho needs a color to distinguish between the different
trendlines when using multiple data items.
Luckily the Highcharts patterns have a color in their definition, the
same one used in the pattern.

Before, all trendlines were black:
Screenshot 2020-09-23 at 12 15 05

Now, the trendlines have the correct color based on the series:
Screenshot 2020-09-23 at 12 15 37


janhenrikoverland (approved):
LG

PR #644, opened 2020-10-09
feat: add support for scatter charts

martinkrulltott:
Implements [DHIS2-1369](https://jira.dhis2.org/browse/DHIS2-1369)
Design spec [DHIS2-9775](https://jira.dhis2.org/browse/DHIS2-9775)

**Relates to https://github.com/dhis2/data-visualizer-app/pull/1472**

---

### Key features

1. Add basic support for Scatter charts - consts, layout type, icon, description etc
2. Add layout rules for Scatter
3. Add support for generating a Scatter chart in Highcharts

---

### Description

First part of the Scatter implementation. Focuses on the layout ([DHIS2-10003](https://jira.dhis2.org/browse/DHIS2-10003)) and getting an initial draft of a scatter chart up and running.

### TODO and screenshots
_See the DV PR_

---

### POC TODO

-   [ ] Should Scatter have a legend? If not, always hide "show legend" option in DV?
-   [x] Add a proper icon
-   [x] Add a proper description
-   [x] Add rules. e.g. Columns exactly 1 dimension, exactly 2 items?
-   [ ] Default or customisable units for the hover tooltip? e.g. `67 kg, 171 cm`
-   [ ] Should there be an "auto" option (using the name of the dimension) for axis titles to fill the void of having no category labels?
-   [ ] X axis shouldn't start from 0? Or force Y axis to start from 0 too?
-   [ ] In DV, create new options config to hide options that don't apply to Scatter
-   [ ] Add support for colorsets

---

### POC screenshots

_Scatter first draft_
image

_Scatter second draft: with colors and legend_
image

_Bubble first draft_
image

_Bubble second draft: all items passed on same series item, value labels can be toggled_
image


src/visualizations/config/adapters/dhis_highcharts/index.js
```diff
@@ -109,6 +110,15 @@ export default function({ store, layout, el, extraConfig, extraOptions }) {
         },
     }
 
+    // get plot options for scatter
+    if ([VIS_TYPE_SCATTER].includes(_layout.type)) {
```
edoardo: will there be other vis types that need the same `plotOptions`?

edoardo (approved):
I didn't try, but the code looks good.

src/visualizations/config/adapters/dhis_highcharts/index.js
martinkrulltott: Bubble won't use the same, but similar. Will refactor to facilitate both when we add Bubble (actually Bubble was part of the original PoC a few commits back in this PR, but I took it out since we're only doing Scatter now).

martinkrulltott: so `_layout.type === VIS_TYPE_SCATTER` is better, changing it now

PR #650, opened 2020-10-16
feat: legend sets for column + bar (DHIS2-147)

martinkrulltott:

**Relates to https://github.com/dhis2/data-visualizer-app/pull/1355**

---

### Key features

1. Add support for legends to Column and Bar
2. Display legend colors for multi axes and dual category
3. Don't show legend colors for column where the item has type `line`
4. Work with trend lines
5. Show legend info in the tooltip on hover
6. Support dynamic configs in DV

---

### Description

This adds support for legends for Column and Bar exclusively (in addition to Pivot Table, Single Value and Gauge that already supports it). All other vis types should be unaffected. Legends work with multi axes, dual category, trend lines, sort options, cumulative values and all other options available. The exception being multi type Column items which use `line`, which will revert to its original color (as Line shouldn't display legends).

The new Legend tab will be added to Options (DV task).

To support dynamic configs in DV, `isLegendSetType`, `isColumnBasedType` and `isMultiType` were added to be able to better filter vis types based on their traits.

---

### TODO

- [x] Add state in the tooltip for when the value is out of bounds / not applicable.

---

### Screenshots
_basic example of a legend applied to a Column visualization_
image

_with trend line_
image

_with dual category_
image

_with multi axes and multi legendSets. ANC 1 and ANC Total have individual legends. Measles added to showcase a series item without a legend still working. All items are on different axes._
image

_with multi axes and multi type. the legend doesn't apply to the `line` item_
image

_tooltip displays legend info for legend by data item_
image

_tooltip displays legend info for a fixed legend_
image

_tooltip displays `-` when the value is out of bounds of legend_
image

### DIFF
--- a/src/modules/visTypes.js
+++ b/src/modules/visTypes.js
@@ -137,6 +137,8 @@ const dualAxisTypes = [
     VIS_TYPE_AREA,
 ]
 
+const multiTypeTypes = [VIS_TYPE_COLUMN, VIS_TYPE_LINE]
+
 const twoCategoryChartTypes = [
     VIS_TYPE_COLUMN,
     VIS_TYPE_STACKED_COLUMN,
@@ -147,13 +149,26 @@ const twoCategoryChartTypes = [
     VIS_TYPE_STACKED_AREA,
 ]
 
+const columnBasedTypes = [
+    VIS_TYPE_COLUMN,
+    VIS_TYPE_BAR,
+    VIS_TYPE_YEAR_OVER_YEAR_COLUMN,
+    VIS_TYPE_STACKED_COLUMN,
+    VIS_TYPE_STACKED_BAR,
+]
+
 const verticalTypes = [VIS_TYPE_BAR, VIS_TYPE_STACKED_BAR, VIS_TYPE_GAUGE]
+
+const legendSetTypes = [VIS_TYPE_COLUMN, VIS_TYPE_BAR]
 
 export const defaultVisType = VIS_TYPE_COLUMN
 export const isStacked = type => stackedTypes.includes(type)
 export const isYearOverYear = type => yearOverYearTypes.includes(type)
 export const isDualAxisType = type => dualAxisTypes.includes(type)
+export const isMultiType = type => multiTypeTypes.includes(type)
 export const isSingleValue = type => type === VIS_TYPE_SINGLE_VALUE
 export const isTwoCategoryChartType = type =>
     twoCategoryChartTypes.includes(type)
 export const isVerticalType = type => verticalTypes.includes(type)
+export const isLegendSetType = type => legendSetTypes.includes(type)
+export const isColumnBasedType = type => columnBasedTypes.includes(type)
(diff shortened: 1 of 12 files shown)


janhenrikoverland (approved):

src/modules/visTypes.js
```diff
@@ -147,13 +149,26 @@ const twoCategoryChartTypes = [
...
+    VIS_TYPE_COLUMN,
+    VIS_TYPE_BAR,
+    VIS_TYPE_YEAR_OVER_YEAR_COLUMN,
+    VIS_TYPE_STACKED_COLUMN,
+    VIS_TYPE_STACKED_BAR,
+]
+
 const verticalTypes = [VIS_TYPE_BAR, VIS_TYPE_STACKED_BAR, VIS_TYPE_GAUGE]
 
+const legendSetTypes = [VIS_TYPE_COLUMN, VIS_TYPE_BAR]
```
edoardo, line 162: Shouldn't this include PT and SV to be correct?

src/visualizations/config/adapters/dhis_highcharts/index.js
```diff
@@ -135,6 +137,25 @@ export default function({ store, layout, el, extraConfig, extraOptions }) {
...
+    // DHIS2-147 add legendset to Column and Bar
+    /* 
+    ** Note: This needs to go last, after all other data manipulation is done, as it changes 
+    ** the format of the data prop from an array of values to an array of objects with y and color props.
+    */
+    const legendSets = extraOptions.legendSets
+
+    if (legendSets?.length && isLegendSetType(layout.type)) {
+        if (_layout.legendDisplayStrategy === LEGEND_DISPLAY_STRATEGY_BY_DATA_ITEM) {
+            config.series = config.series.map(seriesObj => {
```
edoardo, line 149: Shouldn't this filter out the dual category fake serie?
I guess `applyLegendSet` is not changing the serie's data if values are `null` anyway...

edoardo (approved):
A couple minor comments but it looks good.

src/modules/visTypes.js
martinkrulltott: I'm not sure. All "legend" types include PT, SV and Gauge, but what I wanted to do is create a prop that only includes the "regular" types (regular = that has a vertical y axis and horisontal x axis) that support legends, which are only Column and Bar. 
I discussed this with Jan briefly, this could either be renamed (I can't come up with a good name for it though) or used in conjunction with a list of all "regular" types. But I think this really overcomplicates things 🤔 

Open for suggestions to rename this prop though...

src/visualizations/config/adapters/dhis_highcharts/index.js
martinkrulltott: https://github.com/dhis2/analytics/blob/6f73d83ba4f07464d2e356cfef7bc298ed8ca175/src/visualizations/config/adapters/dhis_highcharts/legendSet.js#L11 filters out both 2C fake series and Column series with custom type `line` 👍

edoardo: Right, I overlooked that one and now I remember we discussed it already 👍

src/modules/visTypes.js
edoardo: I see.
I didn't think about Gauge, the other 2 use different engines for generating the visualization.
Keep it simple.

PR #654, opened 2020-10-20
feat: add function that returns a modified layout based on a given visualization type and original layout [DHIS2-9599]

jenniferarnesen:
Part of implementation for [DHIS2-9599](https://jira.dhis2.org/browse/DHIS2-9599)

PRs for apps:
* DV: https://github.com/dhis2/data-visualizer-app/pull/1360
* Dashboards: https://github.com/dhis2/dashboards-app/pull/1196

---

### New function to get the adapted visualization layout for the provided type

This functionality is being moved from data-visualizer to analytics because the dashboard also needs to use it.  

When a visualization is switched from its original type to a different type (e.g. from table -> column), some layout changes are needed. DV was making these changes, but dashboards wasn't, resulting in different charts between DV and dashboards.

The implementation is therefore moved from DV to analytics and modified from its original form in DV to expose a single function, instead of separate functions for different vis types. The single function accepts the visualization type to provide the layout for, as well as the current layout


janhenrikoverland (approved):
LGTM

PR #655, opened 2020-10-20
feat: implement FileMenu using @dhis2/ui components

edoardo:
Implements [DHIS2-9821](https://jira.dhis2.org/browse/DHIS2-9821)

Requires https://github.com/dhis2/d2-ui/pull/657.
~Requires https://github.com/dhis2/ui/pull/24.~

**Relates to https://github.com/dhis2/data-visualizer-app/pull/1361**

---

### Key features

1. FileMenu
---

### Description

This is a complete rewrite of `d2-ui-file-menu` to use `@dhis2/ui` components and widgets.
It still needs `d2` for the `Open` and `Translate` menu items to work, as they still use old `d2-ui` components.
`Share` uses the upcoming new `SharingDialog` widget of `ui` ([PR here](https://github.com/dhis2/ui/pull/24)).

---

### TODO

-   [x] somehow fix the empty FavoritesDialog (https://github.com/dhis2/d2-ui/pull/657)
-   [x] replace MUI icons with `ui-icons` (waiting to be released)
-   [ ] ~find out if it's possible to pass a proper `d2` mock for the TranslationDialog to be functional~
`d2` is still required for the FavoritesDialog anyway, so no rush to mock it for TranslationDialog.

---

### Known issues

-   **Fixed**: FavoritesDialog opens empty
The reason is that the data fetching for populating the dialog is based on prop change (via the old `componentWillReceiveProps`) which does not trigger anymore because `FavoritesDialog` is now added conditionally, instead of being always rendered and toggled via the `open` prop.
- **Fixed**: There might be an issue with the rounded borders around the menu. @cooper-joe does it look right to you?

---

### Screenshots
Screenshot 2021-01-25 at 10 25 45


src/components/FileMenu/DeleteDialog.js
```diff
@@ -0,0 +1,83 @@
...
+const visualizationMutation = {
+    resource: 'visualizations',
+    id: ({ id }) => id,
+    type: 'delete',
+}
+*/
+export const DeleteDialog = ({ type, id, onClose, onDelete, onError }) => {
+    const [mutate, { error }] = useDataMutation(
+        //        type === 'map' ? mapMutation : visualizationMutation,
+        getMutation(type)
```
edoardo: @amcgee using it this way I get a warning about the mutation not being static.
I guess I didn't understand what we said on Slack about using a function.

amcgee: Try this - it's a tad hacky but works fine:

```
const mutation = useMemo(getMutation(type), [type])
const [mutate, { error }] = useDataMutation(mutation)
```

Otherwise, the "fully static" way to do it would be to have a separate const mutation object for each resource type.

amcgee: (you could even exclude the `[type]` dependency and leave it `[]` to explicitly show that the mutation won't change if the type prop changes - this is what the [`useStaticInput`](https://github.com/dhis2/app-runtime/blob/master/services/data/src/react/hooks/useStaticInput.ts) hook does internally to the `useDataMutation` hook

The "fully static" mechanism would be best since it supports static analysis, but `useMemo` is OK for now

edoardo: Fixed as suggested.

cooper-joe (changes requested):
From the screenshot it looks like the font used is not `Roboto`. Looks like its `Helvetica` or `Arial`. These are `ui` components and should be automatically using the `ui` font, `Roboto`.

edoardo:
> From the screenshot it looks like the font used is not `Roboto`. Looks like its `Helvetica` or `Arial`. These are `ui` components and should be automatically using the `ui` font, `Roboto`.

I took a new screenshot this time in DV app instead of from analytics Storybook.
In the app looks correct.

cooper-joe (changes requested):
Items that open another dialog should have an ellipsis at the end of the label. See the provided design mockup:
image
`Open...`, `Save as...`, `Rename...`, `Translate...`, `Share...` and `Get link...` should all have ellipsis. In the updated screenshot only `Save as...` has the correct ellipsis.

src/components/FileMenu/RenameDialog.js
```diff
@@ -0,0 +1,97 @@
...
+    type: 'update',
+    partial: true,
+    data: ({ name, description }) => ({ name, description }),
+})
+
+export const RenameDialog = ({ type, object, onClose, onRename, onError }) => {
+    const [name, setName] = useState(object.name)
+    const [description, setDescription] = useState(object.description)
+
+    const mutation = useMemo(() => getMutation(type), [])
```
jenniferarnesen, line 31: Curious if there was a performance issue that triggered memoizing the mutation?

src/components/FileMenu/__tests__/FileMenu.spec.js
```diff
@@ -0,0 +1,340 @@
...
+import SharingDialog from '@dhis2/d2-ui-sharing-dialog'
+import { Popover } from '@dhis2/ui'
+
+import { FileMenu } from '../FileMenu'
+import { RenameDialog } from '../RenameDialog'
+import { DeleteDialog } from '../DeleteDialog'
+import { GetLinkDialog } from '../GetLinkDialog'
+import { SaveAsDialog } from '../SaveAsDialog'
+
+describe('The FileMenu component ', () => {
```
jenniferarnesen, line 15: Nice test coverage. Some of these tests where you are checking what is rendered based on props could be accomplished with snapshots. Also, if I understand correctly, people are starting to favor testing-library over enzyme.  That said, the tests are good as is too, if you don't feel like doing this now.  (PS, this comment is probably more relevant for other components like the dialogs, than this one)

src/components/FileMenu/RenameDialog.js
edoardo: The problem was passing the `type` and getting back a different function, otherwise a mutation definition for each `type` should be used which is not feasible.
This is the workaround suggested by @amcgee.

src/components/FileMenu/__tests__/FileMenu.spec.js
edoardo: I will have to work on this component again when the new SharingDialog is ready, I will review the tests then 👍

edoardo:
> `Open...`, `Save as...`, `Rename...`, `Translate...`, `Share...` and `Get link...` should all have ellipsis. In the updated screenshot only `Save as...` has the correct ellipsis.

Fixed (see screenshot).
Also, `Save` / `Save…` is used now depending on an existing visualization or a newly created one.

src/components/FileMenu/RenameDialog.js
amcgee: @edoardo I believe eslint will complain here unless `[type]` is in the dependencies?  We do want to update the mutation if the type is ever changed, however it won't be picked up by the mutation hook - so what you have here is technically correct, just curious why eslint isn't complaining.

cooper-joe:
Looks good @edoardo! The last issue I can see that is not right is the lack of rounded corners on the `menu` container. Is this to avoid the double-bottom-border issue from `ui`?

jenniferarnesen (approved):

edoardo:
> Looks good @edoardo! The last issue I can see that is not right is the lack of rounded corners on the `menu` container. Is this to avoid the double-bottom-border issue from `ui`?

It's now using `Menu` instead of `FlyoutMenu`.
The double border issue happens when the `FlyoutMenu` is used inside a `Popover`.
@cooper-joe  which one is the correct component to use?

cooper-joe:
> > Looks good @edoardo! The last issue I can see that is not right is the lack of rounded corners on the `menu` container. Is this to avoid the double-bottom-border issue from `ui`?
> 
> It's now using `Menu` instead of `FlyoutMenu`.
> The double border issue happens when the `FlyoutMenu` is used inside a `Popover`.
> @cooper-joe which one is the correct component to use?

Both of these components are implementations of the Design System menu component. I think it's best to check with the #ui channel regarding the intended use of the `Menu` or `FlyoutMenu`.

edoardo:
Fixed the double border issue by using `Popper` wrapped in a `Layer` (for supporing the click outside the menu functionality) instead of `Popover`.

cooper-joe (approved):

PR #656, opened 2020-10-21
fix: display info msg when no legend is found (DHIS2-147)

martinkrulltott:

**Relates to https://github.com/dhis2/data-visualizer-app/pull/1355**

---

### Key features

1. Info message for no legend

---

### Description

When the legend option _“Use pre-defined legend per data item”_ is used, each series item is assigned their own legend. In case the series item has no legend assigned to it, the hover tooltip will no display a correct info message to inform the user why no legend is applied.

---

### Screenshots

_previously it was incorrectly showing a an empty row with just a `:` character when no legend was found_
image

_no legend now shows a proper info message_
image

_items with legend still show their regular tooltip as before_
image


edoardo (approved):
👍

PR #679, opened 2020-10-30
test: Support Cypress tests

martinkrulltott:
**Relates to https://github.com/dhis2/data-visualizer-app/pull/1412**

---

### Key features

1. Add `data-test` to components

---

### Description

Increase support for external Cypress tests to run by adding the `data-test` prop to various components

_Note:_ No visual or functional changes, only the `dataTest` / `data-test` props are touched.

### DIFF
--- a/src/components/DataDimension/DataDimension.js
+++ b/src/components/DataDimension/DataDimension.js
@@ -215,6 +215,7 @@ export class DataDimension extends Component {
                     <DataTypes
                         currentDataType={this.state.dataType}
                         onChange={this.onDataTypeChange}
+                        dataTest={'data-dimension-data-types-select-field'}
                     />
                     <Groups
                         dataType={this.state.dataType}
@@ -223,11 +224,13 @@ export class DataDimension extends Component {
                         onGroupChange={this.onGroupChange}
                         onDetailChange={this.onDetailChange}
                         detailValue={this.state.groupDetail}
+                        dataTest={'data-dimension-groups-select-field'}
                     />
                     <FilterField
                         text={this.state.filterText}
                         onFilterTextChange={this.onFilterTextChange}
                         onClearFilter={this.onClearFilter}
+                        dataTest={'data-dimension-filter-input-field'}
                     />
                 </div>
             )
@@ -253,6 +256,7 @@ export class DataDimension extends Component {
                 itemClassName="data-dimension"
                 unselected={unselected}
                 selected={selected}
+                dataTest={'data-dimension-item-selector'}
             >
                 {filterZone()}
             </ItemSelector>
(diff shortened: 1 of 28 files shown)


src/components/DimensionsPanel/List/__tests__/__snapshots__/DimensionItem.spec.js.snap
```diff
@@ -17,6 +17,7 @@ exports[`DimensionItem matches the snapshot 1`] = `
   }
 >
   <DimensionLabel
+    dataTest="undefined-button-pe"
```
jenniferarnesen, line 20: Did you intend to use the word "undefined" in these files?

src/components/ItemSelector/SelectedItems.js
```diff
@@ -242,7 +244,10 @@ export class SelectedItems extends Component {
                     </Droppable>
                 </DragDropContext>
                 <div className="deselect-all-button">
-                    <Button onClick={this.onDeselectAll}>
+                    <Button
+                        onClick={this.onDeselectAll}
+                        dataTest={`${this.props.dataTest}-deselect-all-button`}
```
jenniferarnesen, line 249: Based on the decision (https://github.com/dhis2/notes/blob/master/decisions/2020/11/10-cypress-meeting.md#naming), I'm wondering if the format should instead be:
`-button-deselectall`
So, the component first, then the modifier follows it. Also, suggested but no in the "decisions" try to reserve hyphens for component separators.

src/components/PeriodDimension/PeriodTransfer.js
```diff
@@ -68,12 +68,14 @@ class PeriodTransfer extends Component {
                 <Tab
                     selected={this.state.isRelative}
                     onClick={() => this.onIsRelativeClick(true)}
+                    dataTest={`${this.props.dataTest}-relative-periods-button`}
```
jenniferarnesen, line 71: `-button-relativeperiods` ?

src/components/DataDimension/DataDimension.js
```diff
@@ -215,6 +215,7 @@ export class DataDimension extends Component {
                     <DataTypes
                         currentDataType={this.state.dataType}
                         onChange={this.onDataTypeChange}
+                        dataTest={'data-dimension-data-types-select-field'}
```
jenniferarnesen, line 218: `-datadimension-datatypes-selectfield`  or `-datadimension-datatype-select`?

jenniferarnesen (approved):

src/components/DimensionsPanel/List/__tests__/__snapshots__/DimensionItem.spec.js.snap
martinkrulltott: No, this is not intentional, but I think it comes from the component being tested without a parent or because the parent doesn't provide the dataTest prop. Either way I think this is fine since jest and cypress work in isolation.

martinkrulltott:
As @jenniferarnesen correctly points out in the PR comments above, some of the names needs to be updated to follow our preferred pattern / syntax. As this is still being discussed (as well as if / how to use repo prefixes) I'll revisit the names at a later stage and update them to whatever pattern has been decided then.

PR #690, opened 2020-11-02
refactor: convert d2 to app runtime

edoardo:
Implements [TECH-460](https://jira.dhis2.org/browse/TECH-460)

**Relates to https://github.com/dhis2/data-visualizer-app/pull/1369**

---

### Description

Part of the effort to phase out `d2` in favour of `app-runtime`.
All API interaction is being rewritten to use `app-runtime` and move the query use closer to the component that uses it (wherever the query and fetch function does not need to be exported).

---

### BREAKING CHANGE

Apps need to pass the `dataEngine` in place of `d2`

---

### Known issues

-   [ ] `OrgUnitTree` needs to be rewritten to not use `d2.models` in order to use the converted query functions.


src/components/DynamicDimension/DynamicDimension.js
```diff
@@ -1,29 +1,48 @@
...
-import { apiFetchItemsByDimension } from '../../api/dimensions'
+
+const itemsByDimensionQuery = {
+    itemsByDimension: {
+        resource: 'dimensions',
+        id: ({ id }) => `${id}/items`, // TODO replace with the new way when available
+        params: {
+            fields: 'id,displayName~rename(name)',
+            order: 'displayName:asc',
+            //paging: false, // XXX needed?!
```
jenniferarnesen, line 14: needs addressing?

src/api/organisationUnits-dataEngine.js
```diff
@@ -0,0 +1,115 @@
+import { onError } from './index'
```
jenniferarnesen, line 1: Is the goal to remove the old `api/organisationUnits` as soon as Dashboards is converted and OrgUnitDimension is redone with these functions?

edoardo: Actually this one with the `dataEngine` postfix will be renamed to `organisationUnits` and replace the old one.
For now we need both.
The reason is that `OrgUnitDimension` uses `d2-ui` components and d2 models are required there.
The new OrgUnitTree component in `ui` is still missing some features.

janhenrikoverland:
Open DV, data, event data items, program Antenatal care, select an item, update -> produces this request:

```dimension=dx:lxAQ7Zs9VYR.sWoqcoByYmD,pe:LAST_12_MONTHS```

Should be:

```
dimension=dx:lxAQ7Zs9VYR.sWoqcoByYmD
dimension=pe:LAST_12_MONTHS
```

janhenrikoverland:
Same as above for program indicators.

janhenrikoverland:
DV, data elements, details (operands), select a group, then enter a filter text -> the request has "{$filterText}" rather than the actual text:

```filter=displayName:ilike:{$filterText}```

edoardo:
> DV, data elements, details (operands), select a group, then enter a filter text -> the request has "{$filterText}" rather than the actual text:
> 
> `filter=displayName:ilike:{$filterText}`

Bug fixed.

src/components/DynamicDimension/DynamicDimension.js
janhenrikoverland: Could be fixed as part of https://jira.dhis2.org/browse/DHIS2-9274

janhenrikoverland (approved):
LG, let's get this merged :+1:

PR #730, opened 2020-12-11
feat: use dataEngine (trigger publish)

edoardo:
BREAKING CHANGE: Apps need to pass the dataEngine in place of d2

**See https://github.com/dhis2/analytics/pull/690**


amcgee (approved):
LGTM

PR #754, opened 2021-01-12
fix: add custom title and subtitle for scatter (DHIS2-10170)

martinkrulltott:
Implements [DHIS2-10170](https://jira.dhis2.org/browse/DHIS2-10170)

---

### Key features

1. New title and subtitle format for Scatter

---

### Description

Changes the format of the auto-generated title and subtitle for Scatter. Title now gets `Points: Vertical, Horizontal` and subtitle is always `Filters` (regardless if a user-defined title is present or not). 

---

### Screenshots

_auto generated_
image

_with user-defined title_
image

_with user-defined subtitle (but default title)_
image

### DIFF
--- a/src/visualizations/config/adapters/dhis_highcharts/subtitle/index.js
+++ b/src/visualizations/config/adapters/dhis_highcharts/subtitle/index.js
@@ -3,6 +3,7 @@ import getFilterText from '../../../../util/getFilterText'
 import {
     VIS_TYPE_YEAR_OVER_YEAR_LINE,
     VIS_TYPE_YEAR_OVER_YEAR_COLUMN,
+    VIS_TYPE_SCATTER,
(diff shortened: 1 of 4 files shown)


src/visualizations/config/adapters/dhis_highcharts/subtitle/index.js
```diff
@@ -59,6 +60,9 @@ export default function (series, layout, metaData, dashboard) {
                     Boolean(!dashboard)
                 )
                 break
+            case VIS_TYPE_SCATTER:
+                subtitle.text = dashboard || filterTitle
```
martinkrulltott: Not sure if this is a safe way to do the `dashboard` check (returns `text: true`?), but it's already in use on line 33 so I assume it's ok.

edoardo (approved):

PR #758, opened 2021-01-13
fix: YOY with relative week periods DHIS2-9729

edoardo:
Fixes [DHIS2-9729](https://jira.dhis2.org/browse/DHIS2-9729)

**Relates to https://github.com/dhis2/data-visualizer-app/pull/1560**

---

### Description

Reworked the way data series for YOY charts which uses relative week periods are generated.
To avoid the problems described in the JIRA ticket, where data points are not correctly aligned when comparing week ranges of different years, the logic is now based on a matrix of xAxis index and period ids, which is generated from the analytics responses in the DV plugin.
From the matrix a map object is computed, and used to get the correct xAxis position where to plot a value based on the period id.
This solves both the problem of week 53 that some years have, and the problem of different week ranges for different years when using a relativePeriodDate.   

For more information see the DV PR.


src/visualizations/store/adapters/dhis_highcharts/yearOnYear.js
```diff
@@ -1,13 +1,24 @@
...
-    const serieData = []
+export default function (
+    acc,
+    series,
+    categories,
+    idValueMap,
+    metaData,
+    extraOptions
+) {
+    const serieData = Array.from(
```
jenniferarnesen: Even though you didn't create this variable name, you could fix it to `seriesData`

src/visualizations/store/adapters/dhis_highcharts/yearOnYear.js
```diff
@@ -1,13 +1,24 @@
...
+        { length: extraOptions?.xAxisLabels?.length || 0 },
+        () => null
+    )
 
     categories[0].forEach(categoryItemId => {
+        const position = extraOptions.periodKeyAxisIndexMap[categoryItemId]
```
jenniferarnesen, line 15: Is `extraOptions` assured to exist?  Or does it need to be `extraOptions?.`

src/visualizations/store/index.js
```diff
@@ -31,12 +31,14 @@ export default function ({
         type,
         seriesId = initialSeriesId,
         categoryIds = [initialCategoryId],
+        extraOptions,
```
jenniferarnesen: Related to above comment, should this default to and empty object?

src/visualizations/store/adapters/dhis_highcharts/yearOnYear.js
edoardo: In this case it must exists, otherwise the whole series cannot be built.

src/visualizations/store/adapters/dhis_highcharts/yearOnYear.js
edoardo: Renamed.

src/visualizations/store/index.js
edoardo: Initialized as empty object.

jenniferarnesen (approved):

PR #762, opened 2021-01-14
fix: add default color for values out of bounds (DHIS2-147)

martinkrulltott:
Fix for [DHIS2-147](https://jira.dhis2.org/browse/DHIS2-147)

---

### Key features

1. Add a default color for values that are out of bounds

---

### Description

Previously, when a legend set was applied and a value was out of bounds of its range no color change would be applied. This was changed to apply a default color for those types of values. As discussed on [Slack](https://dhis2.slack.com/archives/G3TL0J145/p1610552136022500?thread_ts=1610541490.014900&cid=G3TL0J145).

---

### Screenshots

_before fix: values out of bounds used their original colors (red and orange)._
image

_after fix: values out of bounds are coloured gray._
image

### DIFF
--- a/src/visualizations/config/adapters/dhis_highcharts/legendSet.js
+++ b/src/visualizations/config/adapters/dhis_highcharts/legendSet.js
@@ -3,6 +3,8 @@ import i18n from '@dhis2/d2-i18n'
 
 import { getLegendByValueFromLegendSet } from '../../../../modules/legends'
 
+const OUT_OF_BOUNDS_COLOR = '#CCCCCC'
+
 const getLegend = (value, legendSet) =>
     value && legendSet ? getLegendByValueFromLegendSet(legendSet, value) : {}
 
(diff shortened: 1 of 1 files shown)


src/visualizations/config/adapters/dhis_highcharts/legendSet.js
```diff
@@ -3,6 +3,8 @@ import i18n from '@dhis2/d2-i18n'
 
 import { getLegendByValueFromLegendSet } from '../../../../modules/legends'
 
+const OUT_OF_BOUNDS_COLOR = '#CCCCCC'
```
edoardo, line 6: Is this a specific color part of the "theme"?
Should it be imported from `@dhis2/ui`?

edoardo (approved):

src/visualizations/config/adapters/dhis_highcharts/legendSet.js
martinkrulltott: Good point. Checked with @cooper-joe, the theme colors are meant for the interface, so we're sticking with a custom color in this case.

PR #787, opened 2021-01-28
feat: new data dim selector (DHIS2-7943)

martinkrulltott:
Implements [DHIS2-7943](https://jira.dhis2.org/browse/DHIS2-7943)
Based on [design spec](https://www.sketch.com/s/bf6d1509-6ff2-4b0c-998d-3b8cd206a47b/a/Qqbeyx) 

**Relates to https://github.com/dhis2/data-visualizer-app/pull/1604**

---

### Key features

1. New Data Dimension selector, using `@dhis2/ui` `Transfer` component
2. Lots of functional changes to the different data types, please refer to the spec in the JIRA ticket for details
3. A bunch of old unit tests removed in favor of Cypress tests in DV
4. Bumps `@dhis2/ui` to latest
5. New data item icons

---

### Description

Well, it's a massive change. The [functional spec](https://jira.dhis2.org/browse/DHIS2-7943) is almost two pages long, plus the design spec, so I can't really summarise it here...

---

### BREAKING CHANGE

`DataDimension` no longer accepts the following props: `DataEngine`, `onDeselect`, `onReorder`

---

### TODO

-   [x] _Deprecate `ItemSelector` as its now unused in our apps?_ No, it's still used by `d2-ui`
-   [x] Implement search for both name or id (needs a backend change)
-   [ ] Program suffix for event data items and program indicators (needs a backend change)
-   [x] Show a warning message after 1500ms when only 1 character has been inputed in a scenario where `/dataItems` would've been called (e.g. global search), since the min char is 2.
-   [ ] DataSets are fetched twice, once for the filter and once for the items, but its essentially the same request. Should this be fetched only once? Or should we paginate the filter request?
-   [x] _Should groups / programs / data sets be lazy loaded, to only load when the single select is opened?_ No, not for 2.36
-   [ ] Replace complex `useState` object with `useReducer`?

---

### Known issues

-   [x] Data sets switch name from `DATASET - METRIC` to `DATASET METRIC` when clicking `Update`, requires a backend change. [DHIS2-10386](https://jira.dhis2.org/browse/DHIS2-10386)
-   [ ] _More known issues listed here: https://github.com/dhis2/data-visualizer-app/pull/1604_

---

### Screenshots

_New selector with global search_
image

_Infinite scrolling with loading spinners_
image

_Loading state for groups_
image


src/components/DynamicDimension/ItemSelector.js
```diff
@@ -95,13 +97,16 @@ const ItemSelector = ({
     const [state, setState] = useState({
         filter: '',
         selected: initialSelected,
+        // FIXME: keeping selected in state is redundant, use the initialSelected prop directly instead
```
edoardo, line 89: Do these FIXME need to be addressed in this PR?

src/components/TransferOption.js
```diff
@@ -13,36 +14,61 @@ export const TransferOption = ({
...
         </div>
     )
+
+    return (
+        <div data-value={value} data-test={dataTest} className="wrapper">
+            {tooltipText ? (
+                <Tooltip
+                    key={`${value}`}
+                    content={tooltipText}
+                    placement={'top-start'}
```
edoardo, line 51: I cannot find `top-start` as possible value for placement in `Tooltip`, but `Popper` which is where the prop is passed ultimately, has it.
I assume it uses `top` but is there any proptype warning in the console?

edoardo (approved):

src/assets/DimensionItemIcons/DataElementIcon.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react'
+
+export default (
+    <svg
```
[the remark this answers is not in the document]
martinkrulltott: No the `data element icon` isn't part of `ui` and this custom svg was provided by @cooper-joe 👍

src/components/DynamicDimension/ItemSelector.js
martinkrulltott: This for the `DynamicDimension`, I found this issue as the `DataDimension` was built upon the `DynamicDimension` and they thus share the same issue. I added the `FIXME` reminder for future reference, but I thought the actual fix was out of scope for this PR.

src/components/TransferOption.js
martinkrulltott: There is a warning in the console (screenshot). @cooper-joe will add the `top-start` to the proptypes at some point. For now it still works, as the prop gets passed to `Popper` as you mentioned above, but we'll have to live with the warning for a bit though.

image

src/components/DynamicDimension/ItemSelector.js
edoardo: 👍

package.json
```diff
@@ -72,5 +71,6 @@
     ],
     "resolutions": {
         "@dhis2/ui": "^6.2.2"
```
edoardo: Missing a comma here?

edoardo (approved):

PR #788, opened 2021-01-28
feat: quartile outlier detection method (DHIS2-10261)

janhenrikoverland:
Implements a small part of [DHIS2-10261](https://jira.dhis2.org/browse/DHIS2-10261)

---

### Key features

Quartile outlier detection method for one-dimensional data.

PR #791, opened 2021-01-29
feat: std dev and quartile outlier detection method (DHIS2-10261)

janhenrikoverland:
Implements a small part of [DHIS2-10261](https://jira.dhis2.org/browse/DHIS2-10261)

Based on https://github.com/dhis2/analytics/pull/788

---

### Key features

X/Y relationship normalization and IQR, Z-score and Modified z-score outlier detection methods.


src/modules/outliers/index.js
```diff
@@ -0,0 +1,89 @@
...
+        .sort((a, b) => (a.normalized < b.normalized ? -1 : 1))
+}
+
+const getPercentiles = (percentile, xyStats) => {
+    const xPercentileValue = xyStats.xSum * (percentile / 100)
+    const yPercentileValue = xyStats.ySum * (percentile / 100)
+
+    return [
+        {
+            name: `${percentile}% of Total X Values`,
```
martinkrulltott, line 37: Should this name use `i18n`?

src/modules/outliers/index.js
```diff
@@ -0,0 +1,89 @@
...
+        },
+    ]
+}
+
+export const getOutlierHelper = (data, userConfig) => {
+    const config = {
+        ...defaultConfig,
+        ...userConfig,
+    }
+    console.log('DATA', data)
```
martinkrulltott, line 60: `console.log`'s should be removed before merging

src/modules/outliers/iqr.js
```diff
@@ -0,0 +1,102 @@
...
+        case 0.5:
+            return pos * 2
+        case 0.25:
+        default:
+            return pos
+    }
+}
+
+export const getQuartileValue = (data, q = 0.25) => {
+    if (data.length < 3) {
```
martinkrulltott, line 20: Do you need 3 or 4 items to calculate the quartile? if 4 then this should be `<= 3`, right?

src/modules/outliers/index.js
martinkrulltott: Also applies to: 

- https://github.com/dhis2/analytics/pull/791/files#diff-0a57656b857d5f0d54d896779c950a3858a6eddda1ec37331f9842e707a1a210R89
- https://github.com/dhis2/analytics/pull/791/files#diff-0a57656b857d5f0d54d896779c950a3858a6eddda1ec37331f9842e707a1a210R134
- https://github.com/dhis2/analytics/pull/791/files#diff-47ad76a2032009b8ae25aa31a4978f146a9f243196cd71f9626f2f56e38d7a89R50

src/modules/outliers/iqr.js
martinkrulltott: Also, what happens if you pass e.g. values like this: `[5, 5, 5, 5]`

PR #807, opened 2021-02-08
fix: series item color set color of middle legend in legendset (DHIS2-147)

martinkrulltott:

---

### Key features

1. Sets the series item to the color of the legend in the middle of a legendset

---

### Description

When a legendset is used for a series item, the colored bullet in the legend needs to reflect the legendset color somehow. This is now done by using the color of the legend in the middle of the legendset. E.g. if a legendset contains legends `1, 2, 3, 4, 5` then the bullet will be in the color of `legend 3`.

---

### Screenshots

_example of the color being applied to series items when [this legend](https://debug.dhis2.org/dev/dhis-web-maintenance/index.html#/edit/otherSection/legendSet/fqs276KXCXi) is in use_
image

### DIFF
--- a/src/visualizations/config/adapters/dhis_highcharts/series/index.js
+++ b/src/visualizations/config/adapters/dhis_highcharts/series/index.js
@@ -110,7 +110,7 @@ function getIdColorMap(series, layout, extraOptions) {
     }
 }
 
-function getDefault(series, layout, isStacked, extraOptions) {
+function getDefault(series, metaData, layout, isStacked, extraOptions) {
     const fullIdAxisMap = getFullIdAxisMap(layout.series, series)
     const idColorMap = getIdColorMap(series, layout, extraOptions)
     const indexColorPatternMap = getIndexColorPatternMap(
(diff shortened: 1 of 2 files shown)


src/visualizations/config/adapters/dhis_highcharts/series/index.js
```diff
@@ -168,10 +168,21 @@ function getDefault(series, layout, isStacked, extraOptions) {
...
+        const legendSet = extraOptions?.legendSets?.find(
+            legendSet =>
+                legendSet.id === metaData.items[seriesObj.id]?.legendSet
+        )
+
         // color
-        seriesObj.color = isYearOverYear(layout.type)
-            ? indexColorPatternMap[index]
-            : idColorMap[seriesObj.id]
+        if (isYearOverYear(layout.type)) {
```
jenniferarnesen, line 177: I guess at this point it is hard to add unit tests for the new code (or extract this to a separate function). Given that isn't feasible, this could be a case where a descriptive comment would help explain the logic.

src/visualizations/config/adapters/dhis_highcharts/series/index.js
```diff
@@ -168,10 +168,24 @@ function getDefault(series, layout, isStacked, extraOptions) {
...
+            // YearOverYear: Fetch colors directly from color sets
+            seriesObj.color = indexColorPatternMap[index]
+        } else if (legendSet && legendSet.legends?.length) {
+            // Legendset: Fetch the middle color of the set
+            seriesObj.color = legendSet.legends.sort(
+                (a, b) => a.startValue - b.startValue
+            )[Math.ceil(legendSet.legends.length / 2) - 1]?.color
```
jenniferarnesen: Is there any possibility of not ending up with a color? (since the code has the `?`)

janhenrikoverland: The `else if` is checking for length so it will always find a legend. The `?` is not necessary I think.

jenniferarnesen: True. Agreed, just drop the `?`

martinkrulltott: Good find. Done 👍

janhenrikoverland (approved):

PR #808, opened 2021-02-10
fix: legendset bugs (DHIS2-147)

martinkrulltott:

---

### Key features

1. Prevent legendset colors to be displayed for non-supported vistypes
2. Apply correct colors to series item bullets

---

### Description

Fixes the two issues described above. See screenshots of before and after for more info:

---

### Screenshots

_1: Before, unsupported type (Stacked column) incorrectly showed the legend color that was applied to a supported type and then persisted during the vis type switch_
image

_1: After, unsupported type (Stacked column) is prevented from receiving the legend color_
image

_2: Before, using a single legend for the visualization didn't color the series item bullets correctly_
image

_2: After, both series items get the proper color_
image

### DIFF
--- a/src/visualizations/config/adapters/dhis_highcharts/series/index.js
+++ b/src/visualizations/config/adapters/dhis_highcharts/series/index.js
@@ -18,10 +18,15 @@ import {
     isYearOverYear,
     VIS_TYPE_LINE,
     VIS_TYPE_SCATTER,
+    isLegendSetType,
 } from '../../../../../modules/visTypes'
 import { hasCustomAxes } from '../../../../../modules/axis'
 import { getAxisStringFromId } from '../../../../util/axisId'
 import { axisHasRelativeItems } from '../../../../../modules/layout/axisHasRelativeItems'
+import {
+    LEGEND_DISPLAY_STRATEGY_BY_DATA_ITEM,
+    LEGEND_DISPLAY_STRATEGY_FIXED,
+} from '../../../../../modules/legends'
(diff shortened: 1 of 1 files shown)


src/visualizations/config/adapters/dhis_highcharts/series/index.js
```diff
@@ -168,16 +173,29 @@ function getDefault(series, metaData, layout, isStacked, extraOptions) {
...
+                LEGEND_DISPLAY_STRATEGY_BY_DATA_ITEM
+            ) {
+                legendSet = legendSets.find(
+                    legendSet =>
+                        legendSet.id === metaData.items[seriesObj.id]?.legendSet
+                )
+            } else if (
+                layout.legendDisplayStrategy === LEGEND_DISPLAY_STRATEGY_FIXED
+            ) {
+                legendSet = legendSets[0]
```
edoardo, line 190: I assume in this case only the selected one in options is in the `legendSets` list?

edoardo (approved):

src/visualizations/config/adapters/dhis_highcharts/series/index.js
martinkrulltott: Yeah this follows the pattern introduced here: https://github.com/dhis2/analytics/blob/402cb78c3e4b8c3dd48ed4e0001854a30af2b4a0/src/visualizations/config/adapters/dhis_highcharts/index.js#L201

PR #816, opened 2021-02-16
fix: account for perpendicular totals row when finalizing totals (DHIS2-10499)

amcgee:
Implements [DHIS2-10499](https://jira.dhis2.org/browse/DHIS2-10499)

---

### Description

Correctly calculate average totals for the final row or column when no perpendicular (row or column) totals are shown.

Also adds two stories for visual testing

**BONUS** this also incidentally solves an issue which would occasionally cause large final-column total values to be "cut off" because their value wasn't taken into account when calculating column widths for adaptive clipping

---

### Screenshots (after)

Screen Shot 2021-02-16 at 17 15 57
Screen Shot 2021-02-16 at 17 16 08


janhenrikoverland (approved):
LG

PR #859, opened 2021-03-08
fix: axis min max values

janhenrikoverland:
Fixes the issue reported during team testing where threshold lines were running out of the viewport.

Before:
Screenshot from 2021-03-05 11-29-58

After (threshold factor: 100):
Screenshot from 2021-03-08 11-45-13

### DIFF
--- a/src/modules/outliers/index.js
+++ b/src/modules/outliers/index.js
@@ -59,7 +59,7 @@ const getExtremeLines = (percentage, xyStats) => {
     return lines
 }
 
-const getMinMaxValue = (outlierHelper, prop) => [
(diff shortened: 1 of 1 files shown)


src/modules/outliers/index.js
```diff
@@ -59,6 +59,17 @@ const getExtremeLines = (percentage, xyStats) => {
     return lines
 }
 
+const getMinMaxValue = (outlierHelper, prop, sortFn) => [
```
martinkrulltott, line 62: might be cleaner to pass `x` or `y` plus `min` or `max` and let the fn use the correct sort function internally to simplify the use. But as is works too 👍

martinkrulltott (approved):

PR #919, opened 2021-05-14
feat: OpenFileDialog component

edoardo:
Implements [DHIS2-11002](https://jira.dhis2.org/browse/DHIS2-11002)

**Relates to https://github.com/dhis2/data-visualizer-app/pull/1737**

---

### Key features

1. Rewrite the `d2-ui-favorites-dialog` in analytics using `ui` and `app-runtime`

---

### Description

Part of the effort to remove the old components that use `MUI` and `d2`.
This PR also implements the new design from Joe for this modal.
The component has the same features as the old one:
* filtering by name, vis type (only for visualizations) and owner
* sorting
* using different api endpoints (visualization, map, etc...)

Full spec here:
https://docs.google.com/document/d/1FyFdsdn4ImlrCLOLlifMTbjxSCFYtd3tBkfPcXkjv_k/edit#heading=h.jd28kwz1a5xn

---

### TODO

-   [x] add tests
-   [x] add dataTest props where https://github.com/dhis2/data-visualizer-app/pull/1737 needs them

---

### Known issues

-   [x] FIXED double refetch when resetting filters with a search term defined, the debounced search value is causing the 2nd refetch after the timeout

---

### Screenshots

Normal list:
Screenshot 2021-05-14 at 11 50 05

Filtered list:
Screenshot 2021-05-14 at 11 50 52

Less than 8 results:
Screenshot 2021-05-14 at 12 11 30

Empty list due to filters:
Screenshot 2021-05-14 at 11 54 39

Loading:
Screenshot 2021-05-14 at 14 31 29

Error view:
Screenshot 2021-05-14 at 12 07 39

Localised dates:
Screenshot 2021-06-10 at 14 47 46


i18n/en.pot
```diff
@@ -188,6 +188,111 @@ msgstr ""
...
+msgid "Anyone"
+msgstr ""
+
+msgid "Only you"
+msgstr ""
+
+msgid "Others"
+msgstr ""
+
+msgid "{{firstItem}}-{{lastItem}} of {{total}}"
```
martinkrulltott: maybe this string can contain something that indicates that its the count / number of items? looks a bit hard to translate as it is right now? 
e.g. `{{firstIndex}}-{{lastIndex}} of {{totalNumberOfItems}}`

src/components/OpenFileDialog/VisTypeFilter.js
```diff
@@ -0,0 +1,37 @@
...
+
+export const VisTypeFilter = ({ selected, onChange }) => (
+    <SingleSelect
+        selected={selected}
+        onChange={({ selected }) => onChange(selected)}
+        prefix={i18n.t('Type')}
+        dense
+    >
+        <SingleSelectOption label={i18n.t('All types')} value="all" />
+        <SingleSelectOption label={i18n.t('All charts')} value="chart" />
```
martinkrulltott: I see that the spec says 

> The first two items, labelled All types and All charts, are separated with a divider.

but aren't we using `visualizations` nowdays?

martinkrulltott: @cooper-joe ?

src/components/FileMenu/FileMenu.js
```diff
@@ -169,6 +157,17 @@ export const FileMenu = ({
                     {i18n.t('File')}
                 </button>
             </div>
+            <OpenFileDialog
+                open={openDialog === 'open'}
```
martinkrulltott: can the `'open'` string be made in to a const instead?

martinkrulltott: also, can we rename `openDialog` to something like `currentDialog`? It's quite confusing to have a prop that refers to "which dialog is open" and a component with a similar name that refers to "open up a file"

martinkrulltott: and then you have `onOpen` and `menuIsOpen`... everything is "open" here 😄  would be great to give things more specific names 🧠 🌀

src/components/OpenFileDialog/VisTypeFilter.js
edoardo: PT is also a visualization type, but here we want to distinguish between charts and PT.

src/components/VisTypeIcon.js
```diff
@@ -0,0 +1,162 @@
...
+    VIS_TYPE_SINGLE_VALUE,
+    VIS_TYPE_SCATTER,
+} from '../modules/visTypes'
+
+export const VisTypeIcon = ({ type, size = 24, ...props }) => {
+    let VisIcon
+
+    switch (type) {
+        case VIS_TYPE_PIVOT_TABLE: {
+            VisIcon = size === 16 ? IconTable16 : IconTable24
```
martinkrulltott: Not sure about this pattern of using a number as an enum. Looks like it would be better to use a const like `LARGE_ICON` or something, since all values except specifically 16 and 24 will work 🤔

martinkrulltott: or if there will ever only be 2 sizes, maybe have a prop called ´useSmall´ or just `small` that just switches to the alternative size?

src/components/OpenFileDialog/VisTypeFilter.js
martinkrulltott: So `Single Value` is a "chart"? 🤔

src/components/OpenFileDialog/PaginationControls.js
```diff
@@ -0,0 +1,30 @@
...
+import React from 'react'
+
+export const PaginationControls = ({ page, pager, onPageChange }) => (
+    <Pagination
+        page={page}
+        pageCount={pager.pageCount}
+        pageSize={pager.pageSize}
+        total={pager.total}
+        onPageChange={onPageChange}
+        onPageSizeChange={Function.prototype}
```
martinkrulltott, line 13: why `Function.prototype`?

martinkrulltott: with `hidePageSizeSelect` used there's no need to provide a `onPageSizeChange`?

src/components/OpenFileDialog/DateField.js
```diff
@@ -0,0 +1,17 @@
+import PropTypes from 'prop-types'
+import React from 'react'
+
+export const DateField = ({ date }) => {
+    const d = new Date(date)
+    const time = `${d.getFullYear()}-${('0' + (d.getMonth() + 1)).slice(-2)}-${(
```
martinkrulltott: Date formats are always tricky since they're different between regions. Should we use a localised date string here instead?

src/components/OpenFileDialog/VisTypeFilter.js
edoardo: Good point.
I think this denomination didn't change since the old favorites dialog was introduced, which happened before SV.
I'm not really sure, maybe these 2 filters don't make much sense anymore?
Perhaps we should have a more dynamic filter where you can check/uncheck the types you want to see in the list...
or, "All charts" should exclude both PT and SV (easy fix).

src/components/VisTypeIcon.js
edoardo: So far it's only 2 sizes that `ui` provides, so perhaps the latter solution is better.

src/components/OpenFileDialog/PaginationControls.js
edoardo: 🤔 i remember it didn't work if the callback prop was not defined.
It seems is still the case:
https://github.com/dhis2/ui/blob/master/packages/widgets/src/Pagination/Pagination.js#L137
and there is no `default` for it.
It looks like a 🐛

src/components/OpenFileDialog/DateField.js
edoardo: Yes, that would definitely be better.
I just copied over from the old component.
I guess we need to add a dependency for this.
I remember I looked into https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat but I don't remember why I didn't end up using it...

src/components/OpenFileDialog/VisTypeFilter.js
martinkrulltott: @cooper-joe

src/components/OpenFileDialog/PaginationControls.js
martinkrulltott: oh ok, if that's the case, would you mind raising a ticket for `ui` and add a comment with the ticket on the `onPageSizeChange` line? Feels like we'll be scratching our heads otherwise in the future 🙂

src/components/OpenFileDialog/DateField.js
martinkrulltott: Yeah, I guess this is generic thing for all our apps. Maybe check in #frontend if someone has a generic solution for date string formatting and localisation?

src/components/OpenFileDialog/PaginationControls.js
edoardo: Opened https://github.com/dhis2/notes/discussions/295

i18n/en.pot
edoardo: This comes from `ui` is not controllable by us.
See: https://github.com/dhis2/ui/blob/master/packages/widgets/src/Pagination/PageSummary.js

martinkrulltott: No, it comes from https://github.com/dhis2/analytics/pull/919/files#diff-f8348f51e29e17d5516c6056d1e558539a04771b7598a2da0b75bdd222ddad5fR17. 
The props comes from the `interpolationObject` yes, but you can change the names of those props before you pass them in to i18n. 
E.g.
```
pageSummaryText={interpolationObject =>
    i18n.t(
         '{{propA}}-{{propB}} of {{propC}}',
        {
            propA: interpolationObject.firstItem,
            propB: interpolationObject.lastItem,
            propC: interpolationObject.total
        }
    )
}
```

To clarify, my point was to make it easier for the translators. So I was just suggesting to change the prop names in the translation string (that ends up in en.pot), not in the code necessarily.

martinkrulltott: Alt, you could even do a named destructuring of `interpolationObject` directly, like `pageSummaryText = ({ firstItem: propA, lastItem: propB, total: propC })`

src/components/OpenFileDialog/CustomSelectOption.js
```diff
@@ -0,0 +1,70 @@
...
+}) => (
+    <div
+        onClick={e => onClick({}, e)}
+        data-value={value}
+        data-label={label}
+        className={cx({ active, disabled })}
+    >
+        {icon}
+        <span className="label">{label}</span>
+        <style jsx>{`
```
martinkrulltott: Extract to a separate css file?

i18n/en.pot
edoardo: Indeed, we can rename those placeholders 👍

src/components/OpenFileDialog/FileList.js
```diff
@@ -0,0 +1,40 @@
...
+                <DataTableCell>
+                    <DateField date={visualization.lastUpdated} />
+                </DataTableCell>
+            </DataTableRow>
+        ))}
+    </>
+)
+
+FileList.propTypes = {
+    data: PropTypes.array,
```
martinkrulltott: add `defaultProps` or `isRequired` to all props?

martinkrulltott: `data.map` above will crash otherwise

martinkrulltott: also consider defining `data` as an array of objects like `{ id, displayName, type, created, lastUpdated }`

yarn.lock
```diff
@@ -1620,7 +1663,18 @@
     "@dhis2/prop-types" "^1.6.4"
     classnames "^2.3.1"
 
-"@dhis2/ui@^6.5.3", "@dhis2/ui@^6.9.0":
+"@dhis2/ui@^6.5.3":
```
martinkrulltott: You can run dedupe (`npx yarn-deduplicate yarn.lock`) to remove the duplicate `ui` versions (see `@dhis2/ui@^6.9.0` below).

src/components/OpenFileDialog/OpenFileDialog.js
```diff
@@ -0,0 +1,444 @@
...
+import PropTypes from 'prop-types'
+import React, { useEffect, useMemo, useState } from 'react'
+import { FileList } from './FileList'
+import { NameFilter } from './NameFilter'
+import { styles } from './OpenFileDialog.styles'
+import { OwnerFilter } from './OwnerFilter'
+import { PaginationControls } from './PaginationControls'
+import { VisTypeFilter } from './VisTypeFilter'
+
+const getResourceFromType = type => `${type}s`
```
martinkrulltott: This is def short and compact, but note that this will only work as long as all backend resources are pluralised by exactly suffixing an `s`. Can't say if we'll ever add new resources or if the existing ones will ever change, but if we think we will, maybe its better to get the resource by mapping key/value 🤷

martinkrulltott: like `const getResourceFromType = type => {visualization: 'visualizations', eventChart: 'eventCharts', eventReport: 'eventReports', map: 'maps'}[type]` (or extract the object as a const if preferred 🙂 )

martinkrulltott:
```
const types = {visualization: 'visualizations', eventChart: 'eventCharts', eventReport: 'eventReports', map: 'maps'}
const getResourceFromType = type => types[type]
```

then you could also use the same list in the proptypes, like 
```
type: PropTypes.oneOf(Object.keys(types))
        .isRequired,
```
to make sure that we support all types that are passed in. And it makes it a lot easier to maintain the list of types and resources 👍

src/components/OpenFileDialog/OpenFileDialog.js
```diff
@@ -0,0 +1,444 @@
...
+    const [page, setPage] = useState(1)
+    const [searchTerm, setSearchTerm] = useState(defaultFilters.searchTerm)
+    const [searchTimeout, setSearchTimeout] = useState(null)
+    const [filters, setFilters] = useState(defaultFilters)
+
+    const formatFilters = ({ searchTerm, createdBy, visType }) => {
+        const queryFilters = []
+
+        switch (createdBy) {
+            case 'byothers':
```
martinkrulltott: why not camel case `byOthers`?

src/components/OpenFileDialog/OpenFileDialog.js
```diff
@@ -0,0 +1,444 @@
...
+    }
+
+    const [{ sortField, sortDirection }, setSorting] = useState({
+        sortField: 'displayName',
+        sortDirection: 'asc',
+    })
+    const [page, setPage] = useState(1)
+    const [searchTerm, setSearchTerm] = useState(defaultFilters.searchTerm)
+    const [searchTimeout, setSearchTimeout] = useState(null)
+    const [filters, setFilters] = useState(defaultFilters)
```
martinkrulltott, line 83: Why do you need to save "searchTerm" in both `searchTerm` and `filters.searchTerm`? One place is enough?
I remember discussing this but can't remember the outcome 🙈

martinkrulltott:
```
    const resetFilters = () => {
        setFilters(defaultFilters)
        setSearchTerm(defaultFilters.searchTerm)
    }
```
e.g. this  makes me curious. Seems a bit unexpected to have to set both here when `filters` already has a  `searchTerm` of its own 🤔

src/components/OpenFileDialog/OpenFileDialog.js
```diff
@@ -0,0 +1,444 @@
...
+        { files: filesQuery },
+        { lazy: true }
+    )
+
+    const resetFilters = () => {
+        setFilters(defaultFilters)
+        setSearchTerm(defaultFilters.searchTerm)
+    }
+
+    const getString = (type, key) => {
```
martinkrulltott: Hm, I see what you're doing here, the texts are similar but not exactly the same, so you want to parse them somehow. 
It looks to me like this can be achieved by using interpolation in i18n instead (e.g. `errorTitle` can be replaced by `i18n.t("Couldn't load {{itemType}}", {itemType: type})`) and just switch or use a ternary on the spot if they differ too much (like the `noDataText` that has a completely different phrasing between the cases). 
Your solution is good because its very scalable, but it seems a bit too complex for just changing a single word?

src/components/OpenFileDialog/OpenFileDialog.js
```diff
@@ -0,0 +1,444 @@
...
+            field: 'created',
+            label: i18n.t('Created'),
+        },
+        {
+            field: 'lastUpdated',
+            label: i18n.t('Last updated'),
+        },
+    ]
+
+    if (type === 'visualization') {
```
martinkrulltott: I think you should set up consts for the supported types so we don't have to match on these strings everywhere.
e.g. `const TYPE_VISUALIZATION = 'visualization'`

src/components/OpenFileDialog/OpenFileDialog.js
```diff
@@ -0,0 +1,444 @@
...
+                                                                        </div>
+                                                                    </>
+                                                                )}
+                                                            </div>
+                                                        </div>
+                                                    </Box>
+                                                </DataTableCell>
+                                            </DataTableRow>
+                                        )}
+                                    {Boolean(
```
martinkrulltott: won't this literally type "false" in the ui if the condition isn't met?
use a ternary with a null fallback instead?

src/components/OpenFileDialog/OpenFileDialog.js
```diff
@@ -0,0 +1,444 @@
...
+                                                data.files[
+                                                    getResourceFromType(type)
+                                                ]
+                                            }
+                                            onSelect={onFileSelect}
+                                        />
+                                    )}
+                                </DataTableBody>
+                            </DataTable>
+                            {Boolean(
```
martinkrulltott: same question as above

src/components/OpenFileDialog/OpenFileDialog.js
```diff
@@ -0,0 +1,444 @@
...
+                    )}
+                    <style jsx>{styles}</style>
+                </Box>
+            </ModalContent>
+        </Modal>
+    )
+}
+
+OpenFileDialog.defaultProps = {
+    open: false,
```
martinkrulltott: `currentUser` missing a default prop or `isRequired`

src/components/OpenFileDialog/OwnerFilter.js
```diff
@@ -0,0 +1,24 @@
...
+import React from 'react'
+
+export const OwnerFilter = ({ selected, onChange }) => (
+    <SingleSelect
+        selected={selected}
+        onChange={({ selected }) => onChange(selected)}
+        prefix={i18n.t('Created by')}
+        dense
+    >
+        <SingleSelectOption label={i18n.t('Anyone')} value="all" />
```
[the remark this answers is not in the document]
martinkrulltott: This will be hard to translate in languages that has a different flow to the sentence "Created by anyone". Can we change it to become sort of a two part thing, e.g. "Creator: Anyone" instead of a sentence? So they can be translated separately.  @cooper-joe

src/components/OpenFileDialog/OpenFileDialog.js
martinkrulltott: Realising now that it's used for the input field, so its necessary to keep it as a separate state prop as you do. I suggest renaming it to e.g. `searchTermInput` to make it less confusing and clear that its tied to the input field itself.

src/components/OpenFileDialog/OpenFileDialog.js
martinkrulltott: Actually, this works perfectly fine (new to me!), but since we use the ternary in other places it might be best to just follow that path

src/components/OpenFileDialog/OpenFileDialog.js
martinkrulltott: or, better yet, should be `isRequired`

src/components/FileMenu/FileMenu.js
edoardo: Renamed `currentDialog` but left the other 2.
`onOpen` for consistency with the other callback props which follow the button names, and for backwards compatibility so there is no need to change the apps using this component.

src/components/OpenFileDialog/VisTypeFilter.js
edoardo: Added a TODO note.

i18n/en.pot
edoardo: Renamed the placeholders.

src/components/VisTypeIcon.js
edoardo: Refactored to use `useSmall`.

src/components/OpenFileDialog/CustomSelectOption.js
edoardo: Done.

src/components/OpenFileDialog/OwnerFilter.js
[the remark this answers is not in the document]
edoardo: Added a TODO comment.

src/components/OpenFileDialog/FileList.js
edoardo: Fixed.

yarn.lock
edoardo: Done.

src/components/OpenFileDialog/OpenFileDialog.js
edoardo: Renamed to `nameFilterValue` and `setNameFilterValue` to avoid confusion with `searchTerm` which is now only used in `filters` and is debounced.

src/components/OpenFileDialog/OpenFileDialog.js
edoardo: As discussed, I moved the function in `utils.js` but kept the current approach until we have the 3rd app that will use this and will need its own strings.

src/components/OpenFileDialog/OpenFileDialog.js
edoardo: Rewritten removing the casting and testing for `length > 0` instead.

src/components/OpenFileDialog/OpenFileDialog.js
edoardo: Same as above.

src/components/OpenFileDialog/OpenFileDialog.js
edoardo: Required it is.

src/components/OpenFileDialog/DateField.js
edoardo: I tried using `moment` but `en` by default uses the American format, which is then different from the current default.
Although is technically more correct, I'm wondering if this can be a problem.
We don't seem to have the possibility to choose the variant of English in the user settings, so just by having `en` `moment` seem to default to the American variant.

src/components/OpenFileDialog/OpenFileDialog.js
edoardo: Refactored this using constants for the select options, exported from the `CreatedByFilter` component (renamed from `OwnerFilter`).

src/components/OpenFileDialog/FileList.js
```diff
@@ -32,9 +32,17 @@ export const FileList = ({ type, data, onSelect }) => (
 )
 
 FileList.propTypes = {
-    data: PropTypes.array,
-    type: PropTypes.string,
-    onSelect: PropTypes.func,
+    data: PropTypes.arrayOf(
+        PropTypes.shape({
+            created: PropTypes.string,
```
martinkrulltott: Should these be `isRequired` too?

src/components/OpenFileDialog/OpenFileDialog.js
edoardo: Added constants for the AO types and a map object in `utils.js` with the endpoint names.

src/components/OpenFileDialog/FileList.js
edoardo: I was thinking about that, I think it should be safe since they are listed in the query's fields, so they should always be there.

src/components/OpenFileDialog/OpenFileDialog.js
edoardo: Done. See comment above.

src/components/OpenFileDialog/FileList.js
edoardo: Fixed.

src/components/OpenFileDialog/DateField.js
edoardo: Fixed using `moment`.
As mentioned above, this has the effect that for `en` the date format is the American one, which is different compared to what we had before, which was a fixed format `yyyy-mm-dd`.

martinkrulltott (approved):

PR #952, opened 2021-06-17
feat: show indicator type in SV visualization (DHIS2-7420)

edoardo:
**PUBLISHED v20.1.0-alpha.2 FOR TESTING**

Implements [DHIS2-7420](https://jira.dhis2.org/browse/DHIS2-7420)

**Relates to https://github.com/dhis2/data-visualizer-app/pull/1830**

---

### Key features

1. Add indicator type info in SV visualizations

---

### Description

Show % symbol for indicators with factor 100 and a text for others (ie. "Per thousand")

---

### TODO

-   [ ] tests

---

### Known issues

-   [ ] in dashboard app when resizing the item, the text under the value can become too small. Agreed with @janhenrikoverland to merge what we have and see if it's a problem we need to address later.
-   [x] ~translations don't seem to work for the indicator type text returned in the analytics response~ (This was never an issue. DB language must be set in user settings for this to work.) Screenshot 2021-08-17 at 08 57 12

---

### Screenshots

DV per cent:
Screenshot 2021-06-22 at 11 51 34

DV per thousand:
Screenshot 2021-06-22 at 11 51 56

DV per hundred thousand:
Screenshot 2021-06-22 at 11 52 20

Dashboard item:
Screenshot 2021-06-17 at 12 00 29

### DIFF
--- a/src/visualizations/config/adapters/dhis_dhis/index.js
+++ b/src/visualizations/config/adapters/dhis_dhis/index.js
@@ -2,6 +2,8 @@ import getSubtitle from './subtitle'
 import getTitle from './title'
 import getValue from './value'
 
+export const INDICATOR_FACTOR_100 = 100
+
 export default function ({ store, layout, extraOptions }) {
     const data = store.generateData({
         type: layout.type,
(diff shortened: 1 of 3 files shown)


cooper-joe:
A couple of design suggestions:
- use a different color for the text label ("Per thousand") - I suggest `grey600`.
- reduce the margin between the value and the `%` symbol, if possible. Right now it looks like they are separate, and it makes it slightly difficult to parse. It should, ideally, just look like a typed figured `103.5%`

martinkrulltott (approved):

PR #984, opened 2021-07-07
feat: fixed headers in PT (DHIS2-11057)

edoardo:
**PUBLISHED v20.1.0-alpha.4 FOR TESTING**

Implements [DHIS2-11057](https://jira.dhis2.org/browse/DHIS2-11057)

**Relates to https://github.com/dhis2/data-visualizer-app/pull/1797**

---

### Key features

1. implement fixed column headers and fixed row headers in pivot table
2. update storybook with a decorator that allows to set some options to all stories, so it's possible to test them without replicating the same stories

---

### Description

The fixed headers feature in PT is controlled by 2 options in DV (under the Style tab, see related PR).
The options can be used separately or combined together.
The storybook provide an easy way to simulate toggling of these options so all stories can be tested out with all possible combinations of the options.

The technical solution is to use `position: sticky` on the `th` cells that need to be fixed, together with specific values for `top` and `left` positioning of the cells.

There is some refactoring around the `isColumnHeader` prop, which is now renamed to the more generic `isHeader` as the same prop is used also on row headers, also now "converted" to `th` instead of `td`.

---

### TODO

-   [ ] _task_

---

### Known issues
-   [ ] Border-collapse: separate is needed to avoid seeing white borders when scrolling, but the border is double the thickness, couldn’t find another way…
-   [x] FIXED Label on column header is not centered when scrolling horizontally
-   [ ] Column headers sliding under when fixed but no vertical scrolling
-   [ ] Dimension header cells labels not centered when scrolling (because the cell width does not change)

---

### Screenshots

Quick video demo:

https://user-images.githubusercontent.com/150978/124780392-d76ccf00-df42-11eb-8e80-2b7d17ca53b6.mov

### DIFF
--- a/src/components/PivotTable/PivotTableColumnHeaderCell.js
+++ b/src/components/PivotTable/PivotTableColumnHeaderCell.js
@@ -37,10 +37,7 @@ export const PivotTableColumnHeaderCell = ({
                     minWidth: width,
                 }
 
-                if (
-                    engine.options.fixedColumnHeaders ||
-                    engine.options.fixedRowHeaders
-                ) {
+                if (engine.options.fixedColumnHeaders) {
                     style.top =
                         level * (engine.fontSize + engine.cellPadding * 2 + 2)
                     // left value for the column header cells should be sum of row headers' width when engine.options.fixedRowHeaders is true
--- a/src/components/PivotTable/PivotTableColumnHeaderCell.js
+++ b/src/components/PivotTable/PivotTableColumnHeaderCell.js
@@ -30,24 +30,41 @@ export const PivotTableColumnHeaderCell = ({
                     header.span === 1 &&
                     engine.isSortable(index)
 
+                const style = {
+                    cursor: isSortable ? 'pointer' : 'default',
+                    width,
+                    maxWidth: width,
+                    minWidth: width,
+                }
+
+                if (
+                    engine.options.fixedColumnHeaders ||
+                    engine.options.fixedRowHeaders
+                ) {
+                    style.top =
+                        level * (engine.fontSize + engine.cellPadding * 2 + 2)
+                    // left value for the column header cells should be sum of row headers' width when engine.options.fixedRowHeaders is true
+                    style.left = engine.options.fixedRowHeaders
+                        ? engine.rowHeaderPixelWidth
+                        : 0
+                }
+
                 return (
                     <PivotTableCell
-                        isColumnHeader
-                        classes={
+                        isHeader
+                        classes={[
                             header.label &&
                             header.label !== 'Total' &&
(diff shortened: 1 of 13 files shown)


src/components/PivotTable/PivotTableColumnHeaderCell.js
```diff
@@ -30,24 +30,41 @@ export const PivotTableColumnHeaderCell = ({
...
+                    cursor: isSortable ? 'pointer' : 'default',
+                    width,
+                    maxWidth: width,
+                    minWidth: width,
+                }
+
+                if (
+                    engine.options.fixedColumnHeaders ||
+                    engine.options.fixedRowHeaders
+                ) {
```
amcgee: Shouldn't this condition just be `if (engine.options.fixedColumnHeaders)` ?

src/modules/pivotTable/clipAxis.js
```diff
@@ -1,11 +1,29 @@
 import times from 'lodash/times'
 
-export const clipAxis = ({ position, size, step, totalCount, headerCount }) => {
+export const clipAxis = ({
+    position,
+    size,
+    step,
+    totalCount,
+    headerCount,
+    fixedColumnHeaders,
```
amcgee: `clipAxis` is generic and used for both rows and columns, should this just be `fixedHeaders` ?

amcgee (commented):
Looks good!  Added some comments

src/modules/pivotTable/clipAxis.js
edoardo: I'm not sure, what should `fixedHeaders` be then?
`headerCount` is looking at the columns only, and `count` (which is the number of rows) is then adjusted removing the number of header rows at the top of the table, but only when fixed columns is enabled.
I could rename the variable, but I think it's important to know which one of the 2 options is actually passed there.

src/components/PivotTable/PivotTableColumnHeaderCell.js
edoardo: 🤔 I tried removing the or condition, and it seems to work the same, so yes, I don't think it's needed.
Removed.

edoardo: I see now why it didn't make a difference, it does set `top` and `left` on the column header cells, but the `position` is not set to fixed when the `fixedColumnHeaders` option is not passed.

src/modules/pivotTable/clipAxis.js
edoardo: It turns out we don't need to change this part, I reverted the changes to this file as we discovered a bug.

janhenrikoverland (approved):

PR #1038, opened 2021-09-07
fix: SV indicator type increase font size and spacing DHIS2-7420

edoardo:
Implements [DHIS2-7420](https://jira.dhis2.org/browse/DHIS2-7420)

---

### Key features

1. Increase font size for the indicator type text under the single value
2. Increse the spacing between the valuer and the indicator type text

### Screenshots

Before:
Screenshot 2021-09-07 at 16 00 22

After:
Screenshot 2021-09-07 at 16 04 21

### DIFF
--- a/src/visualizations/config/generators/dhis/singleValue.js
+++ b/src/visualizations/config/generators/dhis/singleValue.js
@@ -107,7 +107,7 @@ const generateDashboardItem = (config, legendSet) => {
             config.formattedValue,
             config.subText,
             legendSet,
-            20
+            40
         )
     )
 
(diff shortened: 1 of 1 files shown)


src/visualizations/config/generators/dhis/singleValue.js
```diff
@@ -106,7 +106,8 @@ const generateDashboardItem = (config, legendSet) => {
             config.value,
             config.formattedValue,
             config.subText,
-            legendSet
+            legendSet,
+            40
```
jenniferarnesen, line 110: maybe use a constant throughout the file?

edoardo: There are 2 different values, do you mean to have 2 constants each used in one place?

jenniferarnesen: So it's just coincidence that both use the number 40?

jenniferarnesen (approved):

src/visualizations/config/generators/dhis/singleValue.js
edoardo: Yes, one is for the font size, the other is for the `y` attribute on the SVG text node to control the spacing with the value above.

PR #1040, opened 2021-09-10
fix: avoid double thickness border

edoardo:
### Key features

1. do not use `border-collapse: separate` which causes the table border to have double thickness

---

### Description

`border-collapse: separate` was introduce to solve the issue of seeing the table content in the border space between cells when using fixed headers and scrolling.
The solution implemented in this PR is to manually add border rules for `th`, `td` and `table` to simulate a 1px border all around.
This solves the issue described above, but there is an issue, read below.

---

### Known issues

-   [x] When using fixed headers where the bottom/right border of the header disappears when scrolling.
Screenshot 2021-09-10 at 14 29 32

---

### Screenshots

Before:
Screenshot 2021-09-10 at 14 27 53

After:
Screenshot 2021-09-10 at 14 28 58


src/components/PivotTable/styles/PivotTable.style.js
```diff
@@ -9,21 +9,54 @@ import {
...
         color: ${colors.grey900};
     }
+
     table {
         border-spacing: 0;
-        border-collapse: separate;
         white-space: nowrap;
         box-sizing: border-box;
         text-align: center;
+        border: 1px solid #b2b2b2;
```
martinkrulltott: ui color instead of hex value?

martinkrulltott: same for the border a few lines below

src/components/PivotTable/PivotTableContainer.js
```diff
@@ -1,19 +1,41 @@
...
+                data-test="visualization-container"
+            >
+                <style jsx>{tableStyle}</style>
+                {width === 0 || height === 0 ? null : (
+                    <table
+                        className={classnames({
+                            'fixed-headers':
+                                engine.options.fixColumnHeaders &&
+                                engine.options.fixRowHeaders,
+                            'fix-column-headers':
```
martinkrulltott: why both `fixed` and `fix`? use `fixed` everywhere?

```suggestion
                            'fixed-column-headers':
```

martinkrulltott (approved):
LGTM, but there are 2 very minor things that can be improved.

src/components/PivotTable/styles/PivotTable.style.js
edoardo: There is no such color in `ui`, but I added a `const` instead.

PR #1036, opened 2021-09-06
fix: adjust row clipping partitions after sorting [DHIS2-11724]

amcgee:
Implements [DHIS2-11724](https://jira.dhis2.org/browse/DHIS2-11724)

---

### Key features

1. fix an issue while scrolling vertically in a large table after sorting on a column

PR #1041, opened 2021-09-14
fix: also reset row partitions on clearSort

amcgee:
Follow-on to #1036


amcgee:
Closing in favor of #1036

PR #1050, opened 2021-09-20
fix: use lighter grey for No data text in SV

edoardo:
Part of the fix for [DHIS2-11810](https://jira.dhis2.org/browse/DHIS2-11810)

---

### Key features

1. Use lighter grey for the No data text
2. Refactored the params passed around functions, use an object instead of passing all in a list

---

### Screenshots
Screenshot 2021-09-20 at 16 27 05

### DIFF
--- a/src/visualizations/config/generators/dhis/singleValue.js
+++ b/src/visualizations/config/generators/dhis/singleValue.js
@@ -39,11 +39,13 @@ const generateValueSVG = ({
         svgValue.setAttribute('y', y)
     }
 
-    const fillColor = legendSet
-        ? getColorByValueFromLegendSet(legendSet, value)
-        : formattedValue === noData.text
-        ? colors.grey600
-        : colors.grey900
+    let fillColor = colors.grey900
+
+    if (legendSet) {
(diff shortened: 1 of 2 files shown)


src/visualizations/config/generators/dhis/singleValue.js
```diff
@@ -34,6 +41,8 @@ const generateValueSVG = (value, formattedValue, subText, legendSet, y) => {
 
     const fillColor = legendSet
         ? getColorByValueFromLegendSet(legendSet, value)
+        : formattedValue === noData.text
```
martinkrulltott: technically correct but maybe not the most readable to have nested ternaries. consider just keeping it simple with a couple of `if`'s for readability.

martinkrulltott (approved):
LGTM. Could we add a simple cypress test for the no data text to make sure we dont run in to this in the future? (e.g. create a SV, hijack the request and remove the data, expect no data to be shown).

src/visualizations/config/generators/dhis/singleValue.js
edoardo: Changed.

PR #1055, opened 2021-09-28
feat: visualization options in Analytics

martinkrulltott:
Relates to https://github.com/dhis2/data-visualizer-app/pull/1907

### Key features

1. Moves `VisualizationOptions` to Analytics, to be reused in the new upcoming ER app


src/components/Options/VisualizationOptions.js
```diff
@@ -0,0 +1,160 @@
...
+    tabSectionOptionItem,
+    tabSectionOptionToggleable,
+    tabSectionToggleableSubsection,
+    tabSectionOptionComplexInline,
+    tabSectionOptionText,
+    tabBar,
+    tabContent,
+    tabSectionOptionIcon,
+} from './styles/VisualizationOptions.style.js'
+//import styles from './styles/VisualizationOptionsManager.module.css'
```
edoardo: Remove?

src/components/Options/VisualizationOptions.js
```diff
@@ -0,0 +1,160 @@
...
+
+export class VisualizationOptions extends Component {
```
edoardo: Convert to functional component with hooks?

src/components/Options/VisualizationOptions.js
```diff
@@ -0,0 +1,160 @@
...
+                {modalContent.styles}
+            </Modal>
+        )
+    }
+}
+
+VisualizationOptions.propTypes = {
+    optionsConfig: PropTypes.array.isRequired,
+    onClose: PropTypes.func,
+    onPrimaryClick: PropTypes.func,
```
edoardo: This could be called `onUpdate`.

edoardo (approved):

src/index.js
```diff
@@ -266,3 +266,5 @@ export {
 } from './components/PeriodDimension/utils'
 export { getRelativePeriodsOptionsById } from './components/PeriodDimension/utils/relativePeriods'
 export { getFixedPeriodsOptionsById } from './components/PeriodDimension/utils/fixedPeriods'
+
+export { default as VisualizationOptions } from './components/Options/VisualizationOptions'
```
edoardo, line 270: I see we mostly export the const directly instead of the default which needs to be aliased, any special reason for this?

edoardo (approved):

src/index.js
martinkrulltott: I just followed the same pattern as the other components, e.g. DataDimension and 
PeriodDimension. I think the unnamed export comes from connected components being exported in that way. But since these aren't connected I don't see why they can't be exported directly 🤷

edoardo: 👍

PR #1062, opened 2021-10-04
feat: add support for ER (TECH-704)

edoardo:
Partially implements [TECH-704](https://jira.dhis2.org/browse/TECH-704)

---

### Key features

1. add support for ER in FileMenu
2. add support for ER in OpenFileDialog
3. filter only ER AO with `dataType` EVENT (Line list ERs)

---

### Description

The new ER app will use the same generic components from analytics as DV/Maps.
Specifically, the `FileMenu` and the `OpenFileDialog`.
These need to be updated in order to support ER AOs.
In particular, the `OpenFileDialog` for 2.38 need to filter only the Line list ERs (`dataType=EVENT`), and inform the user about the filtering.

---

### Screenshots
Screenshot 2021-10-04 at 09 59 43

### DIFF
--- a/i18n/en.pot
+++ b/i18n/en.pot
@@ -5,8 +5,8 @@ msgstr ""
 "Content-Type: text/plain; charset=utf-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 "Plural-Forms: nplurals=2; plural=(n != 1)\n"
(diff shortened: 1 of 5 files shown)


martinkrulltott (approved):
`en.pot` changes?

PR #1080, opened 2021-10-27
feat: export formatValue function

edoardo:
This formats the numeric values according to the given digit group separator.

**Relates to https://github.com/dhis2/data-visualizer-app/pull/853**

---

### Key features

1. export `renderValue` so ER can use it to format numeric values according to the digit group separator option

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -241,7 +241,7 @@ export {
 
 // Modules: pivotTable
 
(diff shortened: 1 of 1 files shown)


martinkrulltott (approved):
Quite straight-forward to review and approve this 🙂  but maybe we should consider renaming the exported prop to give it some context to what it actually does? `renderValue` doesn't really have any semantic meaning to it...

PR #1081, opened 2021-11-01
feat: new org unit dimension (TECH-707)

martinkrulltott:
Implements [TECH-707](https://jira.dhis2.org/browse/TECH-707)

**Relates to https://github.com/dhis2/data-visualizer-app/pull/1939**

---

### Key features

1. Implements the new org unit tree from `ui`

---

### Description

Replaces the old `d2` based org unit tree with the org unit tree plus `Checkbox`es and `MultiSelect`s from `ui`.
New props are passed in (see _breaking changes_ below), which are more in line with the other dimensions (data, period, dynamic dimensions).

Removed the unit test for org unit dimension, as it didn't really do anything and these dimensions are better tested through Cypress tests in the respective apps.

Note: `organisationUnits-dataEngine.js` renamed to `organisationUnits.js` and the old file was removed (was only used by the old org unit tree and wasn't exported externally anymore either)

---

### BREAKING CHANGE

changed props for org the unit dimension
_removed_: `current`, `d2`, `displayNameProperty`, `ouItems`, `onDeselect`
_added_: `root`, `selected`
_unaffected_: `onSelect`

---

### TODO

- [ ] Remove `apiFetchOrganisationUnits`?
- [ ] Fetch `root` from backend settings instead of passing it in as a prop?

---

### Known issues

- [x] `Failed prop type: Invalid org unit path 'ImspTQPwCqd' supplied to 'OrganisationUnitTree.initiallyExpanded[0]'` is thrown on first open. Need to discuss how to handle multiple roots first.

---

### Screenshots

image


src/api/organisationUnits.js
```diff
@@ -1,64 +1,115 @@
...
-        ','
-    )}`
-
-    return d2.Api.getApi()
-        .get(url)
-        .then(({ organisationUnits }) => organisationUnits[0])
-        .catch(onError)
+const orgUnitLevelsQuery = {
+    resource: 'organisationUnitLevels',
+    params: ({ displayNameProp = 'displayName' }) => ({
```
HendrikThePendric: It could be worthwhile to also default to an empty object so the fn can be called without any args.
```suggestion
    params: ({ displayNameProp = 'displayName' } = {}) => ({
```

src/api/organisationUnits.js
```diff
@@ -1,64 +1,115 @@
...
-export const apiFetchOrganisationUnits = (d2, displayNameProperty) => {
-    const fields = [
-        'id',
-        'path',
-        `${displayNameProperty}~rename(displayName)`,
-        'children::isNotEmpty',
-    ]
+const orgUnitGroupsQuery = {
+    resource: 'organisationUnitGroups',
+    params: ({ displayNameProp = 'displayName' }) => ({
```
HendrikThePendric: Same as above:
```suggestion
    params: ({ displayNameProp = 'displayName' } = {}) => ({
```

src/api/organisationUnits.js
```diff
@@ -1,64 +1,115 @@
...
+        userDataViewFallback: true,
         paging: false,
+    },
+}
+
+const orgUnitsQuery = {
+    resource: 'organisationUnits',
+    params: ({ displayNameProp }) => ({
+        fields: `id,path,${displayNameProp}~rename(displayName),children::isNotEmpty`,
         level: 1,
```
HendrikThePendric, line 32: This `level: 1,` was already there so it must be correct, but I don't understand how having a hardcoded `level` is correct if this is a generic way to get a list of org-units.

HendrikThePendric (approved):
Left two minor suggestions and a question, but LGTM.

edoardo (commented):
In general I would like to reach a conclusion on which way we should use `dataEngine`, `dataQuery`, perhaps that can be addressed in a separate PR for this component.

src/api/organisationUnits.js
martinkrulltott: I think this fetches the "highest level" for the user (which is dynamic, could be one level for you and another for me, depending on our root org units). Not entirely sure, but I believe @janhenrikoverland knows more about how the backend treats this.

martinkrulltott:
No rush to merge this and as its a breaking change its worth figuring out everything before merging. So we can hold off until we've got @edoardo's concerns answered if needed.

src/components/PeriodDimension/FixedPeriodSelect.js
```diff
@@ -1,5 +1,5 @@
-import propTypes from '@dhis2/prop-types'
 import { SingleSelectField, SingleSelectOption } from '@dhis2/ui'
+import propTypes from 'prop-types'
```
edoardo: Typo?
```suggestion
import PropTypes from 'prop-types'
```

and also rename in `propTypes` below.

src/components/TransferOption.js
```diff
@@ -1,6 +1,6 @@
-import propTypes from '@dhis2/prop-types'
 import { Tooltip } from '@dhis2/ui'
 import cx from 'classnames'
+import propTypes from 'prop-types'
```
edoardo:
```suggestion
import PropTypes from 'prop-types'
```

and also rename in `propTypes` below.

src/components/OrgUnitDimension/OrgUnitDimension.js
```diff
@@ -1,276 +1,336 @@
...
+                        count: numberOfGroups,
+                        defaultValue: '{{count}} group',
+                        defaultValue_plural: '{{count}} groups',
+                    })
+                )
+            }
+            userOrgUnits.forEach(orgUnit => {
+                parts.push(orgUnit.name || orgUnit.displayName)
             })
+            summary = i18n.t('Selected: ') + parts.join(', ')
```
edoardo: I have a similar thing for the sharing summary.
I don't know what is the best solution, as for some languages this "hardcoded" comma-separated list of text can be a problem.
I don't have a better suggestion tho...

edoardo (approved):
The `dataEngine` usage can be addressed in a separate PR once we have a decision.

src/components/OrgUnitDimension/OrgUnitDimension.js
martinkrulltott: I had a long chat with Joe about this, because it started as "A," + "B," + "C"+ " selected, which is bad because they're all part of the same sentence. However with the "Selected: " + "parts", at least the sentence is split up. I could do the same as you do though, inject the comma list with interpolation instead of just concating the string.

src/__demo__/OrgUnitDimension.stories.js
```diff
@@ -0,0 +1,136 @@
...
+        const [selected, setSelected] = useState([])
+
+        return (
+            <OrgUnitDimension
+                selected={selected}
+                onSelect={response => setSelected(response.items)}
+                roots={defaultRootOrgUnits}
+            />
+        )
+    })
```
HendrikThePendric, line 26: The StoryBook `storiesOf` API has not been deprecated, and there are no concrete plans to deprecate it in the near future ([source](https://github.com/storybookjs/storybook/blob/master/lib/core/docs/storiesOf.md)). However StoryBook recommends writing stories in the [Component Story Format (CSF)](https://storybook.js.org/docs/react/api/csf).

Personally I find CFS easier to write and there are some additional advantages in case you want to create interactive demoes (with prop-tables). 

This isn't a critical point at all, probably more of a recommendation for future PRs.

HendrikThePendric (approved):
LGTM. 

_Looking at some of the code in this PR made me wonder why we don't export custom hooks instead of imperative `apiFetch` functions. But that change needs to be discussed more thoroughly and is out-of-scope for this PR._

src/components/OrgUnitDimension/OrgUnitDimension.js
HendrikThePendric: Curtesy of @mediremi :
```js
const formatList = items => {
    // Wrap Intl.ListFormat in try/catch as DHIS2 locales are not always ISO 639 compliant
    try {
        const formatter = new Intl.ListFormat(i18n.language, {
            style: 'long',
            type: 'conjunction',
        })
        return formatter.format(items)
    } catch (error) {
        return items.join(', ')
    }
}
```
([Source](https://github.com/dhis2/user-app/blob/8b323c366b9e18421856f47e1ba76a4d2ad5d3db/src/containers/UserList/OrganisationUnitFilter.js#L10-L21))

martinkrulltott: Looks good, added 👍  Thanks!

PR #1088, opened 2021-11-19
fix: use loading state instead of disabling subscribe button

edoardo:
### Key features

1. use the `ui` buttons `loading` prop while waiting for the subscribe/unsubscribe mutations to complete

### Known issues

-   [ ] on slow connections, after the mutation completes the button returns to its previous look, but its label changes after a while because it's controlled by a different query that is refetched after the mutation completes.
So for a moment, the button does not have the correct label: ie. the user is subscribed but the button still says "Subscribe" until the visualization query comes back with the updated value for `subscribed`.

---

### Screenshots
Screenshot 2021-11-19 at 09 22 47


HendrikThePendric (approved):
Good idea!

PR #1105, opened 2021-12-07
feat: use new translations dialog (TECH-873)

edoardo:
Implements [TECH-873](https://jira.dhis2.org/browse/TECH-873)

---

### Key features

1. new TranslationDialog with new design and new stack
2. use the new TranslationDialog in the FileMenu

---

### Description

This component will eventually be moved in `ui`.
The functionality is the same as before, allowing a user to add translations for properties on an object (like an AO).
For analytics apps the dialog opens from the FileMenu and typically allows to add translations for the name and description of the AO.
The component is generic and should work with any object and any number of fields to translate.

---

### TODO

-   [x] storing the selected locale in sessionStorage

---

### Known issues

-   [ ] The input labels are defined by each field name passed to `fieldsToTranslate`, the problem is how to get translations for those since they are dynamic? In the old implementation the translations are "static" and must be provided within the app's translations, ie. if the filed is `name` and the locale is `it` the translation would be `Nome`.
In this PR, as a temporary solution, no translation is used, but the `field` is converted to have first letter uppercase and camel case turned into spaces.

---

### Screenshots

Translations loading, Save button disabled:
Screenshot 2021-12-09 at 10 53 50

Translations loaded, no locale selected, Save button disabled:
Screenshot 2021-12-09 at 14 00 53

Translations loaded, locale selected, Save button enabled:
Screenshot 2021-12-09 at 14 01 06


src/components/TranslationDialog/TranslationModal/TranslationForm.js
```diff
@@ -0,0 +1,138 @@
...
+    translations,
+    onUpdateTranslations,
+}) => {
+    const [newTranslations, setNewTranslations] = useState(translations)
+    const [translationLocale, setTranslationLocale] = useState()
+
+    const camelCaseToUnderscores = field =>
+        field
+            .replace(/[a-z][A-Z]/g, match =>
+                [match.chartAt(0), match.chartAt(1)].join('_')
```
janhenrikoverland: chartAt should be charAt

src/components/TranslationDialog/TranslationModal/TranslationForm.js
```diff
@@ -0,0 +1,138 @@
...
+            property: camelCaseToUnderscores(field).toUpperCase(),
+            value: translation,
+        }
+
+        const translationIndex = getTranslationIndexForField(field)
+
+        if (translationIndex !== -1) {
+            newTranslations.splice(translationIndex, 1, newTranslation)
+        } else {
+            newTranslations.push(newTranslation)
```
janhenrikoverland: 65 and 67: do not modify the state variable

src/components/TranslationDialog/TranslationModal/TranslationForm.js
edoardo: Good catch 😅 👍 
Fixed.

src/components/TranslationDialog/TranslationModal/TranslationForm.js
edoardo: Yep, figured it didn't work, fixed.

src/components/TranslationDialog/TranslationModal/LocalesSelect.js
```diff
@@ -0,0 +1,50 @@
...
+export const LocalesSelect = ({ onChange, selected }) => {
+    const { data, fetching } = useDataQuery(query)
+
+    return (
+        <SingleSelect
+            prefix={
+                selected ? i18n.t('Translating to') : i18n.t('Choose a locale')
+            }
+            onChange={({ selected }) => onChange(selected)}
+            loading={fetching}
```
HendrikThePendric, line 22:
```suggestion
            loading={fetching}
            loadingText={i18n.t('Loading locales')}
```
I've looked into showing the selected locale name while the available locales request is pending, but I don't think this is possible, or at least not easy at all. I think your current implementation is probably the best we can do if we want the select to load it's own options.

Alternatively, we could load the locales in the modal, and wait for both the locales and the translations to load before showing the form. Then we can just make this a regular `SingleSelect`.

Both solutions are fine, but for the current implementation I think a loading text would be a good addition.

src/components/TranslationDialog/TranslationModal/TranslationForm.js
```diff
@@ -0,0 +1,227 @@
...
+                    </Button>
+                    <Button
+                        primary
+                        onClick={save}
+                        loading={saveInProgress}
+                        disabled={!translationLocale}
+                    >
+                        {i18n.t('Save translations')}
+                    </Button>
+                </ButtonStrip>
```
HendrikThePendric: We could consider extracting this part into a component because it's used both in the loader UI and the form UI.

HendrikThePendric (approved):
Code looks good. I've added some comments.
I couldn't really run the example locally because there is no story. Could you please add one?
I'm approving already but still hoping you will add a story ~and a loading text to the locale select~ _(not useful)_.

src/components/TranslationDialog/TranslationModal/LocalesSelect.js
HendrikThePendric: Sorry, I just came to realise the `loadingText` shows up in the dropdown, not in the "input" part of the component. This suggestion doesn't really add much value at all. Please ignore it.

src/components/TranslationDialog/TranslationModal/TranslationForm.js
edoardo: Done in b7589f6.

edoardo:
> I couldn't really run the example locally because there is no story. Could you please add one?

There are actually stories for the `FileMenu` so the `TranslationDialog` can be tested by opening the `FileMenu` and then clicking `Translate`.
We could add more dedicated stories tho.

src/components/TranslationDialog/TranslationModal/TranslationForm.js
```diff
@@ -0,0 +1,216 @@
...
+
+        const updatedTranslations = [...newTranslations]
+
+        if (translationIndex !== -1) {
+            updatedTranslations.splice(translationIndex, 1, newTranslation)
+        } else {
+            updatedTranslations.push(newTranslation)
+        }
+
+        setNewTranslations(updatedTranslations)
```
HendrikThePendric: The current code creates a shallow copy first, which is then mutated. We could rewrite this in accordance with the Redux immutable update patterns where they advice [using a map](https://redux.js.org/usage/structuring-reducers/immutable-update-patterns#updating-an-item-in-an-array) for updating array items.
```suggestion
        setNewTranslations(
            translationIndex === -1
                ? [...newTranslations, newTranslation]
                : newTranslations.map((translation, index) =>
                      index === translationIndex ? newTranslation : translation
                  )
        )
```

src/components/TranslationDialog/TranslationModal/TranslationForm.js
```diff
@@ -0,0 +1,216 @@
...
+    }
+
+    const translationsMutation = useMemo(
+        () => getTranslationsMutation(resource),
+        []
+    )
```
HendrikThePendric: We should not use `useMemo` for this [according to the docs](https://reactjs.org/docs/hooks-reference.html#usememo): 
> **You may rely on useMemo as a performance optimization, not as a semantic guarantee.**
> In the future, React may choose to “forget” some previously memoized values and recalculate them on next render, e.g. to free memory for offscreen components. Write your code so that it still works without `useMemo` — and then add it to optimize performance.

As we've done in the events-reports app, we should use `useRef` instead.

src/components/TranslationDialog/TranslationModal/useTranslationsResults.js
```diff
@@ -0,0 +1,28 @@
...
+    })
+
+    const { data, fetching, refetch } = useDataQuery(
+        translationsQueryRef.current,
+        {
+            onError: error => showError(error),
+        }
+    )
+
+    const { show: showError } = useAlert(error => error.message, {
```
HendrikThePendric: I'm not sure how correct this is right now, but in the past I have ran into situations where the error didn't have a `message` property (I believe the error contained a more complex object with an error report 🤷 ). Perhaps we should implement a fallback message?
```suggestion
    const fallbackMsg = i18n.t('Could not load translations')
    const { show: showError } = useAlert(error => error.message || fallbackMsg, {
```

HendrikThePendric (approved):

HendrikThePendric:
> * The input labels are defined by each field name passed to `fieldsToTranslate`, the problem is how to get translations for those since they are dynamic?

I seem to remember that these labels *should* be translated server side. I had a look at some of the backend PRs related to JIRA epic [DHIS2-8636](https://jira.dhis2.org/browse/DHIS2-8636) like [this one](https://github.com/dhis2/dhis2-core/pull/6352), and it does seem to be the case (not quite sure).

Otherwise, perhaps we can check with Viet how we should tackle this best.

src/components/TranslationDialog/TranslationModal/TranslationForm.js
edoardo: 👍 
Done.

src/components/TranslationDialog/TranslationModal/TranslationForm.js
edoardo: Done.

src/components/TranslationDialog/TranslationModal/useTranslationsResults.js
edoardo: Added.

edoardo:
> I seem to remember that these labels _should_ be translated server side. I had a look at some of the backend PRs related to JIRA epic [DHIS2-8636](https://jira.dhis2.org/browse/DHIS2-8636) like [this one](https://github.com/dhis2/dhis2-core/pull/6352), and it does seem to be the case (not quite sure).
> 
> Otherwise, perhaps we can check with Viet how we should tackle this best.

Yeah, not sure how/where to get those translations tho 🤔 
The old d2-ui implementation required some translations to be present in the apps.
We still have the `i18n_old` folder for that, which we can actually remove now 🗒️

edoardo:
After talking with @janhenrikoverland we decided to merge and fix the translation issue when we know how to do it.

PR #1106, opened 2021-12-08
chore: lint with latest cli-style

martinkrulltott:
Similar pattern to dhis2/event-reports-app#877, but Analytics had a lot more implicit imports that had to be handled manually.

plus added missing proptypes to `PivotTableClippedAxis`


HendrikThePendric (approved):
👍

PR #1114, opened 2021-12-15
feat: remove d2 and implement a generic provider component and hook

edoardo:
Part of  [TECH-705](https://jira.dhis2.org/browse/TECH-705)

**Required by https://github.com/dhis2/event-reports-app/pull/886**

---

### Key features

1. Implement `CachedDataQueryProvider` component
2. Implement `useCachedDataQuery` hook

---

### Description

`OpenFileDialog` needs the current user information which was taken from `d2`.
To remove the need for `d2` in the whole FileMenu and this in the apps that use it, another way of getting the user information is needed.
These provider component and hook answer that need.
Eventually this could be implemented at app platform level, since it's something always needed and currently there are multiple requests to the `/me` endpoint in different parts of the platform and app.

Apps can define multiple queries to be executed in parallel by the provider and the data can be accessed anywhere via the hook.
This makes it generic and flexible, for example ER has 2 queries for user info and user settings.
Because the queries are defined at app level, it's easy to accommodate small differences (ie. in `fields`).

Example of usage in ER:
https://github.com/dhis2/event-reports-app/pull/886/files#diff-cc53afa34ba20baa7d10a5a4b9ef436dae14fc7d0609856e4b5aaaf1ac8429e9

---

### BREAKING CHANGE

`currentUser` is a new prop required in `FileMenu` in place of `d2`.
Once https://github.com/dhis2/analytics/pull/1105 is merged, the currently required`d2` prop can be removed.

---

### TODO

-   [x] remove `d2` once https://github.com/dhis2/analytics/pull/1105 is merged
-   [ ] approve [this PR](https://github.com/dhis2/event-reports-app/pull/886) that fixes the breaking change in ER


src/components/CachedDataQueryProvider.js
```diff
@@ -0,0 +1,60 @@
...
+                    </CenteredContent>
+                </Layer>
+                {!loading && (
+                    <CachedDataQueryCtx.Provider value={data}>
+                        {children}
+                    </CachedDataQueryCtx.Provider>
+                )}
+            </>
+        )
+    }
```
HendrikThePendric: This part should be simplified a lot because we don't support refetching or SWR as I did in my initial version. Since the data is fetched once the component mounts and cannot be refetched, the code could simply look like this:
```suggestion
    if (loading) {
        return (
            <Layer translucent>
                <CenteredContent>
                    <CircularLoader />
                </CenteredContent>
            </Layer>
        )
    }
```

HendrikThePendric (changes requested):
Looks good. I think one part could be simplified though.

src/components/CachedDataQueryProvider.js
edoardo: Done.

HendrikThePendric (approved):

PR #1115, opened 2021-12-16
Feat/current-user-hook-SUGGESTIONS

HendrikThePendric:
There's a few points I am trying to make here:
- I don't think we need to use effect or state. We just fetch some data and want to apply some transformations to it.
- Since we are not rendering a loader or error in this component, but always render ther children, I think it is important we do provide access to it, so it should be part of he provider's value


edoardo:
We went for the generic solution. Closing.

PR #1116, opened 2021-12-16
feat: remove d2 and implement generic data provider

HendrikThePendric:
Instead of the current user provider + hook, we could also implement something like this.


jenniferarnesen:
@HendrikThePendric If I understand correctly, this generic hook could replace this, which we have in some of our apps:

```
 <SystemSettingsProvider>
    <UserSettingsProvider>
        <App />
    </UserSettingsProvider>
 </SystemSettingsProvider>
```

Could you add some example code to the description on what an implementation would look like?

HendrikThePendric:
> If I understand correctly, this generic hook could replace this, which we have in some of our apps:
> 
> ```jsx
>  <SystemSettingsProvider>
>     <UserSettingsProvider>
>         <App />
>     </UserSettingsProvider>
>  </SystemSettingsProvider>
> ```
> 
> Could you add some example code to the description on what an implementation would look like?

@jenniferarnesen so the code snippet you just shared would be simply replaced by this:
```jsx
<CachedDataQueryProvider query={query}>
    <App />
</CachedDataQueryProvider>
```
And query would be declared somewhere outside of the component and most likely look a little bit like this:
```js
const query = {
    currentUser: {
        resource: 'me',
        params: {
            fields: 'id,username,displayName~rename(name)',
        },
    },
    userSettings: {
        resource: 'userSettings',
        params: {
            key: ['keyUiLocale', 'keyAnalysisDisplayProperty'],
        },
    },
    systemSettings: {
        resource: 'systemSettings',
        params: {
            key: [
                'keyDateFormat',
                'keyAnalysisRelativePeriod',
                'keyAnalysisDigitGroupSeparator',
                'keyIgnoreAnalyticsApprovalYearThreshold',
                'keyHideBiMonthlyPeriods',
                'keyHideDailyPeriods',
                'keyHideMonthlyPeriods',
                'keyHideWeeklyPeriods',
                'keyHideBiWeeklyPeriods',
            ],
        },
    },
}
```
Then the app can use this data via the `useCachedDataQuery()` hook. By default this data will have the exact same shape as returned by the `useDataQuery` hook. 

This will mean some inconveniently named properties like `data.userSettings.userSettings`. And that is what the `dataTransformation` is for: you can pass a function that cleans up the data as it is received.

What is also a good idea is to derive more specific hooks from the generic `useCachedDataQuery` hook, i.e.
```js
const useSystemSettings = () => {
    const cachedData = useCachedDataQuery()
    // assume the data was not sanitised
    return cachedData.systemSettings.systemSetting
}
```

Hope this clears things up a bit.

HendrikThePendric:
BTW: one thing I have not looked into at all is how this would behave if we were to nest more of these of providers. I mean they are not really meant to be nested, but it would be worth looking into before we add something like this to a lib.

src/components/CachedDataQueryProvider.js
```diff
@@ -0,0 +1,60 @@
...
+                )}
+            </>
+        )
+    }
+
+    if (error) {
+        const fallbackMsg = i18n.t('This app could not retrieve required data.')
+
+        return (
+            <NoticeBox title={i18n.t('Network error')}>
```
HendrikThePendric: Oops forgot to add the error prop.
```suggestion
            <NoticeBox error title={i18n.t('Network error')}>
```

HendrikThePendric (commented):
I can't approve or request changes because I started this PR, but I think this looks good to me. But we do need to make sure we add the error prop to the notice-box otherwise the error will look like info.

HendrikThePendric:
**Update**: I actually committed that suggestion just now. There's not much to be discussed about it 😊

PR #1119, opened 2022-01-04
fix: fetch translations for field labels from api TECH-873

edoardo:
Fix for [TECH-873](https://jira.dhis2.org/browse/TECH-873)

See issue described in https://github.com/dhis2/analytics/pull/1105.

### Key features

1. Fetch translations for the labels of the input fields

---

### Description

Use the `api/i18n` endpoint to request the translations for the fields passed to `fieldsToTranslate` and used as labels of the input fields in the form.

The translation keys can be fetched from the schema, but it's not
possible to filter only the required properties, so instead, transform the
field name to the format used for the translation keys, see `fieldName` and `i18nTranslationKey` in the schemas.


janhenrikoverland (approved):
LG

PR #1126, opened 2022-01-11
feat: add support for vis type filter for ER

edoardo:
Implements [TECH-704](https://jira.dhis2.org/browse/TECH-704)

**Required by https://github.com/dhis2/event-reports-app/pull/898**
**Required by https://github.com/dhis2/data-visualizer-app/pull/1985**

---

### Key features

1. Add LineList vis type
2. export All and All charts types
3. refactor `FileMenu` and `OpenFileDialog` for passing the list of vis types which enables the vis type filter in the dialog
4. vis type description are now moved to the apps

---

### Description

Add Line List vis type and refactor to allow filtering based on the
given list of vis types instead of looking at the AO type, the logic is
now moved outside analytics and to the apps.

Apps that don't need the vis type selector (ie. Maps) simply omit the new props `filterVisTypes` and `defaultFilterVisType`.
Apps that do need the vis type selector pass those props.
`filterVisTypes` is an array with the following structure:
`[{ type: VIS_TYPE_ALL } { type: VIS_TYPE_CHARTS, insertDivider: true }, { type: VIS_TYPE_PIVOT_TABLE, disabled: true }]`
The additional options are:
1) disabled: disables the option and adds a Tooltip (see screenshot)
2) insertDivider: adds a divider after the option

---

### BREAKING CHANGE

Apps that want the vis type filter enabled need to pass new props: `filterVisTypes` and `defaultFilterVisType`.

### Screenshots

No vis type filter:
Screenshot 2022-01-11 at 16 42 35

Vis type filter with divider:
Screenshot 2022-01-11 at 16 42 55

Vis type filter with disabled option:
Screenshot 2022-01-11 at 16 43 10

Vis type filter with tooltip for disabled option:
Screenshot 2022-01-11 at 16 43 19


src/modules/visTypes.js
```diff
@@ -49,6 +52,9 @@ export const visTypeDisplayNames = {
     [VIS_TYPE_YEAR_OVER_YEAR_COLUMN]: i18n.t('Year over year (column)'),
     [VIS_TYPE_SINGLE_VALUE]: i18n.t('Single value'),
     [VIS_TYPE_SCATTER]: i18n.t('Scatter'),
+    [VIS_TYPE_LINE_LIST]: i18n.t('Line list'),
```
martinkrulltott, line 55: This will break DV because of https://github.com/dhis2/data-visualizer-app/blob/eb4b868284491979c1b09bea56c5ea8eb697ca7f/packages/app/src/components/VisualizationTypeSelector/VisualizationTypeSelector.js#L57. Should be added to the breaking change notes and would be good to come up with a solution for this in DV before merging.

edoardo: Yep, the plan is to move this from Draft when both DV and ER PRs are ready.
I will check also Maps, but I think for that one there should not be any breaking change.

martinkrulltott: Sounds good. My point about adding this to the breaking change notes is incorrect though, as adding something to an array shouldn't be considered a "breaking change" per se, as the API remains the same.. so if (i.e. when) things don't work as expected in DV that's just DV's fault for being so tightly coupled to Analytics.

src/__demo__/OpenFileDialog.stories.js
```diff
@@ -16,36 +36,48 @@ const user = {
...
-            type="map"
-            onClose={Function.prototype}
-            onFileSelect={onFileSelect}
-            onNew={Function.prototype}
-            open={true}
-            currentUser={user}
-        />
-    </Provider>
-))
+const DVfilterVisTypes = [
```
martinkrulltott: I get the idea behind testing "real world" scenarios like this, however this needs to be maintained when the app updates to not become outdated. So from Analytics perspective, wouldn't it be better to just test generic cases (e.g. types + divider, types + disabled etc) and then let DV and ER do their own app specific testing in Cypress to make sure the output matches the spec?

martinkrulltott: Hm realising now that these aren't tests, they're stories. Still, the same principle applies. Better to have Analytics show off the available input in a generic fashion rather than tying the stories to apps?

src/components/OpenFileDialog/CustomSelectOption.js
```diff
@@ -1,28 +1,43 @@
...
+            onClick={(e) => onClick({}, e)}
+            data-value={value}
+            data-label={label}
+            className={cx({ active, disabled })}
+        >
+            {icon}
+            <span className={cx({ label: icon })}>{label}</span>
+            <style jsx>{styles}</style>
+        </div>
+        {insertDivider ? <MenuDivider dense /> : null}
```
martinkrulltott:
```suggestion
        {insertDivider && <MenuDivider dense />}
```

src/components/OpenFileDialog/CustomSelectOption.js
```diff
@@ -1,28 +1,43 @@
...
+    </>
 )
 
+export const CustomSelectOption = (props) =>
+    props.disabled ? (
+        <Tooltip content={i18n.t('Not supported by this app yet')}>
```
martinkrulltott, line 34: Just a note... Coupling the `disabled` prop to the case that something isn't supported but hinting that it will be in the future works for our (only) use-case of this, but also feels a bit limiting from Analytics perspective. (what if someone has a use-case where `disabled` is set for a different reason?). However extracting the message to a separate prop could be overcomplicating things as well...

src/components/OpenFileDialog/utils.js
```diff
@@ -2,8 +2,6 @@ import i18n from '@dhis2/d2-i18n'
 
 export const AO_TYPE_VISUALIZATION = 'visualization'
```
martinkrulltott, line 3: `AO_TYPE_VISUALIZATION` should be used further down where it says `case 'visualization': {` (but GH didn't let me comment on that line). Same for `AO_TYPE_EVENT_VISUALIZATION`

martinkrulltott (commented):
Generally looks good, just a few minor comments

src/components/OpenFileDialog/CustomSelectOption.js
edoardo: I guess we could have a different prop `unsupported` or something, which disables the option and adds this Tooltip, while the `disabled` one simply disables the option with no Tooltip.

src/components/OpenFileDialog/utils.js
edoardo: Fixed.

src/components/OpenFileDialog/CustomSelectOption.js
edoardo: Fixed.

src/__demo__/OpenFileDialog.stories.js
edoardo: This stories rely on the api to show the list of AOs.
So the AO type is passed and at that point it makes sense to have the matching vis types list you would expect to see based on the AO type.
To make the whole thing generic we would have to mock everything and not use the api.
Perhaps for another PR 😁

src/components/OpenFileDialog/utils.js
```diff
@@ -29,64 +31,52 @@ export const getTranslatedString = (type, key) => {
...
+        ),
+        noFilteredDataText: i18n.t(
+            "No event visualizations found. Try adjusting your search or filter options to find what you're looking for."
+        ),
+        newButtonLabel: i18n.t('New event visualization'),
+    },
 }
+
+export const getTranslatedString = (type, key) =>
+    Object.keys(texts).includes(type) ? texts[type][key] : texts[NO_TYPE][key]
```
martinkrulltott:
```suggestion
    (texts[type] || texts[NO_TYPE])[key]
```

edoardo: Fixed

martinkrulltott (approved):

PR #1130, opened 2022-01-17
feat: add event visualization ao type (TECH-949)

janhenrikoverland:
Implements [TECH-949](https://jira.dhis2.org/browse/TECH-949)

Adds EVENT_VISUALIZATION as an AO type.

### DIFF
--- a/src/components/OpenFileDialog/OpenFileDialog.js
+++ b/src/components/OpenFileDialog/OpenFileDialog.js
@@ -33,6 +33,7 @@ import {
     getTranslatedString,
     AO_TYPE_VISUALIZATION,
     AO_TYPE_EVENT_REPORT,
+    AO_TYPE_EVENT_VISUALIZATION,
     AOTypeMap,
 } from './utils.js'
 import {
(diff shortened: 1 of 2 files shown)


src/components/OpenFileDialog/OpenFileDialog.js
```diff
@@ -124,7 +125,10 @@ export const OpenFileDialog = ({
         }
 
         // for ER 2.38 only show line list ER types
-        if (type === AO_TYPE_EVENT_REPORT) {
+        if (
+            type === AO_TYPE_EVENT_REPORT ||
+            type === AO_TYPE_EVENT_VISUALIZATION
+        ) {
```
edoardo, line 131: This part has been removed here:
https://github.com/dhis2/analytics/pull/1126/files#diff-785f63826da7b159716f77c9f8a6ddee63e24884dfb3ae542445f66b890b0a76L127
Because now we rely only on the `visType` for the filtering, not the AO type.

This will result in a conflict.

edoardo (approved):

PR #1152, opened 2022-02-11
feat: repeatable events (TECH-894)

martinkrulltott:
Implements [TECH-894](https://jira.dhis2.org/browse/TECH-894)

**Relates to https://github.com/dhis2/event-reports-app/pull/925**

---

### Key features

1. Support repetition when creating dimensions
2. Support repetition in the analytics request

---

### Description

Adds support for repetition in `createDimension` and the analytics request.
e.g. 
repetition for `A03MvHHogjR.a3kGcGDCuk6` would result in the following request:
`dimension:ou:ImspTQPwCqd,w75KJ2mc4zz,zDhUuAYrxNC,A03MvHHogjR[1].a3kGcGDCuk6,A03MvHHogjR[2].a3kGcGDCuk6,A03MvHHogjR[3].a3kGcGDCuk6,A03MvHHogjR[4].a3kGcGDCuk6,A03MvHHogjR[0].a3kGcGDCuk6`

---

### Known issues

-   [ ] doesn't work on the backend so this is currently untested

### DIFF
--- a/src/api/analytics/AnalyticsRequest.js
+++ b/src/api/analytics/AnalyticsRequest.js
@@ -63,10 +63,18 @@ class AnalyticsRequest extends AnalyticsRequestDimensionsMixin(
                 dimension += `:${d.filter}`
             }
 
-            request = request.addDimension(
-                dimension,
-                d.items?.map((item) => item.id)
-            )
+            if (d.repetition?.indexes?.length) {
+                d.repetition.indexes.forEach((index) => {
+                    request = request.addDimension(
+                        dimension.replace(/\./, `[${index}].`)
+                    )
+                })
+            } else {
+                request = request.addDimension(
+                    dimension,
+                    d.items?.map((item) => item.id)
+                )
+            }
         })
 
         // extract filters from visualization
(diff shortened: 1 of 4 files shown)


src/api/analytics/AnalyticsRequest.js
```diff
@@ -63,10 +63,18 @@ class AnalyticsRequest extends AnalyticsRequestDimensionsMixin(
...
             }
 
-            request = request.addDimension(
-                dimension,
-                d.items?.map((item) => item.id)
-            )
+            if (d.repetition?.indexes?.length) {
+                d.repetition.indexes.forEach((index) => {
+                    request = request.addDimension(
+                        dimension.replace(/\./, `[${index}].`)
```
edoardo, line 69: I think the dimension's `programStageId` is not going to be prefixed to the `dimensionId` but passed via `programStage: { id: [id] }` object nested in the dimension object.
So this will be changed.

edoardo (approved):

src/api/analytics/AnalyticsRequest.js
martinkrulltott: True, the`replace` will have to be changed but I suppose the end result will be the same format regardless, e.g. `A03MvHHogjR[1].a3kGcGDCuk6`.

edoardo: Yes, what goes in the analytics request is correct.

edoardo: Actually, we don't need any change, as the dimension is already formatted to `programStageId.dimensionId` before this new code is executed 👍 .

PR #1173, opened 2022-03-14
feat(FileMenu): use custom labels in dialog titles TECH-1035

edoardo:
Implements [TECH-1035](https://dhis2.atlassian.net/browse/TECH-1035)

---

### Key features

1. allow for custom labels in dialog titles

---

### Description

FileMenu dialogs titles depend on the `fileType` passed.
Before we just used the `fileType` directly, but with the new Line list app, we need a custom label for `eventVisualization`,
also, `eventVisualization` and `eventReport` don't work as well because of the camel casing.

---

### Known issues

-   [x] should these labels be translated? Currently are just interpolated in the title (ie. Delete {{fileType}}) so they are not translated and marking the label part for translation is not ideal either because the whole string should rather be translated as one.
**Labels are translated and interpolated with the rest of the sentences.**

---

### Screenshots

Before:
Screenshot 2022-03-14 at 15 04 14
Screenshot 2022-03-14 at 15 04 23
Screenshot 2022-03-14 at 15 04 28

After:
Screenshot 2022-03-14 at 14 46 10
Screenshot 2022-03-14 at 14 46 16
Screenshot 2022-03-14 at 14 46 03


src/components/FileMenu/utils.js
```diff
@@ -1,24 +1,39 @@
...
+            return i18n.t('event report')
+        case FILE_TYPE_EVENT_VISUALIZATION:
+            return i18n.t('line list')
+        case FILE_TYPE_MAP:
+            return i18n.t('map')
+        case FILE_TYPE_VISUALIZATION:
+            return i18n.t('visualization')
         default:
-            return `${fileType}s`
+            return fileType
```
jenniferarnesen, line 28: Are there any other types? Maybe this could just be a map instead of a switch with a default. Though I guess that's more risky if something invalid did get passed in.

jenniferarnesen (approved):

src/components/FileMenu/utils.js
edoardo: The types in the switch are the valid ones from the prop type validation.
I guess we could not handle a default, or it can be done also with a map.

PR #1174, opened 2022-03-15
fix(TranslationDialog): remove warnings

edoardo:
Implements [TECH-1038](https://dhis2.atlassian.net/browse/TECH-1038)

---

### Key features

1. Avoid warning about wrong prop type and value for `selected`
2. Avoid warning about non static mutation

---

### Screenshots

Before:
Screenshot 2022-03-15 at 12 04 57

Screenshot 2022-03-15 at 12 04 36


src/components/TranslationDialog/TranslationModal/TranslationForm.js
```diff
@@ -73,17 +73,16 @@ export const TranslationForm = ({
...
-            resource: 'i18n',
-            type: 'create',
-            data: fieldsToTranslate.map(camelCaseToUnderscores),
-        },
-        {
-            onComplete: (res) => setFieldsTranslations(res),
-            onError: (error) => showError(error),
-        }
-    )
+    const i18nMutationRef = useRef({
```
jenniferarnesen, line 76: Is it possible to put these mutation objects (i18nMutation and translationsMutation on line 87) outside of the component rather than using useRef? I'm not against useRef at all, just wondering.  Also, looks like `camelcaseToUnderscores` does not depend on the component state, so it could (and probably should) be moved outside of the component.

jenniferarnesen (approved):

src/components/TranslationDialog/TranslationModal/TranslationForm.js
edoardo: I think the problem is that both mutation objects need access to props, so without `useRef` I think the warning would show up.

PR #1175, opened 2022-03-16
fix: return the URL portion after origin as default TECH-1036

edoardo:
Implements [TECH-1036](https://dhis2.atlassian.net/browse/TECH-1036)

---

### Key features

1. fix URL for the Get link dialog in FileMenu

---

### Description

Before nothing was returned by default for the AO type not handled in the switch.
Now the URL portion after origin is returned instead, so a valid link is used in the Get link dialog.

---

### Screenshots

Before:
Screenshot 2022-03-16 at 10 21 01

After:
Screenshot 2022-03-16 at 10 20 08


jenniferarnesen (approved):

jenniferarnesen:
Would be useful with a couple of unit tests on the function `appPathFor` to demonstrate the output. (Not tests on the GetLinkDialog -those seems faulty)

PR #1190, opened 2022-03-30
fix: highcharts - series key and trendline broken when exporting svg (DHIS2-12719)

martinkrulltott:
Implements [DHIS2-12719](https://jira.dhis2.org/browse/DHIS2-12719)

**Relates to https://github.com/dhis2/data-visualizer-app/pull/XXX**

---

### Key features

1. sets the `allowHTML` for the svg export to fix the issue with downloading charts in DV
2. converts trendline from `span` to `svg` so it can be downloaded

---

### Description

There's an issue with downloading charts (see JIRA ticket), spoke to Highcharts support, the solution is to set `allowHTML` to true.
Also noticed that trendlines can't be printed (in dashboards) so it was converted to an `svg` like the rest of the legend.


jenniferarnesen (approved):

martinkrulltott:
Don't merge, as using `allowHTML: true` makes highcharts wrap the custom legend code in a `foreignObject` tag, which isn't supported by the backend (Batik), as seen here: https://jira.dhis2.org/browse/DHIS2-13005.
An alternative solution needs to be provided, currently the WIP can be found in the `fix/highcharts-prototype-wip` branch.

martinkrulltott:
Superseeded by https://github.com/dhis2/analytics/pull/1279

PR #1191, opened 2022-03-30
fix: get stage id and filter string for filters just like for columns and rows

jenniferarnesen:
Implements [TECH-1063](https://dhis2.atlassian.net/browse/TECH-1063)

### Key features

1. The filter values and stage id were missing from the filters property of the visualization analytics request

### Screenshots

image

https://user-images.githubusercontent.com/6113918/160862729-4033cae7-3566-45e1-b8a3-0be5a927f74b.mov

### DIFF
--- a/src/api/analytics/AnalyticsRequest.js
+++ b/src/api/analytics/AnalyticsRequest.js
@@ -84,16 +84,25 @@ class AnalyticsRequest extends AnalyticsRequestDimensionsMixin(
         const fixedIds = Object.keys(getFixedDimensions())
 
         filters.forEach((f) => {
-            request =
-                passFilterAsDimension && fixedIds.includes(f.dimension)
-                    ? request.addDimension(
-                          f.dimension,
-                          f.items?.map((item) => item.id)
-                      )
-                    : request.addFilter(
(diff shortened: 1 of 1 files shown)


src/api/analytics/AnalyticsRequest.js
```diff
@@ -84,16 +84,25 @@ class AnalyticsRequest extends AnalyticsRequestDimensionsMixin(
...
+                    ? `${f.programStage.id}.${f.dimension}`
+                    : f.dimension
+
+                if (f.filter) {
+                    filterString += `:${f.filter}`
+                }
+
+                request = request.addFilter(
+                    filterString,
+                    f.items?.map((item) => item.id)
```
jenniferarnesen, line 103: Question - I did not find any examples of the items array being populated. Is it in use for this particular case?

martinkrulltott: Items are only for "regular dimensions" (i.e. that don't have conditions), like org unit and periods. The ones that use conditions (most) won't have items.

martinkrulltott (approved):

PR #1193, opened 2022-03-31
fix: avoid dialog jumping when searching/paging TECH-1064

edoardo:
Implements [TECH-1064](https://dhis2.atlassian.net/browse/TECH-1064)

---

### Key features

1. Use fix height for the dialog to avoid jumping when searching or paging
2. open dialog at the top
3. add spacing between the loading spinner and the text

---

### Description

There can be some jumping when the text in some rows spans on multiple lines.
This is not possible to predict accurately.

The `DataTableToolbar` component has been removed to avoid a missing top border above the `Pagination` component.
The issue is visible if the number of rows is less than the maximum per page, since there is a space between the `DataTable` and the `Pagination` components.

---

### Screenshots
Screenshot 2022-03-31 at 15 48 41
Screenshot 2022-03-31 at 15 48 21
Screenshot 2022-03-31 at 15 55 10

### DIFF
--- a/src/components/OpenFileDialog/OpenFileDialog.js
+++ b/src/components/OpenFileDialog/OpenFileDialog.js
@@ -11,7 +11,6 @@ import {
     DataTableRow,
     DataTableCell,
     DataTableColumnHeader,
-    DataTableToolbar,
     NoticeBox,
     CircularLoader,
     Button,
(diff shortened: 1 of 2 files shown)


janhenrikoverland (approved):
LGTM

PR #1212, opened 2022-04-27
fix: set repetition in filter

martinkrulltott:
Implements [TECH-1131](https://dhis2.atlassian.net/browse/TECH-1131)

1. Sets `repetition` for filters (based on the implementation for `dimensions`)

---

### Known issues
- [ ] using both `recent` and `old` in `Filters` with a `filter` (e.g. https://debug.dhis2.org/tracker_dev/api/apps/line-listing/index.html#/RDg1nZ9uz67) makes the filters act as `AND` instead of `OR`. Backend bug [DHIS2-13160](https://jira.dhis2.org/browse/DHIS2-13160)


janhenrikoverland (approved):
LG

PR #1214, opened 2022-05-04
chore: add unit tests for renderValue

jenniferarnesen:
Add tests prior to fixing [TECH-1150](https://dhis2.atlassian.net/browse/TECH-1150)

### Key features

1. Tried to add enough tests to cover as many different values as possible

---

### Known issues

-   @amcgee Is it expected that the multiple-space compacting only acts on the first occurence? (I wasn't sure what specific scenario this regex was for)


amcgee:
> @amcgee Is it expected that the multiple-space compacting only acts on the first occurence? (I wasn't sure what specific scenario this regex was for)

@jenniferarnesen I think that's a bug, it should replaceAll

jenniferarnesen:
> > @amcgee Is it expected that the multiple-space compacting only acts on the first occurence? (I wasn't sure what specific scenario this regex was for)
> 
> @jenniferarnesen I think that's a bug, it should replaceAll

@amcgee Code fixed using /g (String.prototype.replaceAll function is undefined in node 14).  Could you review the latest commit and approve if you think the tests look correct?

janhenrikoverland (approved):

PR #1216, opened 2022-05-04
feat: export constants for all value types

jenniferarnesen:
Preparation for [TECH-1150](https://dhis2.atlassian.net/browse/TECH-1150)

---

### Known issues

-  @edoardo Do you know anything about 'STRING' value type? I don't find it in the backend (ref: [code line](https://github.com/dhis2/analytics/pull/1216/files#diff-22a4529af6a7511a2c24a427d67756755856bab124a015e1cd4d1cdc67822f76R33)): [backend valueType definitions](https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/ValueType.java)


src/modules/dataTypes.js
```diff
@@ -1,15 +1,26 @@
...
+    VALUE_TYPE_INTEGER,
+    VALUE_TYPE_INTEGER_POSITIVE,
+    VALUE_TYPE_INTEGER_NEGATIVE,
+    VALUE_TYPE_INTEGER_ZERO_OR_POSITIVE,
+    VALUE_TYPE_NUMBER,
+    VALUE_TYPE_UNIT_INTERVAL,
+    VALUE_TYPE_PERCENTAGE,
+} from './valueTypes.js'
 
 export const CHART_AGGREGATE_AGGREGATABLE_TYPES = [
```
jenniferarnesen: @janhenrikoverland  This isn't used anywhere in analytics and is not exported. Should it be removed?

martinkrulltott: It used to be used here https://github.com/dhis2/analytics/pull/787/files?show-deleted-files=true&show-viewed-files=true&file-filters%5B%5D=#diff-84194648b47c3b8ebd22f1db71fb0e6465aeb596863b7b51635f88b628823c80L564, but since https://github.com/dhis2/analytics/pull/787 it's just been commented out. So looks safe to remove IMO

martinkrulltott: image
attaching a screenshot if the link doesn't work

martinkrulltott (approved):

edoardo:
> * @edoardo Do you know anything about 'STRING' value type? I don't find it in the backend (ref: [code line](https://github.com/dhis2/analytics/pull/1216/files#diff-22a4529af6a7511a2c24a427d67756755856bab124a015e1cd4d1cdc67822f76R33)): [backend valueType definitions](https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/ValueType.java)

The code was initially ported from `d2.analysis` and the old apps, so it can be old stuff that has changed in the meantime (or it was never needed).

PR #1325, opened 2022-08-23
fix: change default value for Save as (DHIS2-13328)

martinkrulltott:
Implements [DHIS2-13328](https://jira.dhis2.org/browse/DHIS2-13328)

**Relates to https://github.com/dhis2/data-visualizer-app/pull/2152**

---

### Key features

1. Changes the default value to the Save as modal to append "(copy)" to the end of the original name

---

### Description

Clicking "Save as" for an AO, e.g. "My AO", will now prefill the modal with the name "My AO (copy)"

---

### TODO

- Should look for "(copy)" in the name already to prevent a name like "My AO (copy) (copy)"? Or is this actually considered a feature?

---

### Screenshots

_example of an AO called "My AO"_
image

_saving an AO again appends "(copy)" to the end of the name, resulting in "My AO (copy)"_
image

_saving an AO for the first time leaves the input empty_
image

### DIFF
--- a/src/components/FileMenu/SaveAsDialog.js
+++ b/src/components/FileMenu/SaveAsDialog.js
@@ -16,7 +16,13 @@ import { supportedFileTypes, labelForFileType } from './utils.js'
 const NAME_MAXLENGTH = 230
 
 export const SaveAsDialog = ({ type, object, onClose, onSaveAs }) => {
-    const [name, setName] = useState(object?.name)
+    const [name, setName] = useState(
+        object?.name
+            ? i18n.t('{{objectName}} (copy)', {
+                  objectName: object.name,
+              })
+            : ''
+    )
     const [description, setDescription] = useState(object?.description)
 
     // the actual API request is done in the app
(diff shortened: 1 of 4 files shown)


src/components/FileMenu/SaveAsDialog.js
```diff
@@ -16,7 +16,13 @@ import { supportedFileTypes, labelForFileType } from './utils.js'
 const NAME_MAXLENGTH = 230
 
 export const SaveAsDialog = ({ type, object, onClose, onSaveAs }) => {
-    const [name, setName] = useState(object?.name)
+    const [name, setName] = useState(
+        object?.name
```
edoardo: We should probably use `displayName` if available.
It should always be, but we must take into account the case of apps renaming `displayName` to `name` when fetching the visualization.

edoardo (approved):

edoardo (approved):

src/components/FileMenu/SaveAsDialog.js
edoardo: `displayName` with fallback to `name` 👍

PR #1332, opened 2022-08-30
fix: enable onShare callback (DHIS2-13667)

edoardo:
Implements [DHIS2-13667](https://dhis2.atlassian.net/browse/DHIS2-13667)

**Relates to https://github.com/dhis2/line-listing-app/pull/201**

---

### Key features

1. enable `onShare` callback in `FileMenu`
2. implement a mechanism for triggering a refresh in `AboutAOUnit` from the consumer app

---

### Description

The sharing dialog component supports a callback for when the sharing settings request is successful.
This however was never propagated via the `FileMenu` to the consumer app.

The `AboutAOUnit` component has been modified to provide a way of triggering a refresh of its content from the consumer app.
The mechanism is the same already used in the `InterpretationsUnit` component, which also can be refreshed whenever there is an action in the `InterpretationModal`.

### DIFF
--- a/src/components/AboutAOUnit/AboutAOUnit.js
+++ b/src/components/AboutAOUnit/AboutAOUnit.js
@@ -16,7 +16,13 @@ import {
 import cx from 'classnames'
 import moment from 'moment'
 import PropTypes from 'prop-types'
-import React, { useEffect, useMemo, useState } from 'react'
+import React, {
+    useEffect,
+    useMemo,
+    useState,
+    forwardRef,
+    useImperativeHandle,
+} from 'react'
 import { formatList } from '../../modules/list.js'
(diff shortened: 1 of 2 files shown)


HendrikThePendric (approved):
This is a valid way to get one component to talk to another one. So I approve.

For the future it might be good to have some sort of provider in place that would enable us to abstract away some of this. Currently we are leaking a lot of implementation detail to the consumers of these analytics components. I can’t really remember exactly what we were doing here, but depending on the situation we might want to do one of the following:
1. If we just want to make sure that both components are using the same data, then probably we want to store the data in the provider and read and update it.
2. If for whatever reason that’s not good enough, then we can keep the current implementation with the refs and imperative handle, but then expose the refs and the refresh methods from the provider.

PR #1352, opened 2022-10-04
fix: use styled-jsx instead of css module

martinkrulltott:
https://github.com/dhis2/analytics/pull/1342 incorrectly used a css module instead of styled-jsx


src/components/OfflineTooltip.js
```diff
@@ -17,25 +17,25 @@ const OfflineTooltip = ({
...
-        </Tooltip>
+        <>
+            <Tooltip
+                content={content || i18n.t('Not available offline')}
+                openDelay={200}
+                closeDelay={100}
+            >
+                {({ onMouseOver, onMouseOut, ref }) => (
+                    <span
+                        className={cx('tooltip', notAllowed && 'notAllowed')}
```
jenniferarnesen: I've been informed by platform team that the preferred way to do this is: 
```
 <div
          className={cx(classes.iconWrapper, {
              [classes.activeIcon]: isColumns(cols),
          })}
      >
```
https://github.com/dhis2/dashboard-app/blob/master/src/pages/edit/LayoutModal.js#L90

jenniferarnesen: Or maybe even this from ui:
https://github.com/dhis2/ui/blob/master/components/tag/src/tag.js#L22

martinkrulltott: I used the `css` syntax as it supports things like `.tooltip > :global(button:disabled)`. Is it possible to do this with the syntax you're proposing?

jenniferarnesen: I tested it out and it works: `className={cx('tooltip', {notAllowed})}`

jenniferarnesen (approved):

PR #1356, opened 2022-10-10
test: add some additional tests for renderValue

jenniferarnesen:
renderValue should not add a digitGroupSeparator to the value if the value is not a number type. Phone numbers and dates are not number types.

I also removed a TODO comment since it didn't seem useful. The decision on whether to add support for a Period dgs should come from product manager rather than a code comment.


src/modules/__tests__/renderValue.spec.js
```diff
@@ -251,6 +254,57 @@ const tests = [
...
+    {
+        value: '303-8485-9922',
+        expected: '303-8485-9922',
+        valueType: VALUE_TYPE_PHONE_NUMBER,
+        round: true,
+        dgs: DGS_COMMA,
+    },
+    {
+        value: '303  848  9922',
+        expected: '303 848  9922',
```
martinkrulltott, line 281: why are the amount of spaces between 303 and 848 different in `value` compared to `expected`? (but the same between 848 and 9922)...

martinkrulltott (approved):

src/modules/__tests__/renderValue.spec.js
jenniferarnesen: This happens because of the regex here: https://github.com/dhis2/analytics/blob/master/src/modules/renderValue.js#L62
I discussed this previously with Austin who wrote this, but nothing really came out of it, and since no bugs have been reported, I've just left it as is (I'm very hesitant to go changing regexes).

So this test was added specifically to illustrate the regex behavior. Shall I add a comment?

PR #1370, opened 2022-11-04
feat: custom calculations (DHIS2-13871)

martinkrulltott:
Implements [DHIS2-13871](https://dhis2.atlassian.net/browse/DHIS2-13871)

[Design spec](https://www.figma.com/proto/idNlPeSFYHtGy1TqMSWG9M/Data-Visualizer?page-id=101%3A2&node-id=101%3A5&viewport=410%2C470%2C0.5&scaling=contain&starting-point-node-id=101%3A5)

**Relates to https://github.com/dhis2/data-visualizer-app/pull/2232**

---

### Key features

1. Adds support for listing, creating and editing calculations in the Data dimension modal
2. Fetches data elements that can be searched and filtered - @edoardo 🙏 
3. Supports drag-n-drop from the dimension list and within the formula field - @jenniferarnesen 🙏 
4. Avoids stale cached results from the backend by generating and using a new `edi_cache` prop - @janhenrikoverland 🙏 

---

### New lingo

_EDI_: `Expression Dimension Item`, commonly referred to as a "calculation". 
_formula_: The actual math formula for the EDI, known in the backend as `expression`.

---

### Description

The `DataDimension` component now supports listing, creating and editing EDIs. This consists of several parts:

**The Data Dimension modal**
1. Adds a new `+ Calculation` button at the bottom of the modal for creating new EDIs (opens the new calculations modal), 
5. The dimension list includes EDIs and can filter to view only EDIs by using the new `Calculations` option in the data type filter.
6. Listed EDIs (both selected and non-selected) have an edit button that opens the new calculations modal

**The new Calculation modal**
1. Fetches and lists available data elements and data element operands for selection, which can also be searched and filtered, to be used in a formula (curtesy of @edoardo)
2. Supports drag-and-drop from the dimension list into the formula field and within the formula field itself (curtesy of @jenniferarnesen)
3. The formula supports client-side validation for some basic mistakes with user-friendly error messages, which is followed by a more rigorous backend validation (that however unfortunately doesn't provide useful error message)

**Metadata strategy**

All metadata needed for the separate parts of the formula in an EDI are handled by the `CalculationModal` internally, i.e. no separate metadata needs to be passed in when the `CalculationModal` is opened. This alleviates the consuming apps (namely DV) as they won't need to fetch and store this metadata (which was initially the idea, but turned out to be quite messy).

**Parsing a formula / `expression`**

As a formula is created as separate parts in the `CalculationsModal` (as an array) and the backend format for `expression` is a simple string, there are two new helper functions for handling this: `parseExpressionToArray` and `parseArrayToExpression`. The former uses a regex (curtesy of ChatGPT!) for splitting the `expression` into different parts and maps the parts with provided `metadata` (which is fetched internally, as per the previous section above).

**Cache handling / `edi_cache`**

The backend caches requests based on the provided params and thus disregards if an EDI has been updated. To avoid returning a stale cached response after updating an EDI a special `edi_cache` param are sent with each request that changes with the content of the EDI's formula. This is created by sorting a visualisation's expressions, joining them and hashing it all using `crypto-js/sha1` (curtesy of @janhenrikoverland)

---

### BREAKING CHANGE

The data dimension modal now return `Expression Dimension Items`, which require the consuming app to provide the correct metadata for these items.

---

### TODO

-   [x] remove formula validation when the formula is dirty
-   [x] refactor ’apiValidateExpression’
-   [x] error handling for all requests, show the alert bar + wrap everything after `await` to only continue if the response is ok
_-   [ ] Handle deleting an EDI that is used in an AO_ - delete is handled by the backed and should always be allowed 
-   [x] New design
-   [x] Fetch data elements
-   [x] Search / filters
-   [x] Infinite scrolling
-   [x] Parse string -> components (opening) and components -> string (saving)
-   [x] Fetch metadata for the data elements used within the `expression`
-   [x] Display names for dimensions in the calculations in the formula field
-   [x] Double-click to add
-   [x] Drag-n-drop
-   [x]  `<number>` - dnd with "edit mode"

---

### Known issues

Numbered list for internal use. _italic_ = fixed

1. _The `admin` user can't save EDIs (missing permissions, temporarily using user `system` now instead)_ - enable `F_EXPRESSION_DIMENSION_ITEM` in the [user app](https://debug.dhis2.org/dev/dhis-web-user/index.html#/user-roles/edit/Ufph3mGRmMo)
7. _Requests after the `expression` of an EDI is changed will return values based on the old/outdated `expression`_ - fixed with `edi_cache`
8. _EDIs are missing `dimensionItemType` when opening an AO (temporarily hardcoded for now instead)_ - fixed by backend
9. _EDIs are missing `name` when the `shortName` setting is used_ - fixed by backend
10. _Data elements used in an `expression` is missing metadata (i.e. no way to map an `id` to a `name`)_ - metadata has been added
11. _`expression` value missing from metadata when opening an AO_ - metadata has been added
12. _Trying to delete an EDI that is used by a visualization throws an error_ - delete is now allowed
13. _Opening a saved AO with an EDI and changing the name of the EDI results in the old name still being displayed in the `Transfer` list (due to outdated metadata in the consuming app's store)_
14. Dragging a data element downwards also triggers scrolling in the data element list. A similar problem was fixed in LL.

---

### Screenshots

_New `Calculations` data type and new `+ Calculation` button_

image

_New calculations modal that opens on top of the data dimension modal_

image

_Creating a formula with drag-n-drop and/or click/double-click_

https://user-images.githubusercontent.com/12590483/226877653-df7c3b4e-0a36-4224-bd9d-ea88e55bce50.mov

_Client-side validation_

image

_Searching and filtering_

image


src/api/expression.js
```diff
@@ -0,0 +1,54 @@
...
+    })
+
+    return validationData
+}
+
+export const apiSaveExpressionDimensionItem = (
+    dataEngine,
+    expressionDimensionItem
+) => {
+    const mutation = {
```
edoardo: I think this definition should be moved out of the function, and have 2 mutation for POST/PUT like the other ones.

src/api/expression.js
```diff
@@ -0,0 +1,54 @@
...
+    data: ({ name, expression }) => ({ name, expression }),
+}
+
+export const deleteCalculationMutation = {
+    type: 'delete',
+    resource: 'expressionDimensionItems',
+    id: ({ id }) => id,
+}
+
+export const apiValidateExpression = async (dataEngine, expression) => {
```
edoardo: Is this function used in different places?

src/components/DataDimension/ItemSelector.js
```diff
@@ -376,88 +398,167 @@ const ItemSelector = ({
...
+            case DIMENSION_TYPE_EXPRESSION_DIMENSION_ITEM:
+                return CalculationIcon
             default:
                 return GenericIcon
         }
     }
+
+    const [createCalculation] = useDataMutation(createCalculationMutation)
+    const [updateCalculation] = useDataMutation(updateCalculationMutation)
+    const [deleteCalculation] = useDataMutation(deleteCalculationMutation)
```
edoardo: What happens in case any of these mutations fails?
I don't see any error handling, so the code after the mutation below is still executed, is that correct?

src/api/expression.js
martinkrulltott: That was the legacy function that I forgot to remove, I removed it now 👍

src/api/expression.js
martinkrulltott: no, moved it to the component instead

src/components/DataDimension/DataDimension.js
```diff
@@ -26,6 +27,7 @@ const DataDimension = ({
                 label: item.name,
                 isActive: item.isActive,
                 type: item.type,
+                expression: item.expression,
```
martinkrulltott, line 38: This requires the consuming app to provide `expression` for `selectedDimensions` items, e.g. `expression` needs to be added to `ITEMS` [here in Data Vis](https://github.com/dhis2/data-visualizer-app/blob/c41c02fc81b7351d871a760597e9a679c6a02e06/src/modules/fields/nestedFields.js#L10)

src/assets/FormulaIcon.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react'
+
+const FormulaIcon = () => {
```
jenniferarnesen, line 3: very minor - should we do `export default (<svg...` like in the CalculationIcon file, for consistency?

src/components/DataDimension/DataElementSelector.js
```diff
@@ -0,0 +1,321 @@
...
+                        ? [...prevOptions, ...newOptions]
+                        : newOptions
+                )
+
+                setLoading(false)
+            }
+
+            hasNextPageRef.current = result?.nextPage ? true : false
+        } catch (error) {
+            // TODO handle errors
```
jenniferarnesen: @martinkrulltott  should a subtask be added for handling errors?

src/modules/hash.js
```diff
@@ -0,0 +1,22 @@
+import SHA1 from 'crypto-js/sha1'
+import { DIMENSION_TYPE_EXPRESSION_DIMENSION_ITEM } from './dataTypes.js'
+import { layoutGetAllItems } from './layout/layoutGetAllItems.js'
+
+const isValidValue = (value) => typeof value === 'string' && value.length
+
+export const getHash = (value) =>
```
jenniferarnesen, line 7: Is this exported for the sake of jest tests? Personally I would prefer moving those test scenarios to the tests for `getExpressionHashFromVisualization` and not export the getHash function, to keep things cleaner.

src/components/DataDimension/Calculation/CalculationModal.js
```diff
@@ -0,0 +1,417 @@
...
+            }),
+        },
+    }
+
+    const { data, refetch } = useDataQuery(query, {
+        lazy: true,
+    })
+
+    useEffect(() => {
+        const expressionIds = calculation.expression
```
jenniferarnesen: Suggesting moving this regex to modules/expressions to keep regex's related to expressions in one place, and to make this testable. For instance add a new function `getExpressionDimensionIds` or similar.

src/modules/dataTypes.js
```diff
@@ -18,6 +18,8 @@ export const DIMENSION_TYPE_PERIOD = 'PERIOD'
 export const DIMENSION_TYPE_ORGANISATION_UNIT = 'ORGANISATION_UNIT'
 export const DIMENSION_TYPE_ORGANISATION_UNIT_GROUP_SET =
     'ORGANISATION_UNIT_GROUP_SET'
+export const DIMENSION_TYPE_EXPRESSION_DIMENSION_ITEM =
```
jenniferarnesen, line 21: Should this be exported from `src/index/js` like all the others?

src/components/DataDimension/Calculation/CalculationModal.js
```diff
@@ -0,0 +1,417 @@
...
+                                    onDoubleClick={removeItem}
+                                    loading={!expressionArray}
+                                />
+                                <div className="actions-wrapper">
+                                    <Button
+                                        small
+                                        onClick={validate}
+                                        dataTest="validate-button"
+                                    >
+                                        {/* TODO: add loading state to button? */}
```
martinkrulltott: Add a subtask for this?

src/components/DataDimension/Calculation/CalculationModal.js
```diff
@@ -0,0 +1,417 @@
...
+                                            onChange={({ value }) =>
+                                                setName(value.substr(0, 50))
+                                            }
+                                            value={name}
+                                            dataTest="calculation-label"
+                                            dense
+                                        />
+                                    </div>
+                                    <p style={{ color: 'blue' }}>
+                                        {/* TODO: Remove, for testing only */}
```
martinkrulltott: Should be removed before merging

src/components/DataDimension/Calculation/DataElementSelector.js
```diff
@@ -0,0 +1,321 @@
...
+                        ? [...prevOptions, ...newOptions]
+                        : newOptions
+                )
+
+                setLoading(false)
+            }
+
+            hasNextPageRef.current = result?.nextPage ? true : false
+        } catch (error) {
+            // TODO handle errors
```
martinkrulltott, line 175: Add a subtask for this?

src/modules/expressions.js
```diff
@@ -0,0 +1,114 @@
...
+        }) || []
+    )
+}
+
+export const parseArrayToExpression = (input = []) =>
+    input.map((item) => item.value).join('')
+
+export const validateExpression = (expression) => {
+    let result
+    // TODO: two numbers next to each other
```
martinkrulltott: I don't think we can handle this easily, as the `expression` here is already parsed to a string.

src/assets/FormulaIcon.js
martinkrulltott: Im not sure, `AssignedCategoriesIcon` and `DynamicDimensionIcon` in the same folder both have this format as well, so I can't tell which format is preferred (consistency would be nice though!)

src/components/DataDimension/DataElementSelector.js
martinkrulltott: @edoardo wrote this part and the comment. Thoughts?

martinkrulltott: Subtask: https://dhis2.atlassian.net/browse/DHIS2-15034

src/modules/hash.js
martinkrulltott: @janhenrikoverland

src/modules/dataTypes.js
martinkrulltott: Yes it can be exported. I guess it's not already since it's not used by any of the apps. But I've added the export now 👍

src/modules/hash.js
janhenrikoverland: @jenniferarnesen  We might want to swap out the SHA1 for something else, so I think it's good to have a couple of tests for the hash function directly. If the new function returns something unexpected, the problem isn't hidden somewhere in our dimension structure.

src/components/DataDimension/Calculation/CalculationModal.js
martinkrulltott: done 👍

src/components/DataDimension/Calculation/DataElementSelector.js
martinkrulltott: https://dhis2.atlassian.net/browse/DHIS2-15034

src/components/DataDimension/Calculation/CalculationModal.js
martinkrulltott: https://dhis2.atlassian.net/browse/DHIS2-15046

edoardo (approved):

edoardo (approved):

jenniferarnesen (approved):

janhenrikoverland (approved):

PR #1380, opened 2022-11-18
chore: fix i18n imports and usage

edoardo:
### Key features

1. use the same import for `i18n` everywhere
2. ensure `i18n.t` gets executed for all translatable strings

---

### Description

We had 2 ways of import `i18n` in the library.
I don't think we had problems due to this, but for consistency sake and to avoid possible issues, it's good to import everywhere from the same place.
Importing from `locales/index.js` ensures that the library translations are added and are available in the consumer apps.

Some strings did not show translated in consumer apps, the reason is likely marking `i18n.t` strings in static constants or objects, in that case the `i18n.t` function is not called at runtime.


src/components/OpenFileDialog/utils.js
```diff
@@ -18,65 +18,66 @@ export const AOTypeMap = {
...
+export const getTranslatedString = (type, key) => {
+    const texts = {
+        [NO_TYPE]: {
+            modalTitle: i18n.t('Open'),
+            loadingText: i18n.t('Loading'),
+            errorTitle: i18n.t("Couldn't load items"),
+            errorText: i18n.t(
+                'There was a problem loading items. Try again or contact your system administrator.'
+            ),
+            noDataText: i18n.t('No items found. Create a new to get started.'),
```
jenniferarnesen: Kind of a weird sentence (I know this was already present):  "Create a new to get started."

jenniferarnesen (dismissed):

HendrikThePendric (commented):
I wonder if there is perhaps a better solution to this problem. The current solution has 2 downsides as I see it:
1. importing from the file `locales/index.js` is less ergonomic than importing from the package `@dhis2/d2-i18n`
2. The file `locales/index.js` actually does do quite a lot. If we import this file in a lot of places the calls to `.addResources()` will be happening often. I don't know if this is a problem, but doesn't seem desirable

I'm thinking the following:
1. To ensure we always "load the translations" for consumers of the lib, we can import the `locales/index.js` file from our own index file, since consuming apps should be importing from there exclusively (if this is not true, I think the current implementation in this PR is probably the best solution)
2. To avoid running the code in `locales/index.js` multiple times, we probably would have to do a dynamic import, i.e. something like this:
    1. Import the i18n singleton from `@dhis2/d2-i18n`
    2. check if our translations have been added (not sure how, but I think it may be possible)
     3. If not dynamically import the `locales/index.js` file
 3. We could consider checking how this is done in @dhis2/ui

PR #1401, opened 2022-12-20
fix: do not pass translation object when value is empty

edoardo:
Fixes [DHIS2-14347](https://dhis2.atlassian.net/browse/DHIS2-14347)

---

### Key features

1. fix payload to avoid 409 conflict when saving an empty translation

---

### Description

This causes a 409 response from the translations API. If the input field is cleared, assuming the user wants to remove a translation for a language/string entirely, do not pass the object with an empty value, but remove it entirely from the list of translations. We still need to pass the full list of translations because the API does not support PATCH requests so the logic is a bit more complex.

### Screencast

Screencast demonstrating the fix, no errors are thrown when saving empty translations:

https://user-images.githubusercontent.com/150978/208868968-9d34409e-0945-44bd-ac17-36b89b9c4377.mov

### DIFF
--- a/src/components/TranslationDialog/TranslationModal/TranslationForm.js
+++ b/src/components/TranslationDialog/TranslationModal/TranslationForm.js
@@ -64,13 +64,33 @@ export const TranslationForm = ({
 
         const translationIndex = getTranslationIndexForField(field)
 
-        setNewTranslations(
-            translationIndex === -1
-                ? [...newTranslations, newTranslation]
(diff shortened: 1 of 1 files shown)


martinkrulltott (approved):
LGMT. Some sort of testing (Jest here or Cypress in an app) would be great!

edoardo:
> LGMT. Some sort of testing (Jest here or Cypress in an app) would be great!

I'll write some Cypress tests in LL.

PR #1407, opened 2023-01-02
feat: use multi-calendar-dates library to generate fixed periods

kabaros:
This PR updates the logic that calculates the fixed periods to use the [multi-calendar-dates](https://github.com/dhis2/multi-calendar-dates/pull/7) library to generate fixed periods in calendrical systems other than Gregorian (i.e. Ethiopic and Nepali)

Implements [DHIS2-14226](https://dhis2.atlassian.net/browse/DHIS2-14226)

**Relates to  [Support period generation in multi-calendar-dates library](https://github.com/dhis2/multi-calendar-dates/pull/7)** - this PR should not be merged at least until that PR is approved and merged.

### Why?

The multi-calendar-dates library aims to encapsulate and abstract all date related operations for use across DHIS2 frontend stack. It relies on the [Temporal proposal](https://tc39.es/proposal-temporal) which solves many problems with the native JS Date. In our context, the proposal supports calendars other than the Gregorian calendar making it possible to support Nepali and Ethiopic calendars for example. It also provides ways to do arithmetic operations on dates (adding and subtracting periods) in a DST-safe timezone aware manner.  This ability is what powers the period generation here.

The [calendar component](https://pr-1162--dhis2-ui.netlify.app/components/calendar/) was already implemented (beta) in the UI library, and the next requirement was the one addressed here to allow period generation in different calendars.
 
For more context on the multi-calendar project. You can refer to:

- [The project specs](https://docs.google.com/document/d/19zjyB45oBbqC5KeubaU8E7cw9fGhFc3tOXY0GkzZKqc/edit#)
- [ADR on why we used Temporal](https://github.com/dhis2/multi-calendar-dates/blob/beta/doc/architecture/decisions/0002-use-temporal-api-as-the-backbone-for-the-engine.md)
- proj-multi-calendar channel on Slack

---

### Key features

1. Generate fixed periods in multiple calendrical systems (i.e. Gregorian, Nepali, Ethiopic etc...)
2. Get the current date in multiple calendars
3. Consume and use the currently configured calendar in DHIS2

---

### BREAKING CHANGE

No breaking change. The existing period generation logic in analytics came with an exhaustive suite of tests. The library used those as a specs to guide the implementation, and make sure they all pass, so the library should not change the existing behaviour, only extend it to support other calendars ✅ 

---

### TODO

-   [x] The [multi-calendar-dates PR](https://github.com/dhis2/multi-calendar-dates/pull/7) is still in review. Once merged and published to beta and main channels, then the dependency here should be updated. **There is a [task](https://dhis2.atlassian.net/browse/DHIS2-14226) to update these dependencies once we have a final version**
-  [x]  Test with DHIS2 backends that actually implement analytics in other calendrical systems (i.e. The Ethiopia PR / fork) - **this can be done once a version of this library is published**
-  [x] Localisation - we rely on Intl.Locale to localise the date parts. This works for almost all locales (except Nepali which is not part of the unicode locale repository). I need to do more testing and investigation to know what values we exactly pass, and whether that could lead to unexpected behaviour - there is also a related task in the backlog to make the library more resilient to locale requests, falling back to english if a passed locale is not supported.
- [x] We initially thought we will build a set of hooks to be consumed from the multi-calendar-dates library. I chose not to, as I wanted the changes to analytics to be as minimal as possible, but we can discuss this further to see if it's necessary.

---

### Known issues

None so far. I used the tests from analytics to mimic the previous behaviour exactly, and the tests seem quite extensive. But it's a major change, and should go through much more testing.

---

### Screenshots

https://user-images.githubusercontent.com/1014725/210541982-38e59822-0b9b-4508-9041-bd4b0747661a.mov

[DHIS2-14226]: https://dhis2.atlassian.net/browse/DHIS2-14226?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DHIS2-14226]: https://dhis2.atlassian.net/browse/DHIS2-14226?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,6 +48,11 @@
         "styled-jsx": "^4.0.1",
         "typeface-roboto": "^0.0.75"
     },
+    "jest": {
+        "transformIgnorePatterns": [
+            "/!node_modules\\/multi-calendar-dates/"
+        ]
+    },
     "peerDependencies": {
         "@dhis2/app-runtime": "^3",
         "@dhis2/d2-i18n": "^1.1",
@@ -59,6 +64,7 @@
     },
     "dependencies": {
         "@dhis2/d2-ui-rich-text": "^7.4.0",
+        "@dhis2/multi-calendar-dates": "1.0.0-beta.21",
         "classnames": "^2.3.1",
         "d2-utilizr": "^0.2.16",
         "d3-color": "^1.2.3",
--- a/src/components/PeriodDimension/PeriodTransfer.js
+++ b/src/components/PeriodDimension/PeriodTransfer.js
@@ -25,7 +25,7 @@ const PeriodTransfer = ({
     rightFooter,
     excludedPeriodTypes,
 }) => {
-    const { systemInfo = {} } = useConfig()
+    const { systemInfo } = useConfig()
     const { calendar = 'gregory' } = systemInfo
 
     const defaultRelativePeriodType = excludedPeriodTypes.includes(MONTHLY)
--- a/src/components/PeriodDimension/PeriodTransfer.js
+++ b/src/components/PeriodDimension/PeriodTransfer.js
@@ -35,9 +35,10 @@ const PeriodTransfer = ({
         ? getFixedPeriodsOptionsById(QUARTERLY, calendar)
         : getFixedPeriodsOptionsById(MONTHLY, calendar)
 
+    const now = getNowInCalendar(calendar)
     // use ".eraYear" rather than ".year" because in Ethiopian calendar, eraYear is what our users expect to see (for other calendars, it doesn't matter)
     // there is still a pending decision in Temporal regarding which era to use by default: https://github.com/js-temporal/temporal-polyfill/blob/9350ee7dd0d29f329fc097debf923a517c32f813/lib/calendar.ts#L1964
-    const defaultFixedPeriodYear = getNowInCalendar(calendar).eraYear
+    const defaultFixedPeriodYear = now.eraYear || now.year
 
     const fixedPeriodConfig = (year) => ({
         offset: year - defaultFixedPeriodYear,
--- a/src/components/PeriodDimension/PeriodTransfer.js
+++ b/src/components/PeriodDimension/PeriodTransfer.js
@@ -1,3 +1,5 @@
+import { useConfig } from '@dhis2/app-runtime'
+import { getNowInCalendar } from '@dhis2/multi-calendar-dates'
 import { TabBar, Tab, Transfer } from '@dhis2/ui'
 import PropTypes from 'prop-types'
 import React, { useState } from 'react'
@@ -23,13 +25,20 @@ const PeriodTransfer = ({
     rightFooter,
     excludedPeriodTypes,
 }) => {
+    const { systemInfo = {} } = useConfig()
+    const { calendar = 'gregory' } = systemInfo
+
     const defaultRelativePeriodType = excludedPeriodTypes.includes(MONTHLY)
         ? getRelativePeriodsOptionsById(QUARTERLY)
         : getRelativePeriodsOptionsById(MONTHLY)
     const defaultFixedPeriodType = excludedPeriodTypes.includes(MONTHLY)
-        ? getFixedPeriodsOptionsById(QUARTERLY)
-        : getFixedPeriodsOptionsById(MONTHLY)
-    const defaultFixedPeriodYear = new Date().getFullYear()
+        ? getFixedPeriodsOptionsById(QUARTERLY, calendar)
+        : getFixedPeriodsOptionsById(MONTHLY, calendar)
+
+    // use ".eraYear" rather than ".year" because in Ethiopian calendar, eraYear is what our users expect to see (for other calendars, it doesn't matter)
+    // there is still a pending decision in Temporal regarding which era to use by default: https://github.com/js-temporal/temporal-polyfill/blob/9350ee7dd0d29f329fc097debf923a517c32f813/lib/calendar.ts#L1964
+    const defaultFixedPeriodYear = getNowInCalendar(calendar).eraYear
+
     const fixedPeriodConfig = (year) => ({
         offset: year - defaultFixedPeriodYear,
         filterFuturePeriods: false,
@@ -60,7 +69,8 @@ const PeriodTransfer = ({
                           relativeFilter.periodType
                       ).getPeriods()
                     : getFixedPeriodsOptionsById(
-                          fixedFilter.periodType
+                          fixedFilter.periodType,
+                          calendar
                       ).getPeriods(fixedPeriodConfig(Number(fixedFilter.year)))
             )
         }
@@ -134,8 +144,9 @@ const PeriodTransfer = ({
     const onSelectFixedPeriods = (filter) => {
         setFixedFilter(filter)
         setAllPeriods(
-            getFixedPeriodsOptionsById(filter.periodType).getPeriods(
-                fixedPeriodConfig(Number(filter.year))
+            getFixedPeriodsOptionsById(filter.periodType, calendar).getPeriods(
+                fixedPeriodConfig(Number(filter.year)),
+                calendar
             )
         )
     }
--- a/src/components/PeriodDimension/utils/fixedPeriods.js
+++ b/src/components/PeriodDimension/utils/fixedPeriods.js
@@ -50,7 +50,9 @@ const getPeriods = ({ periodType, config, fnFilter, calendar }) => {
     const isReverse = periodType.match(/^FY|YEARLY/)
         ? true
         : config.reversePeriods
-    const year = getNowInCalendar(calendar).eraYear + offset
+
+    const now = getNowInCalendar(calendar)
+    const year = (now.eraYear || now.year) + offset
 
     const params = {
         periodType,
--- a/src/components/PeriodDimension/utils/fixedPeriods.js
+++ b/src/components/PeriodDimension/utils/fixedPeriods.js
@@ -1,3 +1,7 @@
+import {
+    generateFixedPeriods,
+    getNowInCalendar,
+} from '@dhis2/multi-calendar-dates'
 import i18n from '../../../locales/index.js'
 import {
     DAILY,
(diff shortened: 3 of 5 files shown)


kabaros:
@edoardo @martinkrulltott .. please have a look at this PR when you get a chance. 

Martin - Edoardo probably has more context to the change since he was involved in the discussions in Oslo and in the slack channel. I added you as a reviewer though since I saw you made the most recent changes. I tried to write down the context but please let me know if any of it is unclear.

package.json
```diff
@@ -65,6 +70,7 @@
         "highcharts": "^10.2.0",
         "lodash": "^4.17.21",
         "mathjs": "^9.4.2",
+        "multi-calendar-dates": "1.0.0-alpha.12",
```
edoardo: Minor comment: I expected this to have the `@dhis2` namespace.

src/components/PeriodDimension/PeriodTransfer.js
```diff
@@ -23,13 +25,20 @@ const PeriodTransfer = ({
     rightFooter,
     excludedPeriodTypes,
 }) => {
+    const { systemInfo = {} } = useConfig()
```
edoardo: Not related to the calendar implementation itself, but do we need to have defaults for these hooks?
We never used it this way, doesn't the hook return a default already to make sure it works without the consumer having to worry about it?

src/components/PeriodDimension/utils/fixedPeriods.js
```diff
@@ -40,510 +41,182 @@ const PERIOD_TYPE_REGEX = {
...
+    const offset = parseInt(config.offset, 10)
+    const isFilter = config.filterFuturePeriods
+    const isReverse = config.reversePeriods
+    const year = getNowInCalendar(calendar).eraYear + offset
+
+    const params = {
+        periodType,
+        year,
+        calendar,
+        locale: 'en',
```
edoardo, line 61: Should this come from user settings?

src/components/PeriodDimension/utils/fixedPeriods.js
```diff
@@ -40,510 +41,182 @@ const PERIOD_TYPE_REGEX = {
...
-            date.setDate(date.getDate() + 1)
-
-            biWeek += 1
-        }
-
-        periods = isFilter ? fnFilter(periods) : periods
-        periods = isReverse ? periods.reverse() : periods
-
-        return periods
+        return getPeriods({
```
edoardo, line 86: Does this uses the same logic for the EPI week computation?

src/components/PeriodDimension/utils/fixedPeriods.js
```diff
@@ -40,510 +41,182 @@ const PERIOD_TYPE_REGEX = {
...
-        periods = isFilter ? fnFilter(periods) : periods
-        periods = isReverse ? periods : periods.reverse()
-        // FinancialOctober periods are collected backwards. If isReverse is true, then do nothing. Else reverse to correct order and return.
-
-        return periods
+        return getPeriods({
+            periodType: 'YEARLY',
+            config: {
+                ...config,
+                reversePeriods: true,
```
edoardo: Why is this overridden?

src/components/PeriodDimension/utils/fixedPeriods.js
```diff
@@ -40,510 +41,182 @@ const PERIOD_TYPE_REGEX = {
...
-        periods = isFilter ? fnFilter(periods) : periods
-        periods = isReverse ? periods : periods.reverse()
-        // FinancialNovember periods are collected backwards. If isReverse is true, then do nothing. Else reverse to correct order and return.
-
-        return periods
+        return getPeriods({
+            periodType: 'FYOCT',
+            config: {
+                ...config,
+                reversePeriods: true,
```
edoardo: Same here.

src/components/PeriodDimension/utils/fixedPeriods.js
```diff
@@ -40,510 +41,182 @@ const PERIOD_TYPE_REGEX = {
...
-        periods = isFilter ? fnFilter(periods) : periods
-        periods = isReverse ? periods : periods.reverse()
-        // FinancialJuly periods are collected backwards. If isReverse is true, then do nothing. Else reverse to correct order and return.
-
-        return periods
+        return getPeriods({
+            periodType: 'FYNOV',
+            config: {
+                ...config,
+                reversePeriods: true,
```
edoardo: And here.

src/components/PeriodDimension/utils/fixedPeriods.js
```diff
@@ -40,510 +41,182 @@ const PERIOD_TYPE_REGEX = {
...
-        periods = isFilter ? fnFilter(periods) : periods
-        periods = isReverse ? periods : periods.reverse()
-        // FinancialApril periods are collected backwards. If isReverse is true, then do nothing. Else reverse to correct order and return.
-
-        return periods
+        return getPeriods({
+            periodType: 'FYJUL',
+            config: {
+                ...config,
+                reversePeriods: true,
```
edoardo: See above.

edoardo (commented):
Do periods like `Six-monthly April` or `Financial year (Start October)` make sense in a non gregorian calendar?
What do they actually mean in a Nepali calendar context for example?

package.json
kabaros: > Minor comment: I expected this to have the @dhis2 namespace.

hmm not sure but I think you're probably right .. @HendrikThePendric @amcgee  ?

src/components/PeriodDimension/PeriodTransfer.js
kabaros: not sure - I looked at the type declared [here](https://github.com/dhis2/app-runtime/blob/master/services/config/src/types.ts#L20) and it seems that it could be undefined, so wanted to be defensive about this and not take risks.

maybe, also @amcgee would know if this returns a default? (I couldn't tell where the apps set the provider config)

src/components/PeriodDimension/utils/fixedPeriods.js
kabaros: yeah this is related to one of the ToDo comments - it should come from the user settings, but I also need to potentially do some sanitising for it, as it's passed to the .toLocaleString, and I want to double check what values we allow, and make sure an invalid value doesn't break things.

src/components/PeriodDimension/utils/fixedPeriods.js
kabaros: yes - there were basically [snapshot tests](https://github.com/dhis2/analytics/blob/master/src/components/PeriodDimension/__tests__/__snapshots__/fixedPeriods.spec.js.snap) that cover all the EPI week scenarios, and I went through them until they all passed.   Initially it was missing a lot of subtle bits (the definition of the start of the year and the last week of the year was off) but it's the same logic as before now.

You can see the implementation [here](https://github.com/dhis2/multi-calendar-dates/pull/7/files#diff-62fb85fb1903c24f42fb487f3595ac2d5a4384400c38059a238c563a86245d1d).

src/components/PeriodDimension/utils/fixedPeriods.js
kabaros: this is to mimic the previous behaviour (and make the tests pass) .. Some periods (i.e. Yearly, Financial Years) are displayed from the newest to oldest .. while all the others are displayed from oldest to newest

src/components/PeriodDimension/utils/fixedPeriods.js
kabaros: same answer above

src/components/PeriodDimension/utils/fixedPeriods.js
kabaros: same answer above

src/components/PeriodDimension/utils/fixedPeriods.js
kabaros: same answer above

kabaros:
> Do periods like Six-monthly April or Financial year (Start October) make sense in a non gregorian calendar?
What do they actually mean in a Nepali calendar context for example?

They would mean the equivalent nth month in the specific calendar. For FYNOV, November would mean the 11th month in the calendar (Hamle in Ethiopic, Falgun in Nepal). I believe @abyot give us this one as an example, as the financial year in Ethiopia goes from Hamle to Hamle, which will be FYNOV in our system, referring to the 11th month Hamle (even though this is actually July in the Gregorian calendar).

A better solution would be to redefine these to FY11, I guess, but I assume that would be a more far-reaching change for the backend.

edoardo:
> > Do periods like Six-monthly April or Financial year (Start October) make sense in a non gregorian calendar?
> > What do they actually mean in a Nepali calendar context for example?
> 
> They would mean the equivalent nth month in the specific calendar. For FYNOV, November would mean the 11th month in the calendar (Hamle in Ethiopic, Falgun in Nepal). I believe @abyot give us this one as an example, as the financial year in Ethiopia goes from Hamle to Hamle, which will be FYNOV in our system, referring to the 11th month Hamle (even though this is actually July in the Gregorian calendar).
> 
> A better solution would be to redefine these to FY11, I guess, but I assume that would be a more far-reaching change for the backend.

I was thinking more about the labels, perhaps those can be different based on the calendar (if it makes more sense)?

src/components/PeriodDimension/utils/fixedPeriods.js
edoardo: Ok, I wasn't sure because in the old code `reversePeriod` was passed in the config, and the config is used also in the new code.

kabaros:
> I was thinking more about the labels, perhaps those can be different based on the calendar (if it makes more sense)?

you mean the labels, as in the labels of the drop down list?

src/components/PeriodDimension/PeriodTransfer.js
edoardo: It's the app adaptor that defines the provider:
https://github.com/dhis2/app-platform/blob/ac0350a2562429b941833725b711d1b01c0539ad/adapter/src/components/ServerVersionProvider.js#L57-L72

edoardo:
> you mean the labels, as in the labels of the drop down list?

Yes, the Period type labels. Maybe it's a question for @cooper-joe too.

cooper-joe:
> > you mean the labels, as in the labels of the drop down list?
> 
> Yes, the Period type labels. Maybe it's a question for @cooper-joe too.

Good question. If I understand correctly, then yes, it would be ideal if the labels in the period type dropdown reflected the chosen calendar as best as possible.

abyot:
> > > you mean the labels, as in the labels of the drop down list?
> > 
> > 
> > Yes, the Period type labels. Maybe it's a question for @cooper-joe too.
> 
> Good question. If I understand correctly, then yes, it would be ideal if the labels in the period type dropdown reflected the chosen calendar as best as possible.

sure, the labels will be translated as per the locale(+calendar) but the period id remains the same.

abyot:
> 



> Do periods like `Six-monthly April` or `Financial year (Start October)` make sense in a non gregorian calendar? What do they actually mean in a Nepali calendar context for example?

I think six monthly periods that start on month 4 make sense - is that April (Gre) or Tahesas (Eth) or Shrawan (Nep) is again more of a mapping and translation thing. Also important to note here is we shouldn't expect all periods to be used. I don't think a country uses both Financial July and Financial October periods at the same time

abyot:
> > > Do periods like Six-monthly April or Financial year (Start October) make sense in a non gregorian calendar?
> > > What do they actually mean in a Nepali calendar context for example?
> > 
> > 
> > They would mean the equivalent nth month in the specific calendar. For FYNOV, November would mean the 11th month in the calendar (Hamle in Ethiopic, Falgun in Nepal). I believe @abyot give us this one as an example, as the financial year in Ethiopia goes from Hamle to Hamle, which will be FYNOV in our system, referring to the 11th month Hamle (even though this is actually July in the Gregorian calendar).
> > A better solution would be to redefine these to FY11, I guess, but I assume that would be a more far-reaching change for the backend.
> 
> I was thinking more about the labels, perhaps those can be different based on the calendar (if it makes more sense)?

@kabaros  Re: FYNOV - yes starts on Hamle, but can't finish on Hamle. It should finish one month before (Sene), so that Hamle becomes the next financial year's start month.

kabaros:
> @kabaros Re: FYNOV - yes starts on Hamle, but can't finish on Hamle. It should finish one month before (Sene), so that Hamle becomes the next financial year's start month.

yes, this is how it's implemented @abyot  - you can see the specs [here](https://github.com/dhis2/multi-calendar-dates/pull/7/files#diff-807abe6a30bd4b6088913dd9f120022e43bef331e2783ac349ea158fef8e93f4R207)

kabaros:
> Also important to note here is we shouldn't expect all periods to be used. I don't think a country uses both Financial July and Financial October periods at the same time

@abyot @edoardo do we have this ability now? is the list of periods to show to a user driven by some backend setting?

abyot:
> ethio-nepali-calendar.mov



> > @kabaros Re: FYNOV - yes starts on Hamle, but can't finish on Hamle. It should finish one month before (Sene), so that Hamle becomes the next financial year's start month.
> 
> yes, this is how it's implemented @abyot - you can see the specs [here](https://github.com/dhis2/multi-calendar-dates/pull/7/files#diff-807abe6a30bd4b6088913dd9f120022e43bef331e2783ac349ea158fef8e93f4R207)

great. saw in the video  ethio-nepali-calendar.mov Hamle - Hamle, may be that is a label thing.

abyot:
> > Also important to note here is we shouldn't expect all periods to be used. I don't think a country uses both Financial July and Financial October periods at the same time
> 
> @abyot @edoardo do we have this ability now? is the list of periods to show to a user driven by some backend setting?

no, we don't. users just pick what to use

kabaros:
@edoardo @cooper-joe @abyot I added [a task](https://dhis2.atlassian.net/browse/DHIS2-14435) to localise the dropdown list according to selected calendar - I will tackle it after this PR though, as I just want this one to be focused on getting parity with previous functionality. 

[DHIS2-14435]: https://dhis2.atlassian.net/browse/DHIS2-14435?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

kabaros:
> great. saw in the video ethio-nepali-calendar.mov Hamle - Hamle, may be that is a label thing.

ah true - the video was using a slightly outdated version of the library, I updated it now. 

image

(an aside - it's a bit fiddly to develop this functionality for me, as I have to update multi-calendar-dates and bump the version, then update analytics to use the updated version, then update data-visualiser to use updated analytics - maybe someone can suggest a better workflow)

kabaros:
another question @abyot  - I see a setting in the calendar settings to choose the date format, I am unclear what that setting should affect - does the backend use it to decide how the IDs are processed?

image

abyot:
> another question @abyot - I see a setting in the calendar settings to choose the date format, I am unclear what that setting should affect - does the backend use it to decide how the IDs are processed?
> 
> image

that has no effect on the period IDs.

package.json
HendrikThePendric: Well spotted @edoardo. Yeah, the `name` field in the [`package.json`](https://github.com/dhis2/multi-calendar-dates/blob/beta/package.json#L2) in should be updated to include the `@dhis2` namespace, like [here](https://github.com/dhis2/app-runtime/blob/master/runtime/package.json#L2).

src/components/PeriodDimension/PeriodTransfer.js
HendrikThePendric: Yeah, I think it's safe to remove the default here, since the provider only renders if the systemInfo was fetched successfully.

HendrikThePendric (commented):
Great to have had so many eyes on this already.

One point that seems to have been brought up from different perspectives in this PR is how we deal with available period-types per calendar system.

For the Ethiopian calendar we identified an equivalent for "Financial Year Nov", but can we do the same for Nepali? And are we confident the same can be said for whatever other calendar system? If not we'll need some way to restrict period-types per calendar system.

I believe @abyot informed us that it is not possible to retrieve supported period types from the backend. And if this is indeed the case, then we should probably have a data structure in the multi-calendar-dates lib that lists supported period-types per calendar system.

package.json
kabaros: 👍🏿  updated the package name to be `@dhis2/multi-calendar-dates`

src/components/PeriodDimension/PeriodTransfer.js
kabaros: updated

src/components/PeriodDimension/PeriodTransfer.js
```diff
@@ -23,13 +25,20 @@ const PeriodTransfer = ({
...
     const defaultFixedPeriodType = excludedPeriodTypes.includes(MONTHLY)
-        ? getFixedPeriodsOptionsById(QUARTERLY)
-        : getFixedPeriodsOptionsById(MONTHLY)
-    const defaultFixedPeriodYear = new Date().getFullYear()
+        ? getFixedPeriodsOptionsById(QUARTERLY, calendar)
+        : getFixedPeriodsOptionsById(MONTHLY, calendar)
+
+    // use ".eraYear" rather than ".year" because in Ethiopian calendar, eraYear is what our users expect to see (for other calendars, it doesn't matter)
+    // there is still a pending decision in Temporal regarding which era to use by default: https://github.com/js-temporal/temporal-polyfill/blob/9350ee7dd0d29f329fc097debf923a517c32f813/lib/calendar.ts#L1964
+    const defaultFixedPeriodYear = getNowInCalendar(calendar).eraYear
```
tomzemp: Does this need to pass the timezone as well? The implementation of getNowInCalendar uses UTC by default whereas previously this was using browser time with new Date().getFullYear(), so I guess there would be differences around the new year?

src/components/PeriodDimension/utils/fixedPeriods.js
tomzemp: I also wondered this. Would it be better to perhaps set `reversePeriods: true` if the property is missing from the config?

src/components/PeriodDimension/utils/fixedPeriods.js
```diff
@@ -559,128 +232,134 @@ const filterFuturePeriods = (periods) => {
...
+    return [
+        {
+            id: DAILY,
+            getPeriods: getDailyPeriodType(filterFuturePeriods, calendar),
+            name: i18n.t('Daily'),
+        },
+        {
+            id: WEEKLY,
+            getPeriods: getWeeklyPeriodType(
+                { shortName: '', startDay: 1 },
```
tomzemp: what is the use of the `shortName`? It doesn't seem to be passed further? Also, the `startDay` seems superfluous as the underlying getWeeklyPeriods.ts already has logic for determining the startDay based on the periodType?

tomzemp (commented):
Looks good to me @kabaros! I left some small comments. Since this fixedPeriods.js script got copy/pasted into a number of other apps (e.g. approvals and aggregate data entry), I do wonder if it would be useful to move more of the logic from there into the new multicalendar engine (e.g. putting the closures for getting periods there), so that there is less logic remaining in the various frontend apps?

src/components/PeriodDimension/PeriodTransfer.js
kabaros: I [changed the library](https://github.com/dhis2/multi-calendar-dates/pull/7/commits/03caf6d4a8c0b83924cf47e0aacecb7987168d80#diff-c7d7a37d7a7bb6f37d24e469f78b61d97734db2b1fb3c8332e89e485ac8cf64c) to default to browser timezone rather than UTC .. I think that's a more sensible default, right?

src/components/PeriodDimension/utils/fixedPeriods.js
kabaros: I removed shortName .. `startDay` is needed though - the underlying `getWeeklyPeriods` takes either a startDay or a custom period type like `WEEKLYSAT` to decide the start date. In this case, I kept it using the `startDay` rather than having a map here from the startDay to the type (i.e. startDay = 3 to `WEEKLYWED`)

tomzemp: 👍 Right. Got that re: `startDay`

src/components/PeriodDimension/PeriodTransfer.js
tomzemp: Agreed. That seems like the better default.

src/components/PeriodDimension/utils/fixedPeriods.js
kabaros: ok, I changed this to make the decision higher up [here](https://github.com/dhis2/analytics/pull/1407/files#diff-fd50894017c508c3a737d8e955e26b0b39b76efe96501f53b400da72523e7960R50) based on the period type. This is what the older code was doing in effect  - deciding depending on period type whether to reverse or not. The reason that it was using the same config is that in some period types, it did `isReverse ? periods.reverse() : periods` and in other` it did `isReverse ? periods : periods.reverse()` (the opposite)

package.json
```diff
@@ -48,6 +48,11 @@
         "styled-jsx": "^4.0.1",
         "typeface-roboto": "^0.0.75"
     },
+    "jest": {
+        "transformIgnorePatterns": [
+            "/!node_modules\\/multi-calendar-dates/"
+        ]
+    },
```
HendrikThePendric, line 55: Why is this needed?

kabaros: it is to fix this error with jest .. it's similar to the fix we apply in the case of dhis/ui and some other projects too in few projects like [here](https://github.com/dhis2/data-visualizer-app/blob/master/package.json#L65) as jest doesn't run babel transformation in imported modules by default.
image

kabaros:
> For the Ethiopian calendar we identified an equivalent for "Financial Year Nov", but can we do the same for Nepali? And are we confident the same can be said for whatever other calendar system? If not we'll need some way to restrict period-types per calendar system.
> 
> I believe @abyot informed us that it is not possible to retrieve supported period types from the backend. And if this is indeed the case, then we should probably have a data structure in the multi-calendar-dates lib that lists supported period-types per calendar system.

I created an issue [here](https://dhis2.atlassian.net/browse/DHIS2-14680) to track this discussion about how to handle which calendars to show for each calendar (if we need to).

src/components/PeriodDimension/PeriodTransfer.js
edoardo: I don't see it updated?

edoardo (approved):

src/components/PeriodDimension/PeriodTransfer.js
kabaros: updated now

edoardo (approved):

edoardo (approved):

PR #1413, opened 2023-01-24
fix: use page number from response DHIS2-11505

edoardo:
Fixes an issue related to [DHIS2-11505](https://dhis2.atlassian.net/browse/DHIS2-11505)

---

### Key features

1. reset current page in local state from pager data in response

---

### Description

This is to avoid the edge case where a page does not exist anymore after deleting the visualization that was listed in it.

The page in the component state needs to be set to the page returned in the response when it differs.

Unfortunately there is an extra request with the not existing page, because until we get the response from the api we don't know if a page has been removed or not as a consequence of deleting visualizations.

---

### Screenshots

_supporting text_

[DHIS2-11505]: https://dhis2.atlassian.net/browse/DHIS2-11505?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/OpenFileDialog/OpenFileDialog.js
+++ b/src/components/OpenFileDialog/OpenFileDialog.js
@@ -144,6 +144,11 @@ export const OpenFileDialog = ({
 
     const { loading, error, data, refetch } = useDataQuery(filesQuery, {
         lazy: true,
+        onComplete: (response) => {
+            if (page !== response.files.pager.page) {
+                setPage(response.files.pager.page)
+            }
+        },
     })
 
     const resetFilters = () => {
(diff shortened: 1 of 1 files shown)


martinkrulltott (approved):
Code LGTM, but the actual fix is tricky to understand. Might help with some screenshots. Or even better some tests (in DV or LL).

edoardo:
> Code LGTM, but the actual fix is tricky to understand. Might help with some screenshots. Or even better some tests (in DV or LL).

I'll add some tests, perhaps in LL where we have `fileMenu` tests.

PR #1415, opened 2023-01-31
refactor: rewrote fetch logic for infinite scrolling/filtering/search

edoardo:
Part of [DHIS2-13871](https://dhis2.atlassian.net/browse/DHIS2-13871)

---

### Key features

1. rewrote the fetch logic to solve issues with pagination when combined with search and filters
2. added loading spinner while data is fetched
3. populate group filter selector

---

### Description

Infinite scrolling, group and disaggregation (subGroup) filter, search all seem to work.
The loading spinner has the same look as the one used in the `Transfer` component in the `DataDimension` modal.
The group dropdown has been implemented reusing the same fetch function for groups.

### DIFF
--- a/src/components/DataDimension/DataElementSelector.js
+++ b/src/components/DataDimension/DataElementSelector.js
@@ -263,37 +263,39 @@ const DataElementSelector = ({ displayNameProp, onSelect }) => {
                     onChange={(subGroup) => onFilterChange({ subGroup })}
                 />
             </div>
-            <div className="dimension-list-wrapper" ref={rootRef}>
-                {loading && (
-                    <Cover translucent>
-                        <Center>
-                            <CircularLoader />
-                        </Center>
-                    </Cover>
-                )}
-                <div className="dimension-list">
-                    {options.map(({ label, value, type, disabled }) => (
-                        <TransferOption
-                            label={label}
-                            key={value}
-                            value={value}
-                            icon={getIcon(type)}
-                            tooltipText={getTooltipText({
-                                type,
-                            })}
-                            disabled={disabled}
-                            onClick={Function.prototype}
-                            onDoubleClick={onSelect}
-                        />
-                    ))}
-
-                    <div className="scroll-detector">
-                        <IntersectionDetector
-                            onChange={onEndReached}
-                            rootRef={rootRef}
-                        />
+            <div className="dimension-list-container">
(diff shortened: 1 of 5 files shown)


src/components/DataDimension/DataElementSelector.js
```diff
@@ -86,59 +123,63 @@ const DataElementSelector = ({
...
+            // XXX the following cannot be in here I believe, if it's still needed it should be triggered somewhere else, perhaps in a useEffect that listens to options
+            // but it's not possible to call fetchData() within the effect
 
-            setHasNextPage(result.nextPage)
-            /*  The following handles a very specific edge-case where the user can select all items from a 
-            page and then reopen the modal. Usually Transfer triggers the onEndReached when the end of 
-            the page is reached (scrolling down) or if too few items are on the left side (e.g. selecting 
-            49 items from page 1, leaving only 1 item on the left side). However, due to the way Transfer 
-            works, if 0 items are available, more items are fetched, but all items are already selected 
+            /*  The following handles a very specific edge-case where the user can select all items from a
```
martinkrulltott: I think this can be removed. The edge-case gets even more edge'y for calculations, as the odds of someone already having the 50 first items in their calculation is very low (1. you probably don't write such long calculations, 2. there's no "add all" button as you have in the `Transfer` component, which makes it even harder to do).
If we find this to be a valid use case during our testing we can always add it in later. So I'd say ignore/remove this for now.

edoardo: Removed.

src/components/DataDimension/DataElementSelector.js
```diff
@@ -80,112 +122,123 @@ DisaggregationSelector.propTypes = {
...
-                searchTerm,
+                filter: filterRef.current,
+                searchTerm: searchTermRef.current,
+                page: pageRef.current,
             })
-            const newOptions = []
-            result.dimensionItems?.forEach((item) => {
-                newOptions.push({
+
+            // XXX also check for length?
```
martinkrulltott, line 156: needed?

martinkrulltott (approved):

src/components/DataDimension/DataElementSelector.js
edoardo: No, `dimensionItems` should always be an array when defined.
Will be removed.

PR #1416, opened 2023-02-02
feat: dnd for custom calculations

jenniferarnesen:
dnd for calculations

Some technical notes:
* DndContext is mostly copied from Line-list. The only thing that has really changed is the handleDragEnd function.
* When an item is added to the formula, it is assigned an id based on the id or label of the item and a counter. Dnd requires that each draggable item have an id.
* In the INPUT formula item, there is a setTimeout that was needed in order to get the cursor to stay in the input element once it was added to the formula.
* In the FormulaField, there is a special, invisible drop zone always present in the formula field. It makes it possible to drop an item before the first item.  This is the DropZone component. This is almost identical to what we have in LL
* Also in FormulaField, there is a little trickery that makes it possible to show the full drop indicator on top of the border.  This is using the <div className="border> element. The problem was that with `overflow` set on the formula field, the drop indicator is getting cut off. An alternative to the current solution would be to increase the padding inside FormulaField to leave enough space for the full drop indicator to show, but that means a padding of 12px, which looked strange to me. But it's worth discussing.

Give it a try by doing this: 
1. pull this branch
2. yarn install && yarn start
3. go to localhost:5000 (storybook) and try the CalculationsModal component

### DIFF
--- a/src/components/DataDimension/CalculationModal.js
+++ b/src/components/DataDimension/CalculationModal.js
@@ -25,9 +25,17 @@ import {
     parseExpressionToArray,
     parseArrayToExpression,
 } from '../../modules/expressions.js'
+import { TransferOption } from '../TransferOption.js'
 import DataElementSelector from './DataElementSelector.js'
 import FormulaField from './FormulaField.js'
-import MathOperatorSelector, { getOperators } from './MathOperatorSelector.js'
+import { FormulaItem } from './FormulaItem.js'
+import MathOperatorSelector, {
+    getOperators,
+    TYPE_DATAITEM,
+    TYPE_OPERATOR,
+    TYPE_INPUT,
+} from './MathOperatorSelector.js'
+import { Operator } from './Operator.js'
 import styles from './styles/CalculationModal.style.js'
 
 const VALID_EXPRESSION = 'OK'
@@ -70,7 +78,7 @@ const CalculationModal = ({
 
     const sensor = useSensor(PointerSensor, activateAt15pixels)
     const sensors = useSensors(sensor)
-    const [draggingId, setDraggingId] = useState(null)
+    const [draggingItem, setDraggingItem] = useState(null)
     const [idCounter, setIdCounter] = useState(0)
 
     const expressionStatus = validationOutput?.status
@@ -129,6 +137,8 @@ const CalculationModal = ({
         setValidationOutput(result)
     }
 
+    console.log('expressionArray', expressionArray)
+
     return (
         <>
             <Modal dataTest={`calculation-modal`} position="top" large>
@@ -156,27 +166,21 @@ const CalculationModal = ({
                             <div className="left-section">
                                 <DataElementSelector
                                     displayNameProp={displayNameProp}
-                                    onSelect={({ value }) => {
+                                    onSelect={({ label, value }) => {
                                         setValidationOutput()
                                         setExpressionArray((prevArray) =>
                                             prevArray.concat([
-                                                {
-                                                    label: value,
-                                                    value: `#{${value}}`,
-                                                },
+                                                getNewDataItem(value, label),
                                             ])
                                         )
                                     }}
                                 />
                                 <MathOperatorSelector
-                                    onSelect={({ value }) => {
+                                    onSelect={({ id }) => {
                                         setValidationOutput()
                                         setExpressionArray((prevArray) =>
                                             prevArray.concat([
-                                                {
-                                                    label: value,
-                                                    value,
-                                                },
+                                                getNewOperatorItem(id),
                                             ])
                                         )
                                     }}
@@ -250,9 +254,9 @@ const CalculationModal = ({
                         </div>
                         <style jsx>{styles}</style>
                         <DragOverlay dropAnimation={null}>
-                            {draggingId ? (
-                                <span className={styles.dragOverlay}>
-                                    {draggingId}
+                            {draggingItem ? (
+                                <span className="dragOverlay">
+                                    {getDraggingItem()}
                                 </span>
                             ) : null}
                         </DragOverlay>
@@ -365,8 +369,17 @@ const CalculationModal = ({
         const destAxisId = over.data.current?.sortable?.containerId || over.id
 
         if (sourceAxisId === OPTIONS_PANEL) {
-            // adding an item to the formula
-            const newItem = getNewItem(active.id)
+            let newItem
+            if (active.data.current.type === 'DATA_ELEMENT') {
+                newItem = getNewDataItem(
+                    active.data.current.id,
+                    active.data.current.label
+                )
+            } else {
+                // adding an item to the formula
+                newItem = getNewOperatorItem(active.id)
+            }
+
             if (destAxisId === LAST_DROPZONE_ID) {
                 setExpressionArray([...expressionArray, newItem])
             } else if (destAxisId === FORMULA_BOX_ID) {
@@ -398,14 +411,38 @@ const CalculationModal = ({
             sourceList.splice(destIndex, 0, moved)
             setExpressionArray(sourceList)
         }
-        setDraggingId(null)
+        setDraggingItem(null)
     }
     function handleDragStart({ active }) {
-        setDraggingId(active.id)
+        console.log('drag started', active.data.current)
+        setDraggingItem(active.data.current)
     }
 
     function handleDragCancel() {
-        setDraggingId(null)
+        setDraggingItem(null)
+    }
+
+    function getDraggingItem() {
+        const label = draggingItem.value || draggingItem.label
+        if (draggingItem.sortable.containerId === FORMULA_BOX_ID) {
+            // dragging item looks like a formula item
+            return (
+                <FormulaItem value={draggingItem.value}>
+                    <span>{draggingItem.value}</span>
+                </FormulaItem>
+            )
+        } else {
+            if ([TYPE_OPERATOR, TYPE_INPUT].includes(draggingItem.type)) {
+                return <Operator label={label} />
+            } else {
+                return (
+                    <TransferOption
+                        label={draggingItem.label}
+                        value={draggingItem.value}
+                    />
+                )
+            }
+        }
     }
 
     function setExpressionItemValue({ index, value }) {
@@ -415,12 +452,21 @@ const CalculationModal = ({
         setExpressionArray(updatedFormulaItems)
     }
 
-    function getNewItem(activeId) {
-        const itemToCopy = getOperators().find(
-            (operator) => operator.id === activeId
-        )
+    function getNewDataItem(id, label) {
+        const newItem = {
+            id: `${id}-${idCounter}`,
+            value: `#{${id}}`,
+            label: label,
+            type: TYPE_DATAITEM,
+        }
+        setIdCounter(idCounter + 1)
+        return newItem
+    }
+
+    function getNewOperatorItem(id) {
+        const itemToCopy = getOperators().find((operator) => operator.id === id)
         const newItem = Object.assign({}, itemToCopy, {
-            id: `${activeId}-${idCounter}`,
+            id: `${id}-${idCounter}`,
         })
 
         setIdCounter(idCounter + 1)
--- a/src/components/DataDimension/CalculationModal.js
+++ b/src/components/DataDimension/CalculationModal.js
@@ -134,6 +134,7 @@ const CalculationModal = ({
                 </ModalTitle>
                 <ModalContent dataTest={'calculation-modal-content'}>
                     <DndContext
+                        collisionDetection={rectIntersectionCustom}
                         onDragStart={handleDragStart}
                         onDragEnd={handleDragEnd}
                         sensors={sensors}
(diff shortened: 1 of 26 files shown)


src/components/DataDimension/CalculationModal.js
```diff
@@ -128,97 +114,94 @@ const CalculationModal = ({
...
-                                    {parseArrayToExpression(expressionArray)}
-                                </p>
-                                <div className="actions-wrapper">
-                                    <Button small onClick={validateExpression}>
-                                        {/* TODO: add loading state to button? */}
-                                        {i18n.t('Check formula')}
-                                    </Button>
-                                    {(selectedPart || selectedPart === 0) && (
-                                        <div className={'remove-button'}>
+                                <div className="leftpad">
```
martinkrulltott: Rename to something descriptive of what it is (not what it does), seems to be the pattern we're going for in general. Also, should use `kebab-case`.

src/components/DataDimension/CalculationModal.js
```diff
@@ -128,97 +114,94 @@ const CalculationModal = ({
...
+                                            expressionArray
+                                        )}
+                                    </p>
+                                    <div className="actions-wrapper">
+                                        <Button small onClick={validate}>
+                                            {/* TODO: add loading state to button? */}
+                                            {i18n.t('Check formula')}
+                                        </Button>
+                                        {(selectedId || selectedId === 0) && (
+                                            <div className={'remove-button'}>
```
martinkrulltott:
```suggestion
                                            <div className="remove-button">
```

src/components/DataDimension/DraggingItem.js
```diff
@@ -0,0 +1,47 @@
...
+    const displayLabel =
+        item.type === TYPE_INPUT ? item.value || item.label : item.label
+
+    return (
+        <>
+            <div
+                className={cx('dragging-item', {
+                    operator: item.type === TYPE_OPERATOR,
+                    number: item.type === TYPE_INPUT,
+                    dataelement: item.type === TYPE_DATAELEMENT,
```
martinkrulltott:
```suggestion
                    dataElement: item.type === TYPE_DATA_ELEMENT,
```

src/components/DataDimension/constants.js
```diff
@@ -0,0 +1,7 @@
+export const TYPE_INPUT = 'input'
+export const TYPE_OPERATOR = 'operator'
+export const TYPE_DATAELEMENT = 'DATA_ELEMENT'
```
martinkrulltott: why not just use `DIMENSION_TYPE_DATA_ELEMENT` that already exists?

src/components/DataDimension/DraggingItem.js
martinkrulltott: actually, the key is a css class, so should be `kebab-case` and the condition could just use `DIMENSION_TYPE_DATA_ELEMENT` instead, so that makes for:
```suggestion
                    data-element: item.type === DIMENSION_TYPE_DATA_ELEMENT,
```
(possibly needs to wrap the id in quotes)

src/components/DataDimension/DraggingItem.js
```diff
@@ -0,0 +1,47 @@
...
+    item: PropTypes.shape({
+        label: PropTypes.string,
+        sortable: PropTypes.object,
+        type: PropTypes.string,
+        value: PropTypes.string,
+    }),
+}
+
+DraggingItem.defaultProps = {
+    onClick: Function.prototype,
```
martinkrulltott: how/where is this used? I don't see `onClick` referenced anywhere in the component

src/components/DataDimension/DraggingItem.js
```diff
@@ -0,0 +1,47 @@
...
+                )}
+                <span>{displayLabel}</span>
+            </div>
+            <style jsx>{styles}</style>
+        </>
+    )
+}
+
+DraggingItem.propTypes = {
+    item: PropTypes.shape({
```
martinkrulltott: if the component only accepts a single prop and that prop is never used as a whole, then maybe it's better to just change the input prop to the individual parts instead? i.e.
`const DraggingItem = ({ label, sortable, type, value }) => {`

src/components/DataDimension/DndContext.js
```diff
@@ -0,0 +1,165 @@
...
+    DragOverlay,
+    useSensor,
+    useSensors,
+    PointerSensor,
+} from '@dnd-kit/core'
+import PropTypes from 'prop-types'
+import React, { useState } from 'react'
+import DraggingItem from './DraggingItem.js'
+
+const activateAt15pixels = {
```
martinkrulltott: weird name. also, why does it need to be a const? (only used once and it's not that complex)

src/components/DataDimension/DndContext.js
```diff
@@ -0,0 +1,165 @@
...
+            handleDragCancel()
+            return
+        }
+
+        const activeItem = {
+            id: active.id,
+            sourceAxisId: active.data.current?.sortable?.containerId,
+            sourceIndex: active.data.current?.sortable?.index,
+            data: {
+                id: active.data.current.id,
```
martinkrulltott: all the other props use optional chaining, add that here as well?
```suggestion
                id: active.data.current?.id,
```

src/components/DataDimension/DndContext.js
```diff
@@ -0,0 +1,165 @@
...
+        }
+        const overItem = {
+            axisId: over.data.current?.sortable?.containerId || over.id,
+            index: over.data.current?.sortable?.index,
+        }
+        onDragEnd({ activeItem, over: overItem })
+        setDraggingItem(null)
+    }
+
+    function getIntersectionRatio(entry, target) {
```
martinkrulltott: move this and other pure helper functions to a separate file? (or just outside of the component?)

src/components/DataDimension/DndContext.js
```diff
@@ -0,0 +1,165 @@
...
+        return b - a
+    }
+
+    function rectIntersectionCustom({
+        pointerCoordinates,
+        droppableContainers,
+    }) {
+        // create a rect around the pointerCoords for calculating the intersection
+
+        const pointerRectWidth = 40
```
martinkrulltott: I don't know where `40` comes from or how this works, but if the width and height are equal then maybe just use a single prop instead of two? (e.g. `pointerRectSize`?)

src/components/DataDimension/FormulaItem.js
```diff
@@ -0,0 +1,197 @@
...
+                        x: transform.x,
+                        y: transform.y,
+                        scaleX: 1,
+                        scaleY: 1,
+                    }),
+              transition,
+          }
+        : undefined
+
+    let insertPosition = undefined
```
martinkrulltott:
```suggestion
    let insertPosition
```

src/components/DataDimension/FormulaItem.js
```diff
@@ -0,0 +1,197 @@
...
+        )
+    }
+
+    return (
+        <>
+            <div
+                ref={setNodeRef}
+                {...attributes}
+                {...listeners}
+                className={isLast && 'last-item'}
```
martinkrulltott: this works, but I'd still suggest using `cx` for consistency
```suggestion
                className={cx({ 'last-item': isLast })}
```

src/components/DataDimension/FormulaItem.js
```diff
@@ -0,0 +1,197 @@
...
+        </>
+    )
+}
+
+FormulaItem.propTypes = {
+    onChange: PropTypes.func.isRequired,
+    onClick: PropTypes.func.isRequired,
+    onDoubleClick: PropTypes.func.isRequired,
+    hasFocus: PropTypes.bool,
+    id: PropTypes.string,
```
martinkrulltott: `onDoubleClick` is required and on line 110 it's used with `id` (`onDoubleClick(id)`) so that would make `id` required too?

src/components/DataDimension/MathOperatorSelector.js
```diff
@@ -1,44 +1,48 @@
...
-const MathOperatorSelector = ({ onSelect }) => {
-    const operators = [
-        { value: '+', label: '+' },
-        { value: '-', label: '-' },
-        { value: '*', label: '×' },
-        { value: '/', label: '/' },
-        { value: '(', label: '(' },
-        { value: ')', label: ')' },
+export const getOperators = () => {
+    return [
```
martinkrulltott:
```suggestion
   [
``` 
and remove `{ }`

src/components/DataDimension/Operator.js
```diff
@@ -0,0 +1,36 @@
+import { useSortable } from '@dnd-kit/sortable'
+import { CSS } from '@dnd-kit/utilities'
+import PropTypes from 'prop-types'
+import React from 'react'
+import styles from './styles/Operator.style.js'
+
+const DraggableOperator = ({ id, label, type, onDoubleClick }) => {
```
martinkrulltott: File name is called `Operator` but the component is called `DraggableOperator`, change one or the other?

src/components/DataDimension/DropZone.js
```diff
@@ -0,0 +1,39 @@
...
+    } else {
+        draggingOver = firstElementId === active?.id ? false : isOver
+    }
+
+    return (
+        <>
+            <div
+                ref={setNodeRef}
+                className={cx('first-dropzone', {
+                    over: draggingOver,
```
martinkrulltott: As I understand it `over` is dnd kit lingo, but for our class name maybe we can use something more descriptive? (and less dnd-kit'y)

src/components/DataDimension/styles/DropZone.style.js
```diff
@@ -0,0 +1,51 @@
...
+        position: absolute;
+    }
+
+    /* the vertical line */
+    .over::before {
+        top: 10px;
+        width: 4px;
+        left: 4px;
+        height: 18px;
+        background-color: #4c9ffe;
```
martinkrulltott: is there a `ui` color for this?

src/components/DataDimension/styles/FormulaItem.style.js
```diff
@@ -0,0 +1,136 @@
...
+        padding: 2px;
+    }
+
+    .input-positioner {
+        position: relative;
+        line-height: 18px;
+        margin-right: 1px;
+    }
+
+    input {
```
martinkrulltott: use a class name?

src/components/DataDimension/styles/FormulaItem.style.js
```diff
@@ -0,0 +1,136 @@
...
+        white-space: nowrap;
+    }
+
+    /* highlighted items */
+
+    .highlighted {
+        background: ${theme.secondary800};
+    }
+
+    .highlighted input {
```
martinkrulltott: same as above, use class name instead of element type?

src/components/DataDimension/DndContext.js
jenniferarnesen: They happen to be the same, but they don't have to be. My goal is to refactor this function so it can be used by multiple dnd implementations, so I'll leave it as is.

src/components/DataDimension/constants.js
jenniferarnesen: It was because I wanted to keep the definitions all in one place and have the same naming pattern.

src/components/DataDimension/DndContext.js
jenniferarnesen: Cleaned this up. Don't need as many conditionals as I thought.

src/components/DataDimension/FormulaItem.js
jenniferarnesen: Switching to index

src/components/DataDimension/constants.js
martinkrulltott: Ok, makes sense I suppose. But if you insist on creating a clone of DE it should the const it's cloning, i.e.
```suggestion
export const TYPE_DATAELEMENT = DIMENSION_TYPE_DATA_ELEMENT
```

I would also suggest to use capital snake case which seems to be our de facto standard, i.e.
```suggestion
export const TYPE_DATA_ELEMENT = DIMENSION_TYPE_DATA_ELEMENT
```

src/components/DataDimension/DropZone.js
jenniferarnesen: 'dragging-over'

src/components/DataDimension/styles/DropZone.style.js
jenniferarnesen: No, not right now. But I'll let Joe know. This is the color of the blue drop marker, and there are some blue colors in the design system.

jenniferarnesen: @cooper-joe

src/components/DataDimension/constants.js
martinkrulltott: Actually, thinking about it further, since it's actually exported, it would be best to prefix "TYPE" with what it is, e.g. 
```suggestion
export const DRAGGABLE_TYPE_DATA_ELEMENT = DIMENSION_TYPE_DATA_ELEMENT
```
or sth..

src/components/DataDimension/styles/DropZone.style.js
cooper-joe: Nice catch. Not sure where `#4c9ffe` came from (maybe the original plugin...or past-me) -- anyway, `blue500` can be used instead.

src/components/DataDimension/DndContext.js
jenniferarnesen: I've moved them outside the component. When we factor out a set of dnd tools/components for reuse
, then I'll move it to a separate file

src/components/DataDimension/CalculationModal.js
jenniferarnesen: Fixed.

src/components/DataDimension/constants.js
jenniferarnesen: Fixed.

jenniferarnesen: How about CALCULATION_TYPE_DATA_ELEMENT (and OPERATOR and NUMBER)

PR #1421, opened 2023-02-06
fix: revert "Merge branch 'use-multicalendar-periods' into next"

kabaros:
This reverts commit 0082726c72bf6d199b2a5102890dd7dabd5bb02e, reversing changes made to 766e10fde36fda4a8ad77ba5203c761aa37db260.

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,11 +48,6 @@
         "styled-jsx": "^4.0.1",
         "typeface-roboto": "^0.0.75"
     },
-    "jest": {
-        "transformIgnorePatterns": [
-            "/!node_modules\\/multi-calendar-dates/"
-        ]
-    },
     "peerDependencies": {
         "@dhis2/app-runtime": "^3",
         "@dhis2/d2-i18n": "^1.1",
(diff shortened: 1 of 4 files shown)


ghost:
One thing we should check is the correct syntax for the revert. I'm not sure which conventional commit format we use again. I'll take a look.

Edit:

- So judging by our commitlint config [here](https://github.com/dhis2/cli-style/blob/3d277df447345ebdaedefdfdcfac9e81ba9c74f8/config/commitlint.config.js#L2) we use `@commitlint/config-conventional`. So we don't have to worry about the angular conventions mentioned [here](https://github.com/semantic-release/semantic-release/blob/master/docs/support/FAQ.md#how-can-i-revert-a-release).
- In terms of the revert commit message formatting, the docs mention: `Otherwise, revert the commit that introduced the bug and push the revert commit (or merge it via a pull request) to the release branch`. So I don't think you have to change the default message. And [this](https://github.com/conventional-changelog/conventional-changelog/blob/0e05028f70bbd3109e1a4b16262a9450153060de/packages/conventional-commits-parser/index.js#L26) seems to confirm that a regular revert is fine.

Your commit looks like a regular revert to me, so assuming that this revert everything you want to revert you should be fine.

ghost (approved):

PR #1422, opened 2023-02-06
chore: merge next branch to master

kabaros:
merges next branch to master to avoid issues with publishing to npm in the future. More context in [this conversation](https://dhis2.slack.com/archives/CPGUFVC56/p1675682666142759).

### Background

In order to create a release of Data Visualiser that contains the work on [multi-calendar analytics](https://github.com/dhis2/analytics/pull/1407). I created a `next` branch for the multi-calendar work for it to be released to the `next` release channel. 

### The issue

What I failed to realise is that `next` is treated in a special way other than pre-release branches like alpha or beta, so once a release is made on next (v 24.5.0 in this case) , then releases on master are not allowed unless they are in the range of the current master release version (24.4.2) and next branch, meaning in this case that any attempt to publish a minor version on master would have failed.

My attempts to solve the issue afterwards made it worse: I unpublished the specific package from npm (running `npm unpublish @dhis/analytics@2.4.5`). This gave a false positive that it was unpublished. I, then, force-pushed `next` after hard-resetting it to match `master`. The force-push made the tag 24.5.0 in git point to nothing, which makes it then impossible to have more release on `next` since the unpublish also failed.

### The fix

The correct procedure should have been to revert the change on `next` with a revert commit, which would create a new release on `next` then merge `next` branch to `master` to bring master inline with the next channel. But, after the force-push, we needed to do more. The way we went about fixing the issue was to:

- Remove the original git tag for 24.5.0 and the release notes, and point them to the correct commit using the steps described [here](https://github.com/semantic-release/semantic-release/blob/master/docs/support/troubleshooting.md#release-not-found-release-branch-after-git-push---force).
  - The git tag is used by semantic-release to know what changed since the last version.
- The changes on `next` were reverted, so that merging this branch to `master` doesn't change any functionality.
-  After merging this branch to master, releases on master should no longer be affected.

### Lessons learned

- **Don't force-push on release branches** (we're opening a ticket to enforce that, preferably at an organisation-level).
- Think twice before creating a `next` release channel and  learn how pre-release branches (like `alpha` or `beta`) differ from `next` and `next-major`. More info [here](https://semantic-release.gitbook.io/semantic-release/usage/workflow-configuration#release-branches).
- Don't use `npm unpublish` - It has [many conditions](https://docs.npmjs.com/policies/unpublish) for it to work as expected, and it does not give feedback on success or failures.
- Don't push anything late before a weekend! - this whole sequence of events started with creating the next branch based on my branch instead of master, which I didn't mean .. if the branch was based on master, it wouldn't have triggered the publish and all the subsequent events.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@
 
 * use multi-calendar-dates library to generate fixed periods ([acc3801](https://github.com/dhis2/analytics/commit/acc380156092f888cfa55f70b4bade1c6516d034))
 
+# [24.5.0](https://github.com/dhis2/analytics/compare/v24.4.2...v24.5.0) (2023-02-06)
+
+
+### Features
+
+* use multi-calendar-dates library to generate fixed periods ([acc3801](https://github.com/dhis2/analytics/commit/acc380156092f888cfa55f70b4bade1c6516d034))
+
 ## [24.4.2](https://github.com/dhis2/analytics/compare/v24.4.1...v24.4.2) (2023-01-25)
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,14 +3,7 @@
 
 ### Features
 
-* use multi-calendar-dates library to generate fixed periods ([acc3801](https://github.com/dhis2/analytics/commit/acc380156092f888cfa55f70b4bade1c6516d034))
-
-# [24.5.0](https://github.com/dhis2/analytics/compare/v24.4.2...v24.5.0) (2023-02-06)
-
-
-### Features
-
-* use multi-calendar-dates library to generate fixed periods ([acc3801](https://github.com/dhis2/analytics/commit/acc380156092f888cfa55f70b4bade1c6516d034))
+* [reverted] use multi-calendar-dates library to generate fixed periods ([acc3801](https://github.com/dhis2/analytics/commit/acc380156092f888cfa55f70b4bade1c6516d034)). This work was added to "next" release channel, but was reverted afterwards because of the restrictions "next" channel puts on what can be released from master afterwards. It will be added to "alpha" channel instead until it's ready for the main channel.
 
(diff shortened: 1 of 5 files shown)


ghost:
@kabaros I think we can clean up the changelog a bit so it doesn't confuse users. Remove the duplicate and maybe add a note about the revert?

ghost:
To explain a little what has been fixed on next:

- The 24.5.0 tag was pointing to a commit no longer on next, this broke semantic release
- The tag was moved to an existing commit, notes were added (see [docs](https://github.com/semantic-release/semantic-release/blob/master/docs/support/troubleshooting.md#release-not-found-release-branch-after-git-push---force) and fixed release process [here](https://github.com/dhis2/analytics/actions/runs/4105461343/jobs/7082349696#step:7:45))
- The changes were reverted, so that merging this branch doesn't change any functionality

After merging this to master releases on master should no longer be affected (let me know if I'm missing anything @kabaros)

Mohammer5 (approved):

ghost (approved):

kabaros:
@ismay I added more context about what I did wrong and how we fixed in the [PR description](https://github.com/dhis2/analytics/pull/1422)

ghost:
> @ismay I added more context about what I did wrong and how we fixed in the [PR description](https://github.com/dhis2/analytics/pull/1422)

Thanks! Though I think this is something that Hendrik and I didn't realise either, so don't feel bad about not catching the special limitations of using next as a release branch. As I mentioned in the meeting, I think this is also a grey area currently, where we don't have clear guidelines, docs, best practices, etc. You ran into this, but it might as well have been someone else.

I think what we discussed should help prevent this from happening in the future: a pull request template that clarifies (or links to) the exact usecases and best practices for this and other release scenarios. On top of that, synchronizing branch protection rules across our org should prevent release branches from being accidentally rewritten. Since protecting branches is a manual process that requires special priviliges it can happen that branches aren't protected as well as they should be.

janhenrikoverland (approved):

PR #1433, opened 2023-02-21
feat: cache param for expression dimension items (DHIS2-13871)

janhenrikoverland:
Implements a small part of [DHIS2-13871](https://dhis2.atlassian.net/browse/DHIS2-13871)

- `hash` creates a deterministic fixed-length hash of the input string, it's a separate function with its own test to make it easy to replace the hash function later
- `getHash` accepts a string or an array of strings and returns the hash produced by `hash`
- `getExpressionHashFromVisualization` grabs the expressions (looks for `dimensionItemType` and `expression`), if any EDIs in the visualization, then sorts them by id to get a cache hit as often as possible (now the order of the selected EDI items doesn't matter), and returns the result from `getHash`

In AnalyticsRequest we add the hash as a param `edi_cache` if there are EDIs in the visualization.


src/modules/dataTypes.js
```diff
@@ -6,6 +6,8 @@ export const DIMENSION_TYPE_DATA_ELEMENT = 'DATA_ELEMENT'
 export const DIMENSION_TYPE_DATA_SET = 'DATA_SET'
 export const DIMENSION_TYPE_EVENT_DATA_ITEM = 'EVENT_DATA_ITEM'
 export const DIMENSION_TYPE_PROGRAM_INDICATOR = 'PROGRAM_INDICATOR'
+export const DIMENSION_TYPE_EXPRESSION_DIMENSION_ITEM =
```
martinkrulltott: Im a bit surprised that GH doesn't pick up that this change already exists in my branch (https://github.com/dhis2/analytics/pull/1370/files?show-viewed-files=true&file-filters%5B%5D=#diff-64e11558ada84b560c86145d9ac2a5eb6c83a7f24fc40ef088b7f40aa2cbec48R21) 🤔

martinkrulltott: image
It's on a different line though (I put it last), should be easy enough to detect if we end up with duplicates though

src/modules/hash.js
```diff
@@ -0,0 +1,35 @@
+import Crypto from 'crypto-js'
+import { DIMENSION_TYPE_EXPRESSION_DIMENSION_ITEM } from './dataTypes.js'
+import { layoutGetAllItems } from './layout/layoutGetAllItems.js'
+
+export const hash = (value) => Crypto.SHA1(value).toString()
+
+const testValue = (value) => typeof value === 'string' && value.length
```
martinkrulltott:
```suggestion
const isValidString = (value) => typeof value === 'string' && value.length
```
I'd probably rename this to something like `isValidString` or sth descriptive, to avoid confusion with "test" (like Cypress or Jest tests) and to indicate that it returns a boolean (which normally starts with "is")

src/modules/dataTypes.js
janhenrikoverland: Moved it to the same line, but still not picked up.

src/modules/hash.js
janhenrikoverland: Changed to `isValidValue`

martinkrulltott (approved):

PR #1440, opened 2023-03-02
feat: icon in SV visualization DHIS2-10496

edoardo:
Implements [DHIS2-10496](https://dhis2.atlassian.net/browse/DHIS2-10496)

**Relates to https://github.com/dhis2/data-visualizer-app/pull/2236**

---

### Key features

1. embed SVG icon on left of SV value

---

### Description

An icon can be assigned to a DE/indicator in Maintenance app.
In DV app there is a new option for Single Value visualizations for showing this icon on the side of the value.

Currently the SVG icon is embedded via a `image` tag in the main SV SVG.

---

### TODO

-   [x] when in dashboard, disable font styles on title/subtitle
-   [x] figure out how to change the icon color to match the value color

---

### Known issues

-   [ ] if a subtitle is very long, it gets cut off on the sides
-   [x] is not possible to change the icon color according to the legend applied: the main reason is that the SVG icons have a specific color for the fill of the various SVG elements. It's possible to change the color via CSS, but the icons need  to be modified to use `currentColor` instead of a specific color in their fill, stroke attributes. **Solved by replacing the hardcoded color with `currentColor` after fetching the SVG from the API.**

---

### Screenshots

Example in DV:
Screenshot 2023-03-02 at 10 53 19

With contrast color applied:
Screenshot 2023-04-14 at 16 31 29

Example in dashboard:
Screenshot 2023-03-02 at 10 30 10

Screencast of the toggling of the icon in DV:
https://user-images.githubusercontent.com/150978/222401093-f64acb37-70e2-41d4-ae4b-5b9f4550e8d5.mov

[DHIS2-10496]: https://dhis2.atlassian.net/browse/DHIS2-10496?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/visualizations/config/adapters/dhis_dhis/index.js
+++ b/src/visualizations/config/adapters/dhis_dhis/index.js
@@ -34,5 +34,13 @@ export default function ({ store, layout, extraOptions }) {
         config.subText = indicatorType?.displayName
     }
 
+    // DHIS2-10496: show icon on the side of the value if an icon is assigned in Maintenance app and the option is set in DV options
+    const showIcon = Boolean(layout.icons?.length)
+    const icon = metaData.items[metaData.dimensions.dx[0]].style?.icon
+
+    if (showIcon && icon) {
+        config.icon = icon
+    }
+
     return config
 }
--- a/src/visualizations/config/adapters/dhis_dhis/index.js
+++ b/src/visualizations/config/adapters/dhis_dhis/index.js
@@ -34,13 +34,5 @@ export default function ({ store, layout, extraOptions }) {
         config.subText = indicatorType?.displayName
     }
 
-    // DHIS2-10496: show icon on the side of the value if an icon is assigned in Maintenance app and the option is set in DV options
-    const showIcon = Boolean(layout.icons?.length)
-    const icon = metaData.items[metaData.dimensions.dx[0]].style?.icon
-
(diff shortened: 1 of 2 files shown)


martinkrulltott (commented):
Looks good so far, but some of the todos needs to be addressed before merging

src/visualizations/config/adapters/dhis_dhis/index.js
```diff
@@ -34,5 +34,13 @@ export default function ({ store, layout, extraOptions }) {
         config.subText = indicatorType?.displayName
     }
 
+    // DHIS2-10496: show icon on the side of the value if an icon is assigned in Maintenance app and the option is set in DV options
+    const showIcon = Boolean(layout.icons?.length)
+    const icon = metaData.items[metaData.dimensions.dx[0]].style?.icon
```
martinkrulltott: FWIW, this will crash if there's no `dx[0]` item or no match in `items[ ]`...

edoardo: This part has been moved to the DV plugin component.

martinkrulltott (approved):

PR #1441, opened 2023-03-02
fix: operands should look and act like data elements

jenniferarnesen:
- when an operand is added to the formula, it is now given the same EXPRESSION_TYPE_DATA as data elements. This way it gets the same styling and is given the same markers in the expression `#{xyz.pdq}`

The only real change is in DraggableTransferOption where the type is set to EXPRESSION_TYPE_DATA instead of using the original backend types of DATA_ELEMENT and DATA_ELEMENT_OPERAND when they are added (either by dropping or doubleclicking)

The rest of the changes are just renaming to EXPRESSION_TYPE_DATA (to encompass both element and element operand)

- added a story for testing
image

### DIFF
--- a/src/components/DataDimension/DraggableTransferOption.js
+++ b/src/components/DataDimension/DraggableTransferOption.js
@@ -3,6 +3,7 @@ import { CSS } from '@dnd-kit/utilities'
 import PropTypes from 'prop-types'
 import React from 'react'
 import { getIcon, getTooltipText } from '../../modules/dimensionListItem.js'
+import { EXPRESSION_TYPE_DATA } from '../../modules/expressions.js'
 import { TransferOption } from '../TransferOption.js'
 import styles from './styles/DraggableTransferOption.style.js'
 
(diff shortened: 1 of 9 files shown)


src/components/DataDimension/DraggableTransferOption.js
```diff
@@ -13,9 +14,10 @@ const DraggableTransferOption = ({
     disabled,
     onDoubleClick,
 }) => {
+    const data = { label, value, type: EXPRESSION_TYPE_DATA }
     const { attributes, listeners, setNodeRef, transform } = useSortable({
         id: value,
-        data: { value, label, type, index: -1 },
```
martinkrulltott, line 18: why no more `index: -1`?

jenniferarnesen: I couldn't any situation where it is actually used.

martinkrulltott (approved):

PR #1452, opened 2023-03-15
fix: ensure the base passed to URL is absolute

edoardo:
Implements [DHIS2-14789](https://dhis2.atlassian.net/browse/DHIS2-14789)

---

### Key features

1. fix for the case of baseURL being `..`

---

### Description

This should account for the case where in production `baseUrl` returned by the `useConfig` hook is `..`.

[DHIS2-14789]: https://dhis2.atlassian.net/browse/DHIS2-14789?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/FileMenu/GetLinkDialog.js
+++ b/src/components/FileMenu/GetLinkDialog.js
@@ -17,7 +17,9 @@ export const GetLinkDialog = ({ type, id, onClose }) => {
     const { baseUrl } = useConfig()
 
     // TODO simply use href from the visualization object?
-    const appUrl = new URL(appPathFor(type, id), baseUrl)
+    const appBaseUrl = new URL(baseUrl, self.location.href)
+
+    const appUrl = new URL(appPathFor(type, id), appBaseUrl)
 
     return (
         <Modal onClose={onClose}>


src/components/FileMenu/GetLinkDialog.js
```diff
@@ -17,7 +17,9 @@ export const GetLinkDialog = ({ type, id, onClose }) => {
     const { baseUrl } = useConfig()
 
     // TODO simply use href from the visualization object?
-    const appUrl = new URL(appPathFor(type, id), baseUrl)
+    const appBaseUrl = new URL(baseUrl, self.location.href)
```
jenniferarnesen, line 20: What about combining these into a single function with args (baseUrl, self.location.href) and put it in utils, and add a few jest tests for it?

jenniferarnesen (approved):

src/components/FileMenu/GetLinkDialog.js
edoardo: The TODO line above is still there because I think all of this can be simplified by just returning the `href` prop in the AO.
I think that should work with different instances, basically the origin in the URL should reflect the instance in use.
For that tho we need to make sure all apps fetch the `href` prop.

PR #1456, opened 2023-04-28
fix: address various SV SVG issues after KFMT

edoardo:
### Key features

1. fix: https://dhis2.atlassian.net/browse/DHIS2-15233
2. fix: https://dhis2.atlassian.net/browse/DHIS2-15234
3. fix: https://dhis2.atlassian.net/browse/DHIS2-15235
4. fix: https://dhis2.atlassian.net/browse/DHIS2-15236
5. fix: https://dhis2.atlassian.net/browse/DHIS2-15237

---

### Description

This PR addresses the issues listed in the subtasks above.

Mostly about text color that should consistently use a contrast color when a background color is applied for title, subtitle and indicator type text under the value.
Also, positioning of title/subtitle when using custom text size, to avoid text to overlap or be clipped.

### DIFF
--- a/src/visualizations/config/generators/dhis/singleValue.js
+++ b/src/visualizations/config/generators/dhis/singleValue.js
@@ -156,7 +156,9 @@ const generateDashboardItem = (
     const container = document.createElement('div')
     container.setAttribute(
         'style',
-        `display: flex; flex-direction: column; align-items: center; justify-content: center; width: 100%; height: 100%; background-color:${backgroundColor};`
+        `display: flex; flex-direction: column; align-items: center; justify-content: center; width: 100%; height: 100%; ${
+            backgroundColor ? `background-color:${backgroundColor};` : ''
+        }`
     )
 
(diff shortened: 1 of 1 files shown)


src/visualizations/config/generators/dhis/singleValue.js
```diff
@@ -350,6 +359,7 @@ const generateDVItem = (
             formattedValue: config.formattedValue,
             subText: config.subText,
             valueColor,
+            textColor: titleColor,
```
martinkrulltott, line 362: I'm curious if this will set the subtext color to e.g. green if the _title_ is set to green?

martinkrulltott (approved):

src/visualizations/config/generators/dhis/singleValue.js
edoardo: `titleColor` at this point can only be the contrast color or white.
The custom title color is applied directly on the title element and not passed via `titleColor`.

PR #1478, opened 2023-05-30
feat: toolbar UI update with hoverable menu

HendrikThePendric:
Implements [DHIS2-15167](https://dhis2.atlassian.net/browse/DHIS2-15167)

**Relates to https://github.com/dhis2/line-listing-app/pull/368 & https://github.com/dhis2/data-visualizer-app/pull/2358**

---

### Key features

1. Adds `Toolbar` and `ToolbarSidebar` components to build the toolbar layout
2. Adds `HoverMenubar` and related components to build hoverable menu within the toolbar
3. Adds `UpdateButton` and `InterpretationsAndDetailsToggler` to be placed in the toolbar
4. Updates `FileMenu` to be a hoverable dropdown menu, leveraging the new hover-menu-bar components.
5. Adds `initiallyActiveTabKey` to `VisualizationOptions` component so it can be opened at a specified tab.

---

### BREAKING CHANGE

The updated `FileMenu` is not compatible with the previous version. Apps will be required to update their toolbar before using this version of analytics.

---

### TODO

-   [x] Add tests
-   [x] Add demo

---

### Screenshots

Screenshot 2023-06-12 at 18 20 44

_Not the most glamorous screenshot, but it works._

[DHIS2-15167]: https://dhis2.atlassian.net/browse/DHIS2-15167?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/FileMenu/FileMenu.js
HendrikThePendric, line 1: The diff for this file looks more impressive than the changes I actually made. The component now make use of the new hover menu functionality, so it doesn't need to have an internal open-state, render a layer etc. Due to the removal of the Layer/Popper/etc the JSX nesting changed significantly and this is causing the diff to look big.

src/components/FileMenu/__tests__/FileMenu.spec.js
HendrikThePendric, line 1: I've switched to using `@testing-library/react` because I was running into issues when opening poppers with enzyme. I also prefer this lib to Enzyme in general, but that's not why I started using it here.

src/__demo__/Toolbar.stories.js
```diff
@@ -0,0 +1,78 @@
...
+            </HoverMenuBar>
+            <InterpretationsAndDetailsToggler
+                isShowing={isSidebarShowing}
+                onClick={() => setIsSidebarShowing((current) => !current)}
+            />
+        </Toolbar>
+    )
+}
+
+storiesOf('HoverMenuBar', module).add('default', () => {
```
edoardo: Should this story be called `Toolbar`?

edoardo (approved):

src/__demo__/Toolbar.stories.js
HendrikThePendric: Yeah, you are correct actually. I think this is a remnant of how I started off. Most of the complexity/functionality is in this `HoverMenuBar` so I focussed purely on that to begin with. Later on I decided to also included the other `Toolbar` components too, but I never updated the naming. 

I'll fix this.

PR #1510, opened 2023-06-20
feat: update interpretations component so it can be used in dashboard items

jenniferarnesen:
Implements [DHIS2-15441](https://dhis2.atlassian.net/browse/DHIS2-15441)

**Required for https://github.com/dhis2/dashboard-app/pull/2430**

Changes to the interpretation component so that it can be used in the dashboard app. The main difference between interpretations in dashboard vs DV/LL is that replies are viewed "inline" rather than through opening a modal. In order to add this functionality, the `appUrl` property is sent from the dashboard app, so that a "Launch Url" button can be added to the interpretation button strip. The button opens the relevant app in interpretation view mode. 

The component will continue to work as before for DV and LL so this is safe as a "feat" release.

The `appUrl` property is kind of doing double duty
1. provide the url for the LaunchURL link in the interpretation button strip (right side red arrow)
image
2.  tweak the interface for dashboard
    1. hide the `See interpretation` button (screenshot from DV)
image
    2. add the 👁️ buttonbar button (left side red arrow as shown above) for viewing the interpretation (in place of the `See interpretation` big button)
image

[DHIS2-15441]: https://dhis2.atlassian.net/browse/DHIS2-15441?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

TODO
- retest after fixing the conflicts, that all the access checks were fixed.


HendrikThePendric (approved):
As discussed on Slack, it would be good to add some comments and/or rename some variable in `src/components/Interpretations/common/Interpretation/Interpretation.js` for clarification. But the underlying logic is sound and the implementation is clean. So I'm approving.

PR #1526, opened 2023-06-28
fix(OpenFileDialog): use table toolbars

cooper-joe:
### Key features

1. Place filtering and pagination in `DataTableToolbar` wrapper components.
2. Bump `UI` version to get input size fixes.

---

### Screenshots

_Before:_

image

_After:_

image


edoardo (dismissed):
LGTM

jenniferarnesen:
@cooper-joe I tried out the changes and found that the modal does a brief height change when paging, which results in a jumpy effect.

Before:
[

https://github.com/user-attachments/assets/daca38f9-4930-4f46-a23c-f807e9abe8bb

](url)

After

https://github.com/user-attachments/assets/9876f4c7-cac8-4191-b973-01d2f5b438f7

jenniferarnesen:
@cooper-joe can you take another look at this one?

cooper-joe:
@jenniferarnesen I don't think the jumping modal height is related to these changes. It seems to happen for only the first time a page is changed (e.g. the first 1->2 and the first 2->3). Screen recording from current master branch with the same issue:

https://github.com/user-attachments/assets/035840ab-2af2-4ce8-bed3-acb1018e5f8c

PR #1531, opened 2023-06-29
fix: correct access checks for showing intp reply, edit and delete buttons (DHIS2-15510)

jenniferarnesen:
Implements [DHIS2-15510](https://dhis2.atlassian.net/browse/DHIS2-15510)

("comment" and "reply" mean the same thing.)

These are the access checks that are done in the backend, according to @vietnguyen
* edit or delete an interpretation: `isCreatorOrSuperuser of interpretation`
* add interpretation comment: `interpretation.access.write`
* update comment: `isCreatorOrSuperuser of comment`
* delete comment: `isCreatorOrSuperuser of comment && interpretation.access.write`

Other changes:
- use `createdBy` instead of `user`. According to Viet, createdBy is the correct property and `user` is only kept for backwards compatibility
- don't fetch `comment.access` because the values can be wrong, and it's not needed anyway
- added error handling to the CommentDeleteButton
- error message for edit-comment-failure is made generic. I found that the backend would report a misleading message such as "interpretation does not exist" when it does in fact exist, and user can see it, but user doesn't have access to edit.

> Screenshots 

>> Edit or delete interpretation when user doesn't have the correct access:
before (Edit fails with message and Delete fails silently):
image
after (Edit and Delete buttons are hidden):
image

>> Add interpretation comment, and edit/delete comment when user doesn't have the correct access:
Before. Input field is shown and edit/delete comment buttons are shown, andPost reply fails silently:
image
After. Input field is not shown, and edit/delete comment buttons are not shown:
image

Hover tooltip for reply button
In thread, no reply access:
image
In main interpretations list, no reply access
image
User has reply access
image

Provide feedback if delete fails
image

[DHIS2-15510]: https://dhis2.atlassian.net/browse/DHIS2-15510?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cooper-joe (approved):

edoardo (approved):

jenniferarnesen:
@cooper-joe Edoardo had a good observation that by hiding the Reply icon, the user can no longer see if there are replies. Would it be better to still show the icon but as disabled?

cooper-joe:
> @cooper-joe Edoardo had a good observation that by hiding the Reply icon, the user can no longer see if there are replies. Would it be better to still show the icon but as disabled?

Good point. I think still show the Reply icon, enabled. If the user clicks it, it still navigates to the list of replies, just without the input.

edoardo (approved):

HendrikThePendric (approved):

src/components/Interpretations/InterpretationModal/InterpretationThread.js
```diff
@@ -52,13 +62,13 @@ const InterpretationThread = ({
                                 currentUser={currentUser}
                                 interpretationId={interpretation.id}
                                 onThreadUpdated={onThreadUpdated}
-                                interpretationWriteAccess={
-                                    interpretation.access.write
+                                hasInterpretationReplyAccess={
+                                    interpretationAccess.reply
```
edoardo: Should we use `reply` instead of `comment` everywhere?

edoardo (approved):
Looks good, much more clear with the `get*Access` functions.
One observation: should we decide on `comment` vs. `reply` and stick to one term everywhere? Reply is what is shown in the UI so perhaps go with `reply`?

HendrikThePendric (approved):
I agree with the comment @edoardo left reg. the naming consistency. That would be a good improvement to make. Other than that this looks great. I think having the     `getInterpretationAccess` and `getCommentAccess` functions in place with corresponding unit tests will make this very robust. 👍

src/components/Interpretations/InterpretationModal/InterpretationThread.js
jenniferarnesen: ah, you mean since the component is `Comment`

jenniferarnesen: Done

PR #1552, opened 2023-07-12
fix: avoid undefined in totals (DHIS2-14511)

edoardo:
Implements [DHIS2-14511](https://dhis2.atlassian.net/browse/DHIS2-14511)

---

### Key features

1. fix `undefined` in totals when value is 0 

---

### Description

The issue is well described in the ticket.
The problem in the code is that when the totals are computed, the value is not initialised to 0 when the input value is `undefined`.

---

### Screenshots

Before:
Screenshot 2023-07-12 at 11 00 28

After:
Screenshot 2023-07-12 at 11 00 17

[DHIS2-14511]: https://dhis2.atlassian.net/browse/DHIS2-14511?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/modules/pivotTable/PivotTableEngine.js
+++ b/src/modules/pivotTable/PivotTableEngine.js
@@ -9,6 +9,7 @@ import { DIMENSION_ID_ORGUNIT } from '../predefinedDimensions.js'
 import { renderValue } from '../renderValue.js'
 import { VALUE_TYPE_NUMBER, VALUE_TYPE_TEXT } from '../valueTypes.js'
 import { AdaptiveClippingController } from './AdaptiveClippingController.js'
+import { addToTotalIfNumber } from './addToTotalIfNumber.js'
 import { parseValue } from './parseValue.js'
 import {
     AGGREGATE_TYPE_NA,
@@ -35,7 +36,6 @@ import {
     NUMBER_TYPE_ROW_PERCENTAGE,
     NUMBER_TYPE_VALUE,
 } from './pivotTableConstants.js'
-import { processTotalValue } from './processTotalValue.js'
 
 const dataFields = [
     'value',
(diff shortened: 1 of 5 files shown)


src/modules/pivotTable/PivotTableEngine.js
```diff
@@ -705,7 +705,7 @@ export class PivotTableEngine {
                 dataFields.forEach((field) => {
                     const headerIndex = this.dimensionLookup.dataHeaders[field]
                     const value = parseValue(dataRow[headerIndex])
-                    if (value && !isNaN(value)) {
+                    if (!isNaN(value)) {
```
martinkrulltott: extract this check to a fn that can be tested by a unit test? (instead of repeating it 7 times)

martinkrulltott (commented):
unit tests would be good, as these checks aren't that human readable

edoardo:
> unit tests would be good, as these checks aren't that human readable

Added unit tests for the new helper function.

martinkrulltott (changes requested):
I would strongly suggest switching `processTotalValue` from an impure to a pure function, unless there are specific reasons not to

edoardo:
> I would strongly suggest switching `processTotalValue` from an impure to a pure function, unless there are specific reasons not to

Refactored.

src/modules/pivotTable/__tests__/addToTotalIfNumber.js
```diff
@@ -0,0 +1,99 @@
...
+        testName: 'zero value with existing total',
+        value: 0,
+        totalObj: { [field]: 100 },
+        field,
+        expected: 100,
+    },
+    {
+        testName: 'positive value with existing total',
+        value: 1,
+        totalObj: { [field]: 100 },
```
martinkrulltott: is it necessary to use `totalObj` for the tests? Or could they just pass in e.g. `testName: 'foo', value: 1, total: 100, expected: 101`?

martinkrulltott (approved):

PR #1579, opened 2023-09-13
feat(pivot-table): truncate title and show full in tooltip (DHIS2-14827)

HendrikThePendric:
Implements [DHIS2-14827](https://dhis2.atlassian.net/browse/DHIS2-14827)

---

### Key features

1. Because we now truncate the title, long filters won't stretch the table horizontally anymore
2. Once the title gets truncated a tooltip will be available for users to see the full title

---

### Description

Before adding the tooltip functionality, I also addressed a small bug: the `max-width` for `div.title-cell-content` didn't take into account the cell padding and border and as such the value was a bit too high.

The tooltip is only added if the title text is overflowing the container.

I also added a story to demo the feature.

---

### Considerations

-   The title cell has 5px padding in all directions. As a result, `div.title-cell-content`, which is the tooltip anchor element, has 5px of whitespace around it. So the tooltip only shows when hovering the actual text, not the entire table cell. I'm not sure if this is a bug or a feature. I _think_ changing this behaviour should be relatively simple, but just not sure whether we want to.

---

### TODO

-   [ ] Upgrade DV to the to-be-released version of @dhis2/analytics so it can benefit form this feature.
-   [x] Decide if we want to keep the padding around the tooltip anchor, and make necessary changes if we don't. WE DECIDED TO FILL OUT THE CELL (see [2a47d1a](https://github.com/dhis2/analytics/pull/1579/commits/2a47d1ac9459641058af62d9082fbb8898032002))

---

### Screenshots

Screenshot 2023-09-13 at 09 48 44

[DHIS2-14827]: https://dhis2.atlassian.net/browse/DHIS2-14827?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/PivotTable/PivotTableTitleRow.js
+++ b/src/components/PivotTable/PivotTableTitleRow.js
@@ -1,5 +1,6 @@
+import { Tooltip } from '@dhis2/ui'
 import PropTypes from 'prop-types'
-import React, { useState, useEffect } from 'react'
+import React, { useRef, useMemo } from 'react'
 import { PivotTableCell } from './PivotTableCell.js'
 import { usePivotTableEngine } from './PivotTableEngineContext.js'
 import { cell as cellStyle } from './styles/PivotTable.style.js'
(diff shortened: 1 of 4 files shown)


martinkrulltott (approved):

martinkrulltott (approved):

src/components/PivotTable/PivotTableTitleRow.js
```diff
@@ -10,32 +11,62 @@ export const PivotTableTitleRow = ({
...
     totalWidth,
 }) => {
+    const containerRef = useRef(null)
+    const initialScrollWidthRef = useRef(null)
+    const handleContainerMount = useCallback((node) => {
+        if (node) {
+            containerRef.current = node
+            initialScrollWidthRef.current = node.scrollWidth
+        }
+    }, [])
```
HendrikThePendric: My initial implementation in commit https://github.com/dhis2/analytics/pull/1579/commits/b020edd88891b42407a0463eaa90629189c8c335 had a problem:
- To determine if the title is truncated I was checking if the container node's _current_ `scrollWidth` was greater than its current `clientWidth`.
- But in DV, the container width changes with a CSS transformation/animation, and during this animation the browser would sometimes report both values as equal even though the text doesn't fit
- The result was that sometimes after hiding/showing the details panel (which causes the canvas to shrink) the tooltip would not show, even though it should have.

So my solution revolves around the notion that (due to `white-space: nowrap` and  `overflow: hidden`) the initial `scrollWidth` of the element is correct and will remain constant. I think it all makes sense so far....

But the current implementation is not the easiest to follow and perhaps a bit quirky. I landed on it for the following reasons:
1. To measure a DOM node it [is recommended](https://legacy.reactjs.org/docs/hooks-faq.html#how-can-i-measure-a-dom-node) to use a _callback ref_, which `handleContainerMount` is.
2. But we also need to read some attributes from the container element before the JSX is rendered, so we need a regular ref to it as well. But since we already attached a callback ref, we had to manually set that regular ref in the function body of the callback ref. 
3. And the same applies to the `initialScrollWidthRef`: we can set this value when the element mounts and trust it to remain constant.
4. Since both these things are effectively "passive" values, I did not want to store them in a state variable and possibly triggering additional renders. So that's why both are stored in a `ref`

martinkrulltott (approved):

PR #1586, opened 2023-10-05
fix: handle both Set and array for currentUser.authorities when checking interpretations access [DHIS2-15964]

jenniferarnesen:
Fixes https://dhis2.atlassian.net/browse/DHIS2-15964

In the olden days when we used d2.currentUser, `authorities` was a Set.
With the migration to the app-runtime, requesting currentUser or /me with the dataEngine, the `authorities` will be an array.

Choosing here to handle both instead of requiring array (and thus a major upgrade) since we have no documentation or strict types on our functions/components.

So apps that get the currentUser from d2 (like DV) and apps that get currentUser from dataEngine(like LL) will both work.

Maybe support for Set can be removed in a future major release.

### DIFF
--- a/src/components/Interpretations/common/getInterpretationAccess.js
+++ b/src/components/Interpretations/common/getInterpretationAccess.js
@@ -1,12 +1,15 @@
 // For backwards compatibility
 // accept both Set (from the old d2.currentUser object) and array
-const hasAuthority = (authorities = [], authority) => {
-    if (Array.isArray(authorities)) {
+const hasAuthority = (authorities, authority) => {
+    if (!authority || typeof authority !== 'string) {
+        throw new Error(`"hasAuthority" requires "authority" to be a populated string but received ${authority}`)
+    } else if (Array.isArray(authorities)) {
         return authorities.includes(authority)
+    } else if (typeof authorities.has === 'function') {
+        return authorities.has(authority)
+    } else {
+        throw new Error(`"hasAuthority" requires "authorities" to be an array or set of authorities (strings)`)
     }
-    return typeof authorities.has === 'function'
-        ? authorities.has(authority)
-        : false
 }
 
 const isSuperuser = (authorities) => hasAuthority(authorities, 'ALL')
--- a/src/components/Interpretations/common/getInterpretationAccess.js
+++ b/src/components/Interpretations/common/getInterpretationAccess.js
@@ -1,15 +1,20 @@
 // For backwards compatibility
 // accept both Set (from the old d2.currentUser object) and array
-const hasAuthority = (authorities, authority) => {
-    if (!authority || typeof authority !== 'string) {
-        throw new Error(`"hasAuthority" requires "authority" to be a populated string but received ${authority}`)
-    } else if (Array.isArray(authorities)) {
-        return authorities.includes(authority)
-    } else if (typeof authorities.has === 'function') {
-        return authorities.has(authority)
-    } else {
-        throw new Error(`"hasAuthority" requires "authorities" to be an array or set of authorities (strings)`)
+export const hasAuthority = (authorities, authority) => {
+    if (!authority || typeof authority !== 'string') {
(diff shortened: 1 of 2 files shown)


src/components/Interpretations/common/getInterpretationAccess.js
```diff
@@ -1,6 +1,17 @@
+import { isArray } from 'lodash'
```
HendrikThePendric: We could also just use the built-in `Array.isArray` method.

src/components/Interpretations/common/getInterpretationAccess.js
```diff
@@ -1,6 +1,17 @@
...
+
+// For backwards compatibility
+// accept both Set (from the old d2.currentUser object) and array
+const isSuperuser = (authorities = []) => {
+    if (isArray(authorities)) {
+        return authorities.includes('ALL')
+    }
+    return authorities.has('ALL')
+}
```
HendrikThePendric: It might be nice to have something like this, so we can easily add checks for other roles:
```suggestion
const hasAuthority = (authorities = [], authority) =>
    Array.isArray(authorities)
        ? authorities.includes(authority)
        : authorities.has(authority)

const isSuperuser = (authorities) => hasAuthority(authorities, 'ALL')
```
But not at all required

HendrikThePendric (approved):
LGTM 👍

src/components/Interpretations/common/getInterpretationAccess.js
```diff
@@ -1,9 +1,23 @@
...
-    currentUser?.authorities.has('ALL')
+// For backwards compatibility
+// accept both Set (from the old d2.currentUser object) and array
+const hasAuthority = (authorities = [], authority) => {
+    if (Array.isArray(authorities)) {
+        return authorities.includes(authority)
+    }
+    return typeof authorities.has === 'function'
+        ? authorities.has(authority)
+        : false
```
HendrikThePendric: I'm not sure that returning `false` here is a good idea. This function should receive either an array or a set of authorities. If it doesn't receive that, the function was called in the wrong way and it can't compute the authorities, so should throw an error. Firstly `false` is actually wrong: the user may or may not have the right authority, but we are unable to asses this. Secondly, this will make it harder to debug things, since this way of handling things obscures things that have gone wrong at an earlier stage (i.e. user authorities not being available when the function was called). I think [failing-fast](https://www.martinfowler.com/ieeeSoftware/failFast.pdf) will improve maintainability and debug-ability in the long run.

HendrikThePendric (changes requested):

src/components/Interpretations/common/getInterpretationAccess.js
```diff
@@ -1,9 +1,23 @@
...
+}
```
HendrikThePendric: Actually, taking my own advice, we should probably not have default values either:

```suggestion
const hasAuthority = (authorities, authority) => {
    if (!authority || typeof authority !== 'string) {
        throw new Error(`"hasAuthority" requires "authority" to be a populated string but received ${authority}`)
    } else if (Array.isArray(authorities)) {
        return authorities.includes(authority)
    } else if (typeof authorities.has === 'function') {
        return authorities.has(authority)
    } else {
        throw new Error(`"hasAuthority" requires "authorities" to be an array or set of authorities (strings)`)
    }
}
```

HendrikThePendric (changes requested):

HendrikThePendric (approved):
I approve now after pushing some changes. But I guess another team member should also approve (and merge).

HendrikThePendric (approved):

jenniferarnesen:
Changes by @HendrikThePendric LGTM. The function should throw an error if the needed arguments are not provided by the app.

martinkrulltott:
Superseded by the combo PR https://github.com/dhis2/analytics/pull/1588

PR #1590, opened 2023-10-24
feat: [DHIS2] Generate library documentation with Gen AI

eirikhaugstulen:
Hey everyone,

@jenniferarnesen and I talked briefly about how you could use generative AI to write documentation for you public libraries.
I did something similar on another project, so I'll post a how-to-guide here and you can decide if you want to do something similar. 

## TLDR
This is the resulting documentation created by AI

image

I used an editor called _Cursor_ to write documents. Cursor is basically just a vscode fork where they try to implement GPT 4. You can read more at [their website](https://cursor.sh), but the main gist is that they embed your repository and provide contextual related code snippets to OpenAI. This gives the AI context of your code and surrounding components, increasing the relevance of your AI completions.

## Here's how to create docs
1. Create a new markdown file where you want the docs to be
2. Click `Cmd + K`
3. Reference the Prompt
4. Reference the Component you want documented

https://github.com/dhis2/analytics/assets/12842868/193a4c0d-f9b8-493a-be1a-596d33d6e394

## How to improve your response
Everything written in `Prompts/generate-documentation.md` will be treated as rules for the AI. This means, if you need it to abide by some sort of guidelines, write it in this file and the AI will follow.

### There are some limitations to this though;
- Since you're only using propTypes, the props-completion might not be that great. In some cases, the AI could understand object attributes and function parameters, but most of the times it would just use `object` or `func`. (The same thing as defined in propTypes)


jenniferarnesen:
Not going to implement this right now

PR #1594, opened 2023-11-17
feat: add support for trackedEntities analytics requests (DHIS2-16023)

edoardo:
Implements [DHIS2-16023](https://dhis2.atlassian.net/browse/DHIS2-16023)

**Relates to https://github.com/dhis2/line-listing-app/pull/451/**

---

### Key features

1. implement `Analytics` class for tracked entities request
2. add support for the new `trackedEntity/query` endpoint
3. adjust dimension id formatting

---

### Description

The PR adds support for the new analytics endpoint for tracked entities.
The `outputType` selected in LL needs to be taken into account when composing the request URL's path and when formatting the dimension ids.

[DHIS2-16023]: https://dhis2.atlassian.net/browse/DHIS2-16023?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@dhis2/analytics",
-    "version": "26.3.0",
+    "version": "999.9.9-TE-alpha.3",
```
HendrikThePendric: Should this change be undone?

src/api/analytics/utils.js
```diff
@@ -7,3 +7,21 @@ const whitelistRegExp = new RegExp(`(?:${whitelistURICodes.join('|')})`, 'g')
...
+    [path, program, trackedEntityType].filter(Boolean).join('/')
+
+export const formatDimension = ({
+    outputType,
+    programId,
+    programStageId,
+    dimension,
+}) =>
+    [
+        // XXX it would be clearer to have this consistent with what is sent in the request as for EVENT/ENROLLMENT
```
HendrikThePendric: For an uninitiated reader like myself, this comment is missing some context. From the text that is there, I gather that the current implementation is not ideal, and that you would have liked to have implemented things the same as EVENT/ENROLLMENT. I am guessing that the ternary operator below is an additional complexity that is preventing you doing so. But what is missing is a brief explanation of _why_ this additional complexity is present.

HendrikThePendric (approved):
Code is looking very clear. I don't think I am fully qualified to assess whether or not the changes all make sense, but it certainly looks like a sensible implementation.

I've left 2 comments. Perhaps the one about the version number could be quite vital.

package.json
martinkrulltott: This should indeed be undone yeah, we'll do that just before releasing.

src/api/analytics/utils.js
edoardo: This actually had to do with the `outputType` value.
Initially `TRACKED_ENTITY` was used, which didn't match with the value used in the api.
This [has been updated](https://github.com/dhis2/analytics/commit/a7f9396ab21553e94448ef40519866b1d15db4b3) tho, so this comment is not relevant anymore.

PR #161, opened 2019-11-27
fix: rules structure

martinkrulltott:

PR #1598, opened 2023-12-20
feat: add support for Outliers table vis type in DV (DHIS2-13858)

edoardo:
Part of the implementation for [DHIS2-13858](https://dhis2.atlassian.net/browse/DHIS2-13858)

**Relates to https://github.com/dhis2/data-visualizer-app/pull/2942**

---

### Key features

1. add support for the `outlierDetection` endpoint for aggregate data
2. add support for the Outlier table visualization type and layout
3. cleanup of unused code and updated unit tests

---

### Description

This analytics API endpoint is required for the Outliers table visualization type in DV.

---

### TODO

-   [x] update icons for Outlier table visualization type (used in the OpenFileDialog)

[DHIS2-13858]: https://dhis2.atlassian.net/browse/DHIS2-13858?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


package.json
```diff
@@ -75,6 +75,9 @@
         "react-beautiful-dnd": "^10.1.1",
         "resize-observer-polyfill": "^1.5.1"
     },
+    "resolutions": {
```
martinkrulltott, line 78: Is this a temporary (remove before merge) or permanent solution?

edoardo: > Is this a temporary (remove before merge) or permanent solution?

I guess it's "temporary" until `cli-app-scripts` gets its version bumped to a major 9...

src/components/DataDimension/DataDimension.js
```diff
@@ -1,21 +1,46 @@
...
-    const supportsEDI =
-        `${serverVersion.major}.${serverVersion.minor}.${
-            serverVersion.patch || 0
-        }` >= '2.40.0'
+
+    const filterDataTypesByVersion = useCallback(
+        (dataTypes) =>
+            dataTypes.filter(({ id }) =>
+                // Calculations only available from 2.40
+                id === DIMENSION_TYPE_EXPRESSION_DIMENSION_ITEM
```
martinkrulltott: This is technically correct but personally I'd write this as `id !== DIMENSION_TYPE_EXPRESSION_DIMENSION_ITEM || serverVersion.minor >= 40` to avoid having to use `true`

src/components/DataDimension/DataTypeSelector.js
```diff
@@ -2,57 +2,73 @@ import { SingleSelectField, SingleSelectOption } from '@dhis2/ui'
...
-                .filter(
-                    (type) =>
-                        type.id !== DIMENSION_TYPE_EXPRESSION_DIMENSION_ITEM ||
-                        includeCalculations
-                )
-                .map((type) => (
+}) => {
+    const { visType } = useDataDimensionContext()
+
+    return (
```
martinkrulltott, line 19: There are clearly two different flows here, when `dataTypes.length === 1` (new) and when it isn't (old). For clarity, maybe it's better to make the distinction early to avoid having to do the checks in so many places. So I would prefer to refactor this to something like `return dataTypes.length === 1 ? [ SingleSelectField for NEW ] : [ SingleSelectField for OLD ]`

martinkrulltott: if you then want to avoid some duplication, like `label={i18n.t('Data Type')}`, that can be extracted out to a const like `const label ={i18n.t('Data Type')}` to be used in both `SingleSelectFields` instead

src/components/PeriodDimension/PeriodTransfer.js
```diff
@@ -162,35 +177,40 @@ const PeriodTransfer = ({
...
-            selected={selectedPeriods.map((period) => period.id)}
+            selected={selectedItems.map((period) => period.id)}
             leftHeader={renderLeftHeader()}
             enableOrderChange
             height={TRANSFER_HEIGHT}
             optionsWidth={TRANSFER_OPTIONS_WIDTH}
             selectedWidth={TRANSFER_SELECTED_WIDTH}
             selectedEmptyComponent={renderEmptySelection()}
-            rightHeader={renderRightHeader()}
+            rightHeader={<RightHeader infoText={infoBoxMessage} />}
```
martinkrulltott: `infoText` vs `infoBoxMessage`, a bit confusing given that they're both internal / invented by us.. I think we should settle on only using one of them as our naming convention. Given that `infoBoxMessage` is exported externally maybe `infoText` has to go?

src/components/PeriodDimension/PeriodTransfer.js
```diff
@@ -162,35 +177,40 @@ const PeriodTransfer = ({
...
-                    value: id,
-                })
-            )}
+            options={[...allPeriods, ...selectedItems].map(({ id, name }) => ({
+                label: name,
+                value: id,
+            }))}
             renderOption={(props) => (
                 <TransferOption
+                    /* eslint-disable react/prop-types */
```
martinkrulltott: use `// eslint-disable-next-line` if it's only for a single line?

martinkrulltott: or avoid it altogether with this?
```
renderOption={({ value, ...props }) => (
    <TransferOption
        {...props}
        active={isActive(value)}
        icon={PeriodIcon}
        dataTest={`${dataTest}-transfer-option`}
    />
)}
```

src/modules/layoutUiRules/__tests__/rules.spec.js
```diff
@@ -25,6 +30,9 @@ const allArrayItemsAreValid = (allItems, validItems) =>
 const allArrayItemsAreValidAxisIds = (array) =>
     allArrayItemsAreValid(array, ALL_AXIS_IDS)
 
+const allArrayItemsAreValidDimensionIds = (array) =>
+    allArrayItemsAreValid(array, lockableDims)
```
martinkrulltott: This test is technically correct, but the reuse of `lockableDims` isn't great from a scalability perspective, as it's just a coincidence that the arrays needed contain the same three items. They shouldn't be coupled together though.
The original use was for testing lockable dimensions, but the use now in `testKeysAreValidDimensionIds` are for a different context, so it seems better to provide a separate array to not mix up the two contexts if they would diverge in the future.

src/modules/layoutUiRules/rules.js
```diff
@@ -25,12 +26,14 @@ import {
...
     VIS_TYPE_SCATTER,
     VIS_TYPE_LINE_LIST,
+    VIS_TYPE_OUTLIER_TABLE,
 } from '../visTypes.js'
 
 const RULE_PROP_AVAILABLE_AXES = 'availableAxes',
     RULE_PROP_MAX_DIMS_PER_AXIS = 'maxNumberOfDimsPerAxis',
     RULE_PROP_MIN_DIMS_PER_AXIS = 'minNumberOfDimsPerAxis',
     RULE_PROP_MAX_ITEMS_PER_AXIS = 'maxNumberOfItemsPerAxis',
+    RULE_PROP_MAX_ITEMS_PER_DIMENSION = 'maxNumberOfItemsPerDimension',
```
martinkrulltott: All other rules seems to shorten "dimensions" to "dims" so to follow that pattern this should be called `RULE_PROP_MAX_ITEMS_PER_DIM` then?

martinkrulltott:
```suggestion
    RULE_PROP_MAX_ITEMS_PER_DIM = 'maxNumberOfItemsPerDim',
```

src/modules/layoutUiRules/rules.js
```diff
@@ -169,6 +189,9 @@ export const getMaxNumberOfDimsPerAxisByVisType = (visType) =>
 export const getMinNumberOfDimsPerAxisByVisType = (visType) =>
     getRulesByVisType(visType)[RULE_PROP_MIN_DIMS_PER_AXIS] || {}
 
+export const getMaxNumberOfItemsPerDimensionByVisType = (visType) =>
```
martinkrulltott:
```suggestion
export const getMaxNumberOfItemsPerDimByVisType = (visType) =>
```

src/modules/layoutUiRules/index.js
```diff
@@ -6,10 +6,12 @@ export {
     getAxisMinNumberOfDimsByVisType as getAxisMinNumberOfDimensions,
     getAxisPerLockedDimByVisType as getAxisPerLockedDimension,
     getAllLockedDimIdsByVisType as getAllLockedDimensionIds,
+    getDimensionMaxNumberOfItemsByVisType as getDimensionMaxNumberOfItems,
 } from './rulesHelper.js'
 
 export {
     hasAxisTooManyItemsByVisType as hasAxisTooManyItems,
+    hasDimensionTooManyItemsByVisType as hasDimensionTooManyItems,
```
martinkrulltott, line 14: `hasDimensionTooManyItemsByVisType` doesn't sound correct to me. Should be `dimensionHasTooManyItemsByVisType`? I'm not a native english speaker either though so I'll loop in @jenniferarnesen 🙏

martinkrulltott: I see that this might just be following the item above called `hasAxisTooManyItemsByVisType`, but that can't really be used as a reference as it was initially [added by me](https://github.com/dhis2/analytics/pull/161) 4 years ago (and I don't understand what I was thinking back then 🙈 ). So that should probably change too..

jenniferarnesen: I guess it's following the convention from the line above? It's ugly, yeah, but consistent 😬

src/modules/layoutUiRules/rulesUtils.js
```diff
@@ -20,6 +21,21 @@ export const hasAxisTooManyItemsByVisType = (
...
+    visType,
+    dimensionId,
+    numberOfItems
+) => {
+    const maxNumberOfItemsPerDimension = getDimensionMaxNumberOfItemsByVisType(
+        visType,
+        dimensionId
+    )
+
+    return maxNumberOfItemsPerDimension
```
martinkrulltott: Here's another ternary with an explicit boolean return, like the previous one [I commented on here](https://github.com/dhis2/analytics/pull/1598#discussion_r1510953962). Again I would personally avoid returning `true` explicitly, as to me that's a sign that something can be refactored.
In this case the logic looks like this:
<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
</head>

<body link="#0563C1" vlink="#954F72">


maxNumberOfItemsPerDimension | numberOfItems >   maxNumberOfItemsPerDimension | output
-- | -- | --
TRUE | TRUE | TRUE
TRUE | FALSE | FALSE
FALSE | TRUE | FALSE
FALSE | FALSE | FALSE



</body>

</html>

That simply means that it can be written as 
```
return maxNumberOfItemsPerDimension && numberOfItems > maxNumberOfItemsPerDimension
```

The rule just above (`hasAxisTooManyItemsByVisType`) could do with the same change as well

martinkrulltott (commented):
Reviewed with comments, see separate topics above

src/modules/layoutUiRules/index.js
martinkrulltott: Ok I blame 2019-Martin then 😬

src/components/PeriodDimension/PeriodTransfer.js
edoardo: This code was taken from the `ItemSelector`, so the same comments apply there.

src/components/PeriodDimension/PeriodTransfer.js
edoardo: Fixed by separating `value` from `props`.

src/components/PeriodDimension/PeriodTransfer.js
edoardo: Updated in both components.

src/modules/layoutUiRules/index.js
edoardo: Renaming all this functions is not really part of the PR's scope.
Should we do it here or in a separate refactor PR?

src/modules/layoutUiRules/rules.js
edoardo: Renamed.

src/modules/layoutUiRules/rules.js
edoardo: Renamed.

src/components/DataDimension/DataTypeSelector.js
edoardo: I tried this and had issues with re-rendering of components and errors due to the select trying to preselect an item that is not present in the list...

src/modules/layoutUiRules/__tests__/rules.spec.js
edoardo: Added a separate list.

src/components/DataDimension/DataDimension.js
edoardo: Refactored.

src/modules/layoutUiRules/rulesUtils.js
edoardo: Refactored.

src/components/DataDimension/DataTypeSelector.js
martinkrulltott: The selected item must be available to select, yes. But that sounds like you need to change what goes in to the `selected` prop, maybe add a condition there to make sure it's passed properly?

martinkrulltott (approved):

src/components/DataDimension/DataTypeSelector.js
edoardo: I tried again, I must have done something wrong the previous time, not it seems to work fine.

martinkrulltott: This is much cleaner and easier to read, great improvement 👍

martinkrulltott (approved):

PR #1609, opened 2024-01-08
fix(interpretations): render loading spinner on interpretation/comment actions without causing flashing

jenniferarnesen:
Fixes: https://dhis2.atlassian.net/browse/DHIS2-16400


jenniferarnesen:
Will not implement this

PR #1623, opened 2024-01-30
feat(pivot-table): truncate title and show full in tooltip (V21) (#1579)

HendrikThePendric:
* feat(pivot-table): truncate title and show full in tooltip

* fix(pivot-table): set padding on hoverable div instead of container th

* fix(pivot-table): correctly show tooltip after resizing the container

* chore: simplify code and remove redundant prop

This is a backport of https://github.com/dhis2/analytics/commit/c37ba2d1a187963b3b5aeee5b951bcf9c129dcd2

### DIFF
--- a/src/__demo__/PivotTable.stories.js
+++ b/src/__demo__/PivotTable.stories.js
@@ -1119,3 +1119,44 @@ storiesOf('PivotTable', module).add('DEGS', (_, { pivotTableOptions }) => {
         </div>
     )
 })
+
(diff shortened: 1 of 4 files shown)


HendrikThePendric:
There is no need to backport this fix to v21 because it is not used in any of our supported apps.

HendrikThePendric:
There is no need to backport this fix to v21 because it is not used in any of our supported apps.

PR #1672, opened 2024-06-04
feat: implement shared dashboard plugin wrapper

edoardo:
**Relates to https://github.com/dhis2/data-visualizer-app/pull/3082**
**Relates to https://github.com/dhis2/line-listing-app/pull/396**
**Relates to https://github.com/dhis2/maps-app/pull/3232**

---

### Key features

1. implement a shared dashboard plugin wrapper component

---

### Description

This wrapper adds support for dashboard offline caching.
Before the logic was replicated in the plugin component of each analytics app.
Now all apps can use this wrapper and only implement their specific logic related to the props required to be passed down to the plugin rendering component.

---

### Known issues

~The~ Storybook ~build command fails~ has been updated in a [separate PR](https://github.com/dhis2/analytics/pull/1724) and works.

It used to run automatically after the `build` command, so also in Github, which I'm not sure is necessary.
It has been removed in the PR just to be able to build and copy the artifact on `d2-ci` so it can be used in app's PRs.

I think the issue is related to the upgrade of `cli-app-scripts` which introduces Webpack 5, while the Storybook setup is for Webpack 4.

I'm not sure an upgrade of Storybook belongs to this PR, but after this PR is merged, Storybook is broken.


HendrikThePendric:
Since I don't think we actually use the storybook build anywhere, I think it is fine to not include a fix for it in this PR. We should probably create a new JIRA ticket with a slightly larger scope: both building the storybook, and then also publishing it somewhere (perhaps a netlify deploy?)

src/components/DashboardPluginWrapper/DashboardPluginWrapper.js
```diff
@@ -0,0 +1,100 @@
...
+            startRecording({ onError: console.error })
+        } else if (!isParentCached && isCached) {
+            // Synchronize cache state on load or prop update
+            // -- a back-up to imperative `removeCachedData`
+            remove()
+        }
+
+        // NB: Adding `startRecording` to dependencies causes
+        // an infinite recording loop as-is (probably need to memoize it)
+        // eslint-disable-next-line react-hooks/exhaustive-deps
```
HendrikThePendric: I've left [a comment](https://dhis2.slack.com/archives/CPGUFVC56/p1720780637349409) in the platform-team Slack channel about this. This is a bug in the `useCachebleSection` hook as I see it.

src/components/DashboardPluginWrapper/DashboardPluginWrapper.js
```diff
@@ -0,0 +1,100 @@
...
+    isParentCached,
+    ...props
+}) => {
+    const { pwaEnabled } = useConfig()
+
+    useEffect(() => {
+        // Get & send PWA installation status now
+        getPWAInstallationStatus({
+            onStateChange: onInstallationStatusChange,
+        }).then(onInstallationStatusChange)
```
HendrikThePendric, line 60: I am a bit confused about these lines. I don't know why we have to pass the same function both as a parameter/option and then also call it ourselves. at first glance it appears to me that the `.then()` part is redundant and should be done by `getPWAInstallationStatus`.

src/modules/getPWAInstallationStatus.js
```diff
@@ -0,0 +1,63 @@
...
+        return INSTALLATION_STATES.READY
+    }
+    // note that 'registration.waiting' is skipped - it implies there's an active one
+    if (registration.installing) {
+        handleInstallingWorker({
+            installingWorker: registration.installing,
+            onStateChange,
+        })
+        return INSTALLATION_STATES.INSTALLING
+    }
```
HendrikThePendric, line 45: See my previous comment about the code in the useEffect hook. I guess you could say that this parts computes the initial state and I think we can add some calls to `onStateChange` here to avoid having to call `then` later on.

HendrikThePendric (approved):
I've left some comments and am not sure if we should perhaps not merge this PR until that bug in the `useCacheableSection` hook is solved. But will approve for logistical reasons.

src/components/Toolbar/UpdateButton.js
```diff
@@ -1,7 +1,7 @@
-import { CircularLoader } from '@dhis2-ui/loader'
 import i18n from '@dhis2/d2-i18n'
 import { colors } from '@dhis2/ui-constants'
 import { IconSync16 } from '@dhis2/ui-icons'
+import { CircularLoader } from '@dhis2-ui/loader'
```
jenniferarnesen: Noticed that the import for CircularLoader is coming from a different dependency than lines 2 and 3: @dhis2-ui/loader vs @dhis2/ui-icons. Should they be the same?

KaiVandivier: These should all be from the root package: `@dhis2/ui`

edoardo: Changed all the imports to use `@dhis2/ui` directly.

src/modules/getPWAInstallationStatus.js
edoardo: @KaiVandivier do you have any reply to these comments?
I believe was you that wrote this part originally.

src/components/DashboardPluginWrapper/DashboardPluginWrapper.js
KaiVandivier: I wasn't actually able to reproduce the infinite loop with this setup, but I reproduced it elsewhere and published a fix on `@dhis2/app-runtime` @ 3.10.6

src/components/DashboardPluginWrapper/DashboardPluginWrapper.js
KaiVandivier: I did it this way to communicate that you can definitely expect a value "now" (-ish; the service worker registration API is asynchronous), and you _may_ get another one or two updates later, so you have to account for potential future updates incoming -- maybe that didn't end up being obvious though

The current design also has two small benefits that I can see, though they're currently not taken advantage of:
1. A different handler can be used for the initial state value received than for the subsequent updates
2. This structure can be easily `await`ed if you want to control the flow when rendering the plugin for the first time

Your suggested refactor in another comment would certainly work though, and you guys are welcome to implement it if you prefer 👍 it should be pretty simple: adding `onStateChange(<state>)` before any `return <state>`

src/components/DashboardPluginWrapper/DashboardPluginWrapper.js
edoardo: So I can bump `@dhis2/app-runtime` to 3.10.6 and then it should be safe to fix this.

edoardo: Fixed.

edoardo:
> Since I don't think we actually use the storybook build anywhere, I think it is fine to not include a fix for it in this PR. We should probably create a new JIRA ticket with a slightly larger scope: both building the storybook, and then also publishing it somewhere (perhaps a netlify deploy?)

I created [a task](https://dhis2.atlassian.net/browse/LIBS-662) for this so we (hopefully) don't forget.

This has been done and merged already, see PR #1724 .

src/components/DashboardPluginWrapper/DashboardPluginWrapper.js
jenniferarnesen: Thought I would just point out that the[ current plugin architecture](https://github.com/dhis2/data-visualizer-app/blob/master/src/PluginWrapper.js#L90) has this same implementation. @HendrikThePendric @KaiVandivier Can both of you have a look and resolve the conversations on this PR?

KaiVandivier: I still prefer the way `getPWAInstallationStatus` is factored now for the reasons above, so I think it's up to you analytics folks to decide if you're persuaded 🙂

HendrikThePendric: >  when rendering the plugin for the first time

OK, I see how this mechanism lets you differentiate between specific behaviour and generic behaviour. I'm persuaded.

src/modules/getPWAInstallationStatus.js
HendrikThePendric: The answer @KaiVandivier provided to my previous comments resolves this conversation as well as far as I am concerned.

HendrikThePendric (approved):

jenniferarnesen:
Superceded by https://github.com/dhis2/analytics/pull/1748 because of unfortunate dependency downgrades that were introduced due to regenerating the yarn.lock file.

PR #1676, opened 2024-06-10
chore: new github workflow that copies build of all non-prod branches to d2-ci

jenniferarnesen:
Slightly related to: https://dhis2.atlassian.net/browse/DHIS2-17300

Every branch that is not listed as one of the prod branches (from node-publish.yml) will now have it's build pushed to d2-ci for every pushed commit. We can then use these builds during development and testing.

Changes to README
- remove build badges for versions that are no longer supported
- add docs about how to use the build

New gh workflow file
- after building the lib, use yarn pack to create the correct structure of the packaged lib. Then unzip the result and push that to d2-ci, which can then be installed by the consuming apps.

### DIFF
--- a/.github/workflows/copy-build-to-d2-ci.yml
+++ b/.github/workflows/copy-build-to-d2-ci.yml
@@ -1,83 +1,42 @@
-name: Publish JavaScript Library
+name: Copy build to d2-ci
 
 on:
-  push:
-    branches-ignore:
-      - master
-  release:
-    types: [published]
+    push:
+        branches-ignore:
+            - master
 
 env:
-  GIT_AUTHOR_NAME: '@dhis2-bot'
-  GIT_AUTHOR_EMAIL: 'apps@dhis2.org'
-  GIT_COMMITTER_NAME: '@dhis2-bot'
-  GIT_COMMITTER_EMAIL: 'apps@dhis2.org'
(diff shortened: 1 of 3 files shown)


.github/workflows/copy-build-to-d2-ci.yml
```diff
@@ -0,0 +1,51 @@
...
+              with:
+                  name: output-build
+                  path: output.tgz
+                  retention-days: 2
+
+            - name: Copy package to d2-ci
+              uses: dhis2/deploy-build@master
+              with:
+                  build-dir: package
+                  github-token: ${{env.GH_TOKEN}}
```
HendrikThePendric, line 45: I can't tell from this what the path to this artefact would be, but it would be convenient if:
1. The path/URL only has one dynamic section, which is the branch-name
2. We can have one artefact per branch (this would be the case if we implement point 1)

README.md
```diff
@@ -28,22 +13,47 @@ The analytics library contains components and modules that are used in DHIS 2 an
...
+```
+
+2. point to a branch:
+
+```
+"dependencies": {
+        "@dhis2/analytics": "git+https://github.com/d2-ci/analytics.git#chore/some-chore",
+        ...
+}
+```
```
HendrikThePendric, line 56: OK, this illustrates that what I suggested in my previous commit is already implemented.

HendrikThePendric (approved):

edoardo (approved):

PR #1680, opened 2024-06-12
fix: set `dataLabels.padding` to `0` to avoid labels being cut off in pie charts

HendrikThePendric:
Implements [DHIS2-16505](https://dhis2.atlassian.net/browse/DHIS2-16505)

**Relates to a yet-to-be-created PR in DV**

---

### Key features

1. Prevent labels being cut off in pie charts

---

### Description

Setting the padding to zero fixes the reported issue. 

As far as I am aware, there is no way of knowing exactly when the label is going to get cut off. As such this is an all-or-nothing type of solution: we set `padding` to `0` for _all_ pie-charts, or we keep things the way they are. 

I have checked a few other pie-charts and vieports sizes and do not see any visual degaradations/changes due to this change, so I suggest we make this change.

---

### Reproduction path

1. Open Pie Chart _"Immunization: Fully Immunized by Facility Type"_ in the data-visualizer-app
2. Reduce the viewport height to ~375px
3. Observe that the bottom label is partially cut off

---

### Test the fix

1. Download [the attached custom DV bundle](https://github.com/user-attachments/files/15805725/data-visualizer-999.9.9-2024-06-12T15-12-27.zip) (version `999.9.9-2024-06-12T15-12-27`) with this fix included
2. Install it to a DHIS2 Core instance
3. Follow the reproduction path above and observe that the bottom label is now fully visible

---

### Screenshots

before

_Before_

after

_After_

[DHIS2-16505]: https://dhis2.atlassian.net/browse/DHIS2-16505?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


martinkrulltott (approved):

martinkrulltott:
Code approved. Needs thorough testing in both DV and Dashboards though before merging

HendrikThePendric:
Closing this PR because the fix didn't work reliably. It just happened to fix the issue for a particular chart at a particular height.

PR #1683, opened 2024-06-18
fix: apply legend to all numeric and boolean types (DHIS2-17611)

janhenrikoverland:
Implements [DHIS2-17611](https://dhis2.atlassian.net/browse/DHIS2-17611)

---

### Description

This PR does two things:

- Applies legends to all numeric and boolean values, not just value type `NUMBER`, to support event data items
- Renders booleans as numeric values (integers), ie `.0` is stripped so that the number of "Yes" is presented as `3` instead of `3.0`

---

### TODO

-   [ ] Manual testing
-   [x] Figure out the removed i18n

---

### Screenshots

Before
Screenshot from 2024-06-18 15-43-06

After
Screenshot from 2024-06-18 15-43-10

(The negative integer value is just out of bounds of the legend set)

[DHIS2-17611]: https://dhis2.atlassian.net/browse/DHIS2-17611?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/PivotTable/PivotTableValueCell.js
```diff
@@ -46,7 +49,8 @@ export const PivotTableValueCell = ({
     // TODO: Add support for 'INTEGER' type (requires server changes)
```
edoardo: Does this comment still applies?
`isNumericValueType` returns `true` for integer types.

i18n/en.pot
```diff
@@ -38,11 +38,6 @@ msgstr "Created {{time}} by {{author}}"
 msgid "Created {{time}}"
 msgstr "Created {{time}}"
 
-msgid "Viewed {{count}} times"
```
edoardo: I don't see where these strings were removed in this PR.
Was it perhaps done in another PR but the pot file was not updated?

edoardo (changes requested):
There is an extra `output.tgz` file which shouldn't end up in master.

i18n/en.pot
janhenrikoverland: Looks like the issue where i18n changes are not reflected in the pot files before yarn start or build have been run. So quite often the pot file is updated in the next PR. Not great.

edoardo (approved):

PR #1696, opened 2024-08-21
fix: compute subtotals/totals for boolean types (DHIS2-9155)

edoardo:
Implements [DHIS2-9155](https://dhis2.atlassian.net/browse/DHIS2-9155)

---

### Key features

1. compute subtotals and totals for all numeric and boolean types

---

### Description

In aggregate analytics, boolean types can be summed up.
Prior to this PR, only dimensions with  value type `NUMBER` were summed up when subtotals and/or totals options were enabled.
After this PR, all numeric and boolean types will be summed up.

---

### Screenshots

Before:
Screenshot 2024-08-21 at 10 54 41

After:
Screenshot 2024-08-20 at 14 42 33

[DHIS2-9155]: https://dhis2.atlassian.net/browse/DHIS2-9155?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/modules/pivotTable/PivotTableEngine.js
```diff
@@ -744,7 +749,15 @@ export class PivotTableEngine {
                 totalCell.valueType = currentValueType
             }
 
-            if (dxDimension?.valueType === VALUE_TYPE_NUMBER) {
+            // compute subtotals and totals for all numeric and boolean value types
```
edoardo, line 752: I had to keep the part above with the prev/current comparison because `valueType` can be of types other than numeric and boolean.
In that case, the total cell would not get the type, but always N/A if we simply check for numeric/boolean for assigning the last evaluated `valueType` in the loop and fallback to N/A otherwise.

janhenrikoverland (approved):

janhenrikoverland:
Approved when tests are updated.

PR #1698, opened 2024-09-02
feat: single value as a Highcharts instance

HendrikThePendric:
Implements [DHIS2-17928](https://dhis2.atlassian.net/browse/DHIS2-17928)

**Relates to https://dhis2.atlassian.net/browse/DHIS2-17722**

---

### Key features

1. Implement the Single Value visualization type as a Highcharts Chart instance

---

### Description

We want to deprecate the SVG to PNG/PDF endpoints and export visualisations in the browser instead. To do so we can leverage Highchart's `Chart.exportChartLocal()` method. However, since the Single Value visualisation type was a completely custom implementation, not built in Highcharts, this visualisation type was problematic. We considered two approaches:
1. To implement a custom client-side SVG to PNG/PDF export mechanism. 
2. To refactor/rewrite the Single Value visualization type as a Highcharts Chart instance.

We investigated both options and have summarised our findings in [this document](https://docs.google.com/document/d/1Z9PCJI4VZoBA26XWS12ow7vQm1e_4JSP7aX3tzrTwgY/edit?usp=sharing). The final decision was to refactor the Single Value as a Highcharts Chart, which is what has been done in the current PR.

Apart from the Single Value refactor, this PR introduces some mechanisms to implement future custom visualizations using Highcharts as well.

Note that while this PR adds 1692 lines, this includes a story of 823 lines and that the current state of affairs still includes the old implementation. Removing the old code would remove ~970 lines. So when everything is done we'll actually be reducing the codebase size by about 100 lines (1692-823-970=-101).

---

### TODO / Discussion points

-   [x] Review/discuss comments in PR
-   [x] Test in DV and fix issues
-   [x] Tweak styles (see [a5821c0](https://github.com/dhis2/analytics/pull/1698/commits/a5821c0e91d60d5bfb4fff7c948a1c4d1d4cfac7))
-   [x] Remove `src/visualizations/config/adapters/dhis_dhis`
-   [x] Remove `src/visualizations/config/generators/dhis`
-   [x] Remove `src/visualizations/store/adapters/dhis_dhis`
-   [x] Update story, to remove reference to old implementation

Note that the last 4 points can easily be completed as follows:
1. If needed, rebase branch `chore/remove-old-single-value-code` onto this branch
2. Merge https://github.com/dhis2/analytics/pull/1703

---

[DHIS2-17928]: https://dhis2.atlassian.net/browse/DHIS2-17928?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.storybook/preview-head.html
HendrikThePendric, line 1: This ensures the Roboto font is pre-loaded to the storybook iframe. This was necessary for the Highcharts to compute the dimensions/position of text elements in the SVG correctly. Possibly we may need to do something similar in the data-visualizer-app.

src/visualizations/config/adapters/dhis_highcharts/chart.js
```diff
@@ -31,17 +34,30 @@ const getEvents = () => ({
...
-        getEvents()
+        extraOptions.dashboard ? DASHBOARD_CHART : undefined,
+        getEvents(),
+        layout.type === VIS_TYPE_SINGLE_VALUE
+            ? {
+                  backgroundColor: getSingleValueBackgroundColor(
+                      extraOptions.legendOptions,
+                      extraOptions.legendSets,
+                      series[0]
+                  ),
```
HendrikThePendric: In Highcharts it is possible to define a background color for a chart. As such, I thought it needed to be set here, instead of being updated later by the custom visualisation code.

src/visualizations/config/adapters/dhis_highcharts/chart.js
```diff
@@ -31,17 +34,30 @@ const getEvents = () => ({
...
+              }
+            : undefined,
+        layout.type === VIS_TYPE_SINGLE_VALUE && extraOptions.dashboard
+            ? { spacingTop: 7 }
+            : undefined
```
HendrikThePendric, line 61: The same goes for the `spacingTop`: this is a chart config option.

src/visualizations/config/adapters/dhis_highcharts/customSVGOptions/index.js
HendrikThePendric, line 1: The way this file is structured would allow us to easily add new custom visualization types.

src/visualizations/config/adapters/dhis_highcharts/customSVGOptions/singleValue/renderer/addIconElement.js
HendrikThePendric, line 1: This part was a bit tricky, because the `SVGRenderer` did not work with nested SVG elements. So instead of adding an SVG to the chart, I add a group (`<g>`) to which I add all the SVG children.

src/visualizations/config/adapters/dhis_highcharts/customSVGOptions/singleValue/renderer/constants.js
```diff
@@ -0,0 +1,4 @@
+// multiply value text size with this factor
+// to get very close to the actual number height
+// as numbers don't go below the baseline like e.g. "j" and "g"
+export const ACTUAL_NUMBER_HEIGHT_FACTOR = 2 / 3
```
HendrikThePendric: This is the only "magic value" that had to be ported over from the original implementation. The native SVG `<text>` element does have quite excessive padding, for just showing numbers.

src/visualizations/config/adapters/dhis_highcharts/customSVGOptions/singleValue/renderer/positionElements.js
HendrikThePendric, line 1: Note how positioning is being done using the Highchart `SVGElement.align` method.

.storybook/preview-head.html
HendrikThePendric, line 1: In order for Highcharts to correctly compute text element (`<text>`) dimensions, the font that was used needed to be pre-loaded. Possibly we may need to do something similar in the data-visualizer-app.

src/visualizations/config/adapters/dhis_highcharts/customSVGOptions/index.js
HendrikThePendric, line 1: This file was structured in a way that allows us to easily add more custom visualizations using the Highcharts `SVGRenderer` in the future.

src/visualizations/config/adapters/dhis_highcharts/customSVGOptions/singleValue/renderer/addIconElement.js
HendrikThePendric, line 1: This part was a bit tricky, because `Chart. exportChartLocal()` would crash when it encountered nested SVGs. So now I add the icon-SVG children to a group and add a `rect` to ensure the correct dimensions are present.

src/visualizations/config/adapters/dhis_highcharts/customSVGOptions/singleValue/renderer/positionElements.js
HendrikThePendric, line 1: Note that I could use Highchart's `SVGElement.align()` method here.

src/visualizations/config/adapters/dhis_highcharts/customSVGOptions/singleValue/renderer/positionElements.js
```diff
@@ -0,0 +1,62 @@
...
+        const iconHeight = height * scale
+        const valueElementHeight =
+            valueElementBox.height * ACTUAL_NUMBER_HEIGHT_FACTOR
+        const translateY = layoutRect.y + (valueElementHeight - iconHeight) / 2
+
+        /* The icon is a <g> with <path> elements that contain coordinates.
+         * These path-coordinates only scale correctly when using CSS translate */
+        iconElement.css({
+            transform: `translate(${translateX}px, ${translateY}px) scale(${scale})`,
+        })
```
HendrikThePendric: Since the icon needs to be both scaled and translated, I was not able to use the `.align()` method here.

src/visualizations/config/adapters/dhis_highcharts/customSVGOptions/singleValue/renderer/renderSingleValueSVG.js
```diff
@@ -0,0 +1,50 @@
...
+        )
+    }
+
+    positionElements.call(
+        this,
+        valueElement,
+        subTextElement,
+        iconElement,
+        styles.spacing
+    )
```
HendrikThePendric: Throughout this PR I call functions in this way. This is done so that the `this` scope of the called function is the Highcharts Chart instance. If this pattern is too obscure we can consider passing more parameters around.

src/visualizations/config/adapters/dhis_highcharts/customSVGOptions/singleValue/renderer/styles.js
```diff
@@ -0,0 +1,83 @@
...
+    { valueTop: 8, subTextTop: 4, iconGap: 4, iconSize: 164 },
+    { valueTop: 8, subTextTop: 4, iconGap: 4, iconSize: 146 },
+    { valueTop: 8, subTextTop: 4, iconGap: 4, iconSize: 128 },
+    { valueTop: 8, subTextTop: 4, iconGap: 4, iconSize: 110 },
+    { valueTop: 8, subTextTop: 4, iconGap: 4, iconSize: 92 },
+    { valueTop: 8, subTextTop: 4, iconGap: 4, iconSize: 74 },
+    { valueTop: 8, subTextTop: 4, iconGap: 4, iconSize: 56 },
+    { valueTop: 8, subTextTop: 4, iconGap: 4, iconSize: 38 },
+    { valueTop: 8, subTextTop: 4, iconGap: 4, iconSize: 20 },
+]
```
HendrikThePendric: Here we can adjust the styles for the various sizes

src/visualizations/config/adapters/dhis_highcharts/index.js
```diff
@@ -139,6 +136,19 @@ export default function ({ store, layout, el, extraConfig, extraOptions }) {
...
+         * under the Chart instance's `userOptions` member. This means
+         * that in event callback functions the custom SVG options are
+         * accessible as `this.userOptions.customSVGOptions` */
+        customSVGOptions: getCustomSVGOptions({
+            extraConfig,
+            layout: _layout,
+            extraOptions: _extraOptions,
+            metaData: store.data[0].metaData,
+            series,
+        }),
```
HendrikThePendric, line 147: Here I have invented a convention: custom SVG's need custom options and they can be populated here.

src/visualizations/config/adapters/dhis_highcharts/subtitle/index.js
HendrikThePendric, line 1: I'm not quite sure how I feel about the current state of this file. My initial goal was to reduce the amount of changes I introduced to a minimum, but I actually ended up changing quite a bit. At the same time, I feel like the current implementation is quite messy because instead of just producing all `subtitle` properties for a given viz-type (or the default) in one place, it's a bit of a mix: some things are done using if statements while other things are done in separate files.

src/visualizations/config/adapters/dhis_highcharts/title/index.js
HendrikThePendric, line 1: This file has the same problem as the subtitle

src/visualizations/config/adapters/dhis_highcharts/customSVGOptions/singleValue/renderer/renderSingleValueSVG.js
```diff
@@ -0,0 +1,53 @@
...
+
+        fitsWithinContainer = checkIfFitsWithinContainer(
+            getAvailableSpace.call(this, styles.spacing.valueTop),
+            valueElement,
+            subTextElement,
+            icon,
+            subText,
+            styles.spacing
+        )
+    }
```
HendrikThePendric: This part of the code is probably the most essential part of this PR. The mechanism is simple:
1. We start by rendering the required elements in the biggest size
2. We then check if it fits
3. And repeat the process until it does

So the scaling is not dynamic, it happens in steps. The reason for this is that this way it is more likely that multiple dashboard SV items will end up having the same font size.

src/visualizations/config/adapters/dhis_highcharts/customSVGOptions/singleValue/config/shouldUseContrastColor.js
edoardo, line 1: Is this only relevant for SV?
Is the idea to eventually move this one level up if we ever need to add a new custom SVG type?

src/visualizations/config/adapters/dhis_highcharts/customSVGOptions/singleValue/renderer/addIconElement.js
```diff
@@ -0,0 +1,64 @@
...
+     * 1. The SVG children are all <path> elements and they are not nested
+     * 2. Only the `d`, `fill-rule`, `clip-rule` and `fill` attributes are used */
+    Array.from(svgIconDocument.documentElement.children).forEach((pathNode) => {
+        /* Highcharts expects an array of letters and numbers but the
+         * d-attribute string only puts spaces in between adjacent number,
+         * not between letter and numbers. */
+        const d = pathNode
+            .getAttribute('d')
+            // Add space after letter when followed by number
+            .replace(/([A-Z])([0-9])/g, '$1 $2')
```
edoardo: Are we sure letters are always uppercase?

src/visualizations/config/adapters/dhis_highcharts/exporting.js
```diff
@@ -0,0 +1,25 @@
...
+export default function getExporting(visType) {
+    const exporting = {
+        // disable exporting context menu
+        enabled: false,
+    }
+    switch (visType) {
+        case VIS_TYPE_SINGLE_VALUE:
+            return {
+                ...exporting,
+                chartOptions: {
```
edoardo, line 13: Isn't this returned already in `chart.js`?
Here `getExporting` for SV is returning other things than the config for exporting...

src/visualizations/config/generators/highcharts/index.js
```diff
@@ -3,16 +3,24 @@ import HM from 'highcharts/highcharts-more'
...
 HNDTD(H)
 HE(H)
+HOE(H)
 HPF(H)
 HB(H)
+PEBFP(H)
+
+/* Whitelist some additional SVG attributes here. Without this,
+ * the PDF export for the SingleValue visualization breaks. */
+H.AST.allowedAttributes.push('fill-rule', 'clip-rule')
```
edoardo, line 23: This is another potential breaking point as we play with Highcharts internals.
It happened before that after updating Highcharts stuff broke because they changed something internally.
Ideally we shouldn't have code like this, it's hard to predict a problem even if we read carefully the changelog where changes to internal stuff is not necessarily visible there.

In some cases we don't have an alternative solution tho.
We should just keep in mind these parts whenever we update Highcharts, and test.

src/visualizations/config/adapters/dhis_highcharts/chart.js
```diff
@@ -31,17 +34,30 @@ const getEvents = () => ({
...
     return Object.assign(
         {},
         getType(layout.type),
         { renderTo: el || layout.el },
         DEFAULT_CHART,
-        dashboard ? DASHBOARD_CHART : undefined,
-        getEvents()
+        extraOptions.dashboard ? DASHBOARD_CHART : undefined,
+        getEvents(layout.type),
+        layout.type === VIS_TYPE_SINGLE_VALUE
```
edoardo: These will eventually need to be moved to a separate file when a new custom SVG type comes along...

src/visualizations/config/adapters/dhis_highcharts/exporting.js
HendrikThePendric: Code was refactored in [159727c](https://github.com/dhis2/analytics/pull/1698/commits/159727c42ea6c6c999b7a66f4280752f79e6b9dc) to address this

src/visualizations/config/adapters/dhis_highcharts/customSVGOptions/singleValue/config/shouldUseContrastColor.js
HendrikThePendric: Currently yes, it's only for SV, but it's a very generic helper so I've moved it in [e8f6a88](https://github.com/dhis2/analytics/pull/1698/commits/e8f6a880d432fd5655c32de052101e6ce73f0f03)

src/visualizations/config/adapters/dhis_highcharts/customSVGOptions/singleValue/renderer/addIconElement.js
HendrikThePendric: Yes. We discussed this part during the review call. We decided to just the simpler native DOM implementation instead of this error-prone approach. See [489ea08](https://github.com/dhis2/analytics/pull/1698/commits/489ea087c766b7e0ac736359373d4b11033a3548)

src/visualizations/config/adapters/dhis_highcharts/chart.js
HendrikThePendric: Code was refactored in [159727c](https://github.com/dhis2/analytics/pull/1698/commits/159727c42ea6c6c999b7a66f4280752f79e6b9dc) to address this

src/visualizations/config/adapters/dhis_highcharts/events/index.js
edoardo, line 1: This rename is a bit confusing.
I guess it should have been `chart.js -> chart/default.js`.

edoardo (approved):

PR #1567, opened 2023-08-25
feat: implement cumulative values in PT engine (DHIS2-5497)

edoardo:
Implements [DHIS2-5497](https://dhis2.atlassian.net/browse/DHIS2-5497)

**Relates to https://github.com/dhis2/data-visualizer-app/pull/2746**

**Published in [26.2.0-cumulative-values-alpha.1](https://www.npmjs.com/package/@dhis2/analytics/v/26.2.0-cumulative-values-alpha.1)**

---

### Key features

1. accumulate values horizontally when `cumulativeValues` option is checked
2. add example in Storybook

---

### Description

Similarly to the option for other vis types, values in a pivot table can be accumulated horizontally.
This behaviour is controlled by the `cumulativeValues` option in DV.

---

### TODO

-   [x] address XXX comments after code review

---

### Screenshots

Data before accumulating the values horizontally:
Screenshot 2023-11-13 at 14 42 24

Same data with accumulation:
Screenshot 2023-11-13 at 14 43 48

[DHIS2-5497]: https://dhis2.atlassian.net/browse/DHIS2-5497?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1747, opened 2025-01-24
fix: pivot table crashes when setting hide empty columns

jenniferarnesen:
Fixes [DHIS2-17876](https://dhis2.atlassian.net/browse/DHIS2-17876)

Check that data[row] exists before accessing it.

I think this is where the error was introduced: https://github.com/dhis2/analytics/pull/1567/files#diff-cecda9e65f89fb79253f45505c2d7587509d804de707bfd659ee82d97234daa9R573

[DHIS2-17876]: https://dhis2.atlassian.net/browse/DHIS2-17876?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/modules/pivotTable/PivotTableEngine.js
+++ b/src/modules/pivotTable/PivotTableEngine.js
@@ -607,7 +607,9 @@ export class PivotTableEngine {
         return !this.data[row] || this.data[row].length === 0
     }
     columnIsEmpty(column) {
-        return !this.rowMap.some((row) => this.data[row][column])
+        return !this.rowMap.some(
+            (row) => this.data[row] && this.data[row][column]
+        )
     }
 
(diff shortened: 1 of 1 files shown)


amcgee (approved):
LGTM

An alternative would be to use optional chaining:

```
!!(this.data[row]?.[column])
```

But this should work just as well.  I checked that `data[row][column]` should always be either undefined or an object/array, so we shouldn't run into issues with false-ish 0 or "".  Would be good to refactor this to typescript though so that's actually guaranteed.

edoardo (approved):

PR #1768, opened 2025-03-27
fix: ensure app link are formatted correctly (DHIS2-19270)

edoardo:
Implements [DHIS2-19270](https://dhis2.atlassian.net/browse/DHIS2-19270)

---

### Key features

1. ensure app links to visualizations are correctly formatted and point to the right place

---

### Description

From 42 the `baseUrl` returned by the `useConfig` hook is always an absolute URL.
This caused the code in the `GetLinkDialog` component to return the wrong URL with missing path after the hostname.

---

### TODO

- [x] Tests added
- [ ] PRs for all affected apps created

[DHIS2-19270]: https://dhis2.atlassian.net/browse/DHIS2-19270?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/FileMenu/__tests__/GetLinkDialog.spec.js
+++ b/src/components/FileMenu/__tests__/GetLinkDialog.spec.js
@@ -2,17 +2,57 @@ import { Button, Modal } from '@dhis2/ui'
 import { shallow } from 'enzyme'
 import React from 'react'
 import { GetLinkDialog } from '../GetLinkDialog.js'
-import { appPathFor } from '../utils.js'
 
-const testBaseUrl = 'http://test.tld/test'
+const testBaseUrl = 'http://host.tld/test/'
+
+const mockUseConfig = jest.fn(() => ({ apiVersion: 42, baseUrl: testBaseUrl }))
 
 jest.mock('@dhis2/app-runtime', () => ({
-    useConfig: () => ({ baseUrl: testBaseUrl }),
+    useConfig: () => mockUseConfig(),
 }))
+
+const tests = [
+    {
+        type: 'visualization',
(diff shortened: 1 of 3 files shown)


src/components/FileMenu/__tests__/GetLinkDialog.spec.js
```diff
@@ -25,27 +65,45 @@ describe('The FileMenu - GetLinkDialog component', () => {
...
+            getGetLinkDialogComponent({
+                type: tests[0].type,
+                id: tests[0].id,
+            }).find(Modal)
+        ).toHaveLength(1)
     })
 
-    it('renders a <a> tag containing the type and id props', () => {
-        const href = getGetLinkDialogComponent(props).find('a').prop('href')
+    tests.forEach((test) => {
```
HendrikThePendric: While this is fine, I just wanted to point out that Jest also has `test.each()` / `it.each()` for [these types of test scenarios](https://jestjs.io/docs/api#testeachtablename-fn-timeout). You could do something like this
```js
test.each(tests)(
    'renders an <a> tag with correct path and id for type $type',
    ({ type, baseUrl, id, expected}) => {
        // your test code
    }
)
```

HendrikThePendric (approved):

src/components/FileMenu/__tests__/GetLinkDialog.spec.js
edoardo: Refactored.

edoardo:
Superseded by #1770

PR #1771, opened 2025-04-08
fix: return new name and desc from file menu rename instead of making api request [DHIS2-19433]

jenniferarnesen:
Fixes [DHIS2-19433](https://dhis2.atlassian.net/browse/DHIS2-19433)

Instead of using patch, which causes several backend errors, the rename dialog
now just returns the user's changes to name and description, and leaves it
to the app to save the changes.

This is a workaround until problems with patch are fixed.

App PRs:
[DV](https://github.com/dhis2/data-visualizer-app/pull/3377)
[Maps](https://github.com/dhis2/maps-app/pull/3514)
[LL](https://github.com/dhis2/line-listing-app/pull/672)

---

### TODO

- [x] Tests added
- [ ] Storybook added N/A
- [x] Testing sign-off (checked by the tester, not developer)

---

[DHIS2-19433]: https://dhis2.atlassian.net/browse/DHIS2-19433?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/FileMenu/RenameDialog.js
```diff
@@ -117,14 +72,11 @@ export const RenameDialog = ({ type, object, onClose, onRename, onError }) => {
 }
 
 RenameDialog.propTypes = {
-    id: PropTypes.string,
     object: PropTypes.shape({
-        id: PropTypes.string.isRequired,
```
edoardo, line 122: I guess this `id` is still always passed but not necessary anymore.

jenniferarnesen: True. Will update.

edoardo (approved):

src/components/FileMenu/RenameDialog.js
```diff
@@ -10,58 +9,18 @@ import {
     TextAreaField,
 } from '@dhis2/ui'
 import PropTypes from 'prop-types'
-import React, { useMemo, useState } from 'react'
+import React, { useState } from 'react'
 import i18n from '../../locales/index.js'
 import { modalStyles } from './FileMenu.styles.js'
-import {
-    supportedFileTypes,
-    endpointFromFileType,
```
edoardo, line 18: Should we remove this entirely from the codebase?
It's rather ugly 😁 
Requires a change in `DeleteDialog`.

edoardo (approved):

src/components/FileMenu/utils.js
```diff
@@ -45,11 +49,76 @@ export const appPathFor = (fileType, id, apiVersion) => {
...
+}) => {
     delete visualization.id
     delete visualization.created
     delete visualization.createdBy
     delete visualization.user
 
+    visualization.name =
+        name ||
+        visualization.name ||
+        i18n.t('Untitled {{visualizationType}}, {{date}}', {
```
jenniferarnesen, line 65: Given the issues filtering with comma, should we remove that from the default name?

src/components/FileMenu/RenameDialog.js
jenniferarnesen: I agree that should be fixed to use the new function, but this increases the scope. I'll add it to an analytics refactor task.

src/components/FileMenu/utils.js
edoardo: The problem is not just the comma, apparently the space is used as token separator, so any `Untitled {{visualizationType}}` would match regardless of the date.

edoardo (approved):

src/components/FileMenu/utils.js
```diff
@@ -45,11 +49,76 @@ export const appPathFor = (fileType, id, apiVersion) => {
...
-export const preparePayloadForSaveAs = ({ ...visualization }) => {
+export const preparePayloadForSaveAs = ({
+    visualization,
+    name,
+    description,
+}) => {
     delete visualization.id
     delete visualization.created
     delete visualization.createdBy
     delete visualization.user
```
edoardo, line 61: Not really part of the scope but I think we should also delete subscribers when copying:
```suggestion
delete visualization.subscribers
```

PR #1772, opened 2025-04-09
fix: print the name of relative periods instead of actual periods (DHIS2-19454)

janhenrikoverland:
Implements [DHIS2-19454](https://dhis2.atlassian.net/browse/DHIS2-19454)

Objective: As a filter, all relative periods are printed with their name instead of the actual periods.

### Screenshots

Before: All actual periods are printed.

After:

Screenshot from 2025-04-09 14-49-24
Screenshot from 2025-04-09 14-49-17
Screenshot from 2025-04-09 14-49-07
Screenshot from 2025-04-09 14-48-42
Screenshot from 2025-04-09 14-48-23
Screenshot from 2025-04-09 14-48-00
Screenshot from 2025-04-09 14-47-28

[DHIS2-19454]: https://dhis2.atlassian.net/browse/DHIS2-19454?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/visualizations/util/getFilterText.js
+++ b/src/visualizations/util/getFilterText.js
@@ -31,11 +31,12 @@ export default function (filters, metaData) {
         ) {
             titleFragments.push(getOuLevelAndGroupText(filter, metaData))
         } else if (dimensionIs(filter, DIMENSION_ID_PERIOD)) {
+            const relativePeriodNames = getRelativePeriodsName()
             titleFragments.push(
                 dimensionGetItemIds(filter)
                     .map(
                         (id) =>
-                            getRelativePeriodsName()[id] ||
+                            relativePeriodNames[id] ||
                             metaData.items[id]?.name ||
                             id
                     )
--- a/src/visualizations/util/getFilterText.js
+++ b/src/visualizations/util/getFilterText.js
@@ -30,10 +30,7 @@ export default function (filters, metaData, extraOptions = {}) {
             )
         ) {
             titleFragments.push(getOuLevelAndGroupText(filter, metaData))
-        } else if (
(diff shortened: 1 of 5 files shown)


src/visualizations/util/getFilterText.js
```diff
@@ -25,6 +30,20 @@ export default function (filters, metaData) {
...
             titleFragments.push(getOuLevelAndGroupText(filter, metaData))
+        } else if (
+            dimensionIs(filter, DIMENSION_ID_PERIOD) &&
+            extraOptions.dashboard
+        ) {
+            titleFragments.push(
+                dimensionGetItemIds(filter)
+                    .map(
+                        (id) =>
+                            getRelativePeriodsName()[id] ||
```
edoardo: Perhaps assign the returned value of this function outside the loop and lookup by id here?

edoardo (commented):
We should adjust the tests in `__tests__/getFilterText.spec.js`.
Currently it says it expects already that the relative periods are not expanded, but the test seems to assert something else?
We should also add a case for dashboard, since the behaviour is different.

src/visualizations/util/getFilterText.js
janhenrikoverland: There can only be a single `pe` dimension, it will only execute once, so decided to skip the outside const as it's only relevant in this if-block.

janhenrikoverland:
Removed the dashboard check.

edoardo (approved):

PR #1780, opened 2025-05-21
fix: do not call getTrimmedConfig if hideEmptyRowItems is not available (DHIS2-19611)

edoardo:
Part of the fix for [DHIS2-19611](https://dhis2.atlassian.net/browse/DHIS2-19611)

**Relates to https://github.com/dhis2/data-visualizer-app/pull/3396**

---

### Key features

1. avoid crash in SV when `hideEmptyRowItems` option is set while in another vis type

---

### Description

Single Value is not using the same visualization API as all other Highcharts visualization types.
`getTrimmedConfig` was called if `hideEmptyRowItems` was not passed, which is the case with the changes in linked DV PR where we only pass relevant options for each vis type to the visualization API.
`getTrimmedConfig` expects the series option to be populated, which is not the case for SV.

---

### TODO

- [ ] Tests added N/A
- [ ] Storybook added N/A

---

### Screenshots

Before:
Screenshot 2025-05-21 at 12 05 46

After:
Screenshot 2025-05-21 at 13 11 05

[DHIS2-19611]: https://dhis2.atlassian.net/browse/DHIS2-19611?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/visualizations/config/adapters/dhis_highcharts/index.js
```diff
@@ -170,7 +170,7 @@ export default function ({ store, layout, el, extraConfig, extraOptions }) {
     }
 
     // hide empty categories
-    if (_layout.hideEmptyRowItems !== 'NONE') {
+    if (_layout.hideEmptyRowItems && _layout.hideEmptyRowItems !== 'NONE') {
```
janhenrikoverland: Consider exporting the valid `hideEmptyRowItems` options and check if this value is actually one of them?

edoardo: 🤔 this would not work if we want to only pass the relevant options by type.

janhenrikoverland: This if-check checks for ALL, BEFORE_FIRST, AFTER_LAST, BEFORE_FIRST_AFTER_LAST? So could we make them consts in this repo, then use them here in this check and also here https://github.com/dhis2/analytics/blob/ef877bec63ea3066dbc8b3538f61af409375b726/src/visualizations/config/adapters/dhis_highcharts/getTrimmedConfig.js#L74-L93 ?

edoardo: `hideEmptyRowItems` is not always passed in `_layout` regardless of default/set values.
Is this what you are suggesting?
```suggestion
    if ([ALL, BEFORE_FIRST, AFTER_LAST, BEFORE_FIRST_AFTER_LAST].includes(_layout.hideEmptyRowItems)) {
```

janhenrikoverland: Yes

edoardo: Refactored

janhenrikoverland (approved):

PR #1785, opened 2025-06-17
chore: support react 18

KaiVandivier:
Relevant to [DHIS2-18794](https://dhis2.atlassian.net/browse/DHIS2-18794)

1. Allows React 18 as a peer dependency, in addition to v16
2. Adds React 18 as a dev dependency, so it's used locally
3. Upgrades `@dhis2/cli-app-scripts` to avoid duplicate React dependencies

It also addresses an issue that happened to the file menu, which broke with React 18 changes -- this is what seems to be the cause:

> [Other Breaking Changes:](https://react.dev/blog/2022/03/08/react-18-upgrade-guide#other-breaking-changes)
**Consistent useEffect timing:** React now always synchronously flushes effect functions if the update was triggered during a discrete user input event such as a click or a keydown event. Previously, the behavior wasn’t always predictable or consistent.

With React 18, it looks like a single click on the Menu item triggers:
1. The `onClick` handler (in `HoverMenuBar.js`), which sets `isInHoverMode` to true
1. And then the `useEffect` that depends on `isInHoverMode` triggers (synchronously, as part of the same event it seems)
    1. Event listeners are added to the document
    1. And then somehow those trigger immediately, I think with the same event that triggered the onClick on the Menu item
    1. ...and that closes the menu

[DHIS2-18794]: https://dhis2.atlassian.net/browse/DHIS2-18794?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


package.json
```diff
@@ -46,6 +46,9 @@
         "fs-extra": "^10.1.0",
         "jest-enzyme": "^7.0.2",
         "prop-types": "^15",
+        "react": "^18.3.1",
+        "react-dom": "^18.3.1",
+        "react-scripts": "^5.0.1",
```
KaiVandivier, line 52: `react-scripts` lets storybook work as-is

HendrikThePendric:
I think https://github.com/dhis2/analytics/pull/1785/commits/36cdeb65bc6af99f1fe36e5879e1e73adb790e37 is a slightly more elegant way to address the problem, but it did require a bit of refactoring, so we'll need to do some additional testing. The main premise was this:
- Deferring the call to `document.addEventListener` is not a good approach because it could cause race conditions.
- What we want to do is stop the current click event from bubbling up to the document so it doesn't end up calling the callback we're attaching to it
- To do so we must attach the event listener where we handle the click that opens the menu, which is the `onDropdownMenuClick`. There we have access to the click event so we can call `event.stopPropagation()`
- But we need this function to be stable and adding the `lastHoveredSubMenuEl` state to the dependency array of `onDropdownMenuClick` would make it (very) unstable
- And since we don't actually need this `lastHoveredSubMenuEl` to be a "reactive state" we can actually just store it in a ref. 

So I actually think that apart from addressing the problem, this is actually a good general improvement to the component. It should now re-render less often and is "side-effect-free",

KaiVandivier:
Nice! I think that's the fix this actually needs 🙌 thanks @HendrikThePendric!

src/modules/utils.js
```diff
@@ -24,3 +24,17 @@ export const useDidUpdateEffect = (fn, inputs) => {
...
     }, inputs)
 }
+
+/**
+ * Check if a value is numeric
+ *
+ * @param param Value to be checked
+ * @returns {boolean} Returns true when the `param` is a numeric value
+ */
+export function isNumeric(param) {
```
jenniferarnesen: Remember to add unit tests :)

jenniferarnesen (dismissed):

jenniferarnesen (approved):

PR #1787, opened 2025-07-24
chore: convert tests to use testing-library

edoardo:
Implements [DHIS2-19918](https://dhis2.atlassian.net/browse/DHIS2-19918)

---

### Key features

1. rewrite tests with `testing-library`
2. remove `enzyme`

---

### Description

Similarly to what has been done in the other repos, we want to move away from `enzyme` for unit testing.
The PR is converting all the existing tests that use `enzyme` to `testing-library`.
Some tests had to be adjusted slightly to accommodate for the different testing approach.

---

### TODO

- [ ] Tests added N/A
- [ ] PRs for all affected apps created N/A
- [ ] Storybook added N/A

[DHIS2-19918]: https://dhis2.atlassian.net/browse/DHIS2-19918?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/DimensionsPanel/List/DimensionList.js
```diff
@@ -64,7 +65,7 @@ class DimensionList extends Component {
         )
 
         return (
-            <div className="container">
+            <div className="container" data-test="dimension-list">
```
jenniferarnesen: Looks like this data-test is already in use:  https://github.com/dhis2/analytics/blob/master/src/components/DataDimension/Calculation/DataElementSelector.js#L284

Maybe the one in the calcuation component should actually be renamed?  Not sure...

src/components/DimensionsPanel/List/__tests__/DimensionItem.spec.js
```diff
@@ -1,51 +1,45 @@
...
-    it('matches the snapshot with locked', () => {
-        props.isLocked = true
-        expect(dimensionItem()).toMatchSnapshot()
-    })
-
-    it('matches the snapshot with onOptionsClick', () => {
-        props.onOptionsClick = jest.fn()
-        expect(dimensionItem()).toMatchSnapshot()
-    })
+const props = {
```
jenniferarnesen, line 5: I think it would be cleaner to have a distinct props const in each test. As it is now, each test affects the next (i.e., isRecommended is set to true in the second test, and remains true for the following tests)

src/components/DimensionsPanel/__tests__/DimensionsPanel.spec.js
```diff
@@ -1,43 +1,56 @@
...
+    dimensions: [
+        { id: 'id1', name: 'Dimension 1' },
+        { id: 'id2', name: 'Dimension 2' },
+    ],
+}
+
+test('DimensionsPanel matches the snapshot', () => {
+    const { container } = render(<DimensionsPanel {...props} />)
+
+    expect(container).toMatchSnapshot()
```
jenniferarnesen: I think maybe just remove this test since the snapshot has all the jsx classes from the ui library (e.g., "jsx-1141087211"). This will cause us to sometimes need to regenerate snapshot files when we upgrade ui. Also, I think the next test more or less covers things (you could add the expectation of "Filter dimensions" on line 42 to the next test)

src/components/DimensionsPanel/__tests__/__snapshots__/DimensionsPanel.spec.js.snap
```diff
@@ -0,0 +1,164 @@
...
+exports[`DimensionsPanel matches the snapshot 1`] = `
+<div>
+  <div
+    style="height: 100%; display: flex; flex-direction: column; background-color: rgb(255, 255, 255); padding: 8px; overflow: hidden;"
+  >
+    <div
+      class="container"
+    >
+      <div
+        class="jsx-1141087211 "
```
jenniferarnesen: Imho we should avoid snapshot testing when ui components end up in the snapshot because these classes will sometimes get new ids. An alternative would be to mock the ui components.

src/components/FileMenu/__tests__/GetLinkDialog.spec.js
```diff
@@ -51,29 +51,23 @@ const tests = [
...
+        type: tests[0].type,
+        id: tests[0].id,
+        onClose,
     }
 
-    beforeEach(() => {
-        shallowGetLinkDialog = undefined
-    })
+    test('renders a Modal component', () => {
+        render(<GetLinkDialog {...props} />)
```
jenniferarnesen, line 64: This file is a case where we could un-share the props and make it simpler. For example, in this test we could just do:
```<GetLinkDialog id="myvisualizationid" type="visualization" />```

File: src/components/FileMenu/__tests__/GetLinkDialog.spec.js [hunk 6]
```diff
@@ -84,28 +78,26 @@ describe('The FileMenu - GetLinkDialog component', () => {
...
-                id,
-                onClose,
-            })
-                .find('a')
-                .prop('href')
+            props.type = type
+            props.id = id
 
-            expect(href).toMatch(expected)
+            render(<GetLinkDialog {...props} />)
```
jenniferarnesen: And here we could just do:
```<GetLinkDialog id={id} type={type} />```

File: src/components/FileMenu/__tests__/GetLinkDialog.spec.js [hunk 7]
```diff
@@ -84,28 +78,26 @@ describe('The FileMenu - GetLinkDialog component', () => {
...
-        })
-            .find(Button)
-            .at(1)
-            .simulate('click')
-
-        expect(onClose).toHaveBeenCalled()
+    test('calls the onClose callback when the Close button is clicked', async () => {
+        const user = userEvent.setup()
+
+        render(<GetLinkDialog {...props} />)
```
jenniferarnesen:
```
const onClose = jest.fn()
<GetLinkDialog id="myvisualizationid" type="visualization" onClose={onClose} />
```

src/components/OrgUnitDimension/__tests__/OrgUnitDimension.spec.js
```diff
@@ -1,57 +1,90 @@
...
+        const { container } = renderOrgUnitDimension(props)
+
+        // avoid the act warning due to the snapshot being taken before async code is run
+        // wait for the component to be loaded, here done by testing that the OrganisationUnitTree component is loaded
+        await screen.findByText('Org unit tree component mock')
 
-    it('matches the snapshot', () => {
-        const wrapper = getWrapper()
-        expect(wrapper).toMatchSnapshot()
+        expect(container).toMatchSnapshot()
```
jenniferarnesen: This could potentially cause us headaches with the jsx classes, as mentioned above. We can leave it in for now and if we have to update the snapshots then consider using mocking, or otherwise checking for existing components rather than the snapshot.

src/components/PeriodDimension/__tests__/FixedPeriodSingleSelect.spec.js
```diff
@@ -1,29 +1,14 @@
...
-        shallowFixedPeriodSelect = undefined
-    })
-
-    it('matches the snapshot', () => {
-        const wrapper = getWrapper()
+    const { container } = render(<FixedPeriodSelect {...props} />)
 
-        expect(wrapper).toMatchSnapshot()
-    })
+    expect(container).toMatchSnapshot()
```
jenniferarnesen: Same snapshot problem. Here I would consider creating mocks for  SingleSelectField, InputField, SingleSelectOption from ui. See here for an example: 

https://github.com/dhis2/maps-app/blob/master/src/components/layers/toolbar/__tests__/LayerToolbar.spec.js#L6

src/components/PeriodDimension/__tests__/PeriodDimension.spec.js
```diff
@@ -9,29 +9,14 @@ jest.mock('@dhis2/app-runtime', () => ({
...
-        shallowPeriodDimension = undefined
-    })
-
-    it('matches the snapshot', () => {
-        const wrapper = getWrapper()
+    const { container } = render(<PeriodDimension {...props} />)
 
-        expect(wrapper).toMatchSnapshot()
-    })
+    expect(container).toMatchSnapshot()
```
jenniferarnesen: Same problem as above. The previously existing snapshot looks almost worthless. So we could either remove this test, or else check for the existence of elements in the document. (e.g. Relative Periods, Fixed Periods buttons

src/components/PeriodDimension/__tests__/PeriodSelector.spec.js
```diff
@@ -1,31 +1,16 @@
...
-        shallowPeriodTransfer = undefined
-    })
-
-    it('matches the snapshot', () => {
-        const wrapper = getWrapper()
+    const { container } = render(<PeriodTransfer {...props} />)
 
-        expect(wrapper).toMatchSnapshot()
-    })
+    expect(container).toMatchSnapshot()
```
jenniferarnesen: Same problem. Not sure mocking is a great idea to fix this one since the ui is producing a nested component. Maybe just look for the tabs to be in the document.

src/components/RichText/Parser/__tests__/Parser.spec.js
```diff
@@ -9,35 +9,32 @@ jest.mock('../MdParser.js', () => ({
...
 
-        expect(richTextParser).toHaveLength(1)
+        expect(divEl.style.color).toBe(defaultProps.style.color)
+        expect(divEl.style.whiteSpace).toBe(defaultProps.style.whiteSpace)
     })
 
-    it('should have rendered a result with the style prop', () => {
-        richTextParser = renderComponent(defaultProps, 'test prop')
+    test('should have rendered content', () => {
+        renderComponent({}, 'plain text')
```
jenniferarnesen: The renderComponent abstraction is a one-liner, so it doesn't seem worth it.

jenniferarnesen (changes requested):
See comments

src/components/DimensionsPanel/List/DimensionList.js
edoardo: We don't use a consistent format.
The `ui` format prefixes the name with `dhis2-ui` and in analytics I see many places we do something similar with `dhis2-analytics`, but other places don't have the prefix.
I changed this specific one to avoid using the same in different places.

src/components/DimensionsPanel/List/__tests__/DimensionItem.spec.js
edoardo: Fixed.
I kept the "shared" props object but instead of changing values there I pass the specific options directly in the component.

src/components/DimensionsPanel/__tests__/DimensionsPanel.spec.js
edoardo: Removed the snapshot test.

src/components/FileMenu/__tests__/GetLinkDialog.spec.js
edoardo: Refactored to not use the shared props object.

src/components/PeriodDimension/__tests__/FixedPeriodSingleSelect.spec.js
edoardo: Refactored this one to not rely on snapshot and without mocking.
I just look for some text in the rendered component and check that the selected values are correct.

src/components/PeriodDimension/__tests__/PeriodDimension.spec.js
edoardo: Refactored to get rid of the snapshot testing.

src/components/PeriodDimension/__tests__/PeriodSelector.spec.js
edoardo: I decided to remove this one since I basically added tests for the `PeriodDimension` component which ultimately just renders `PeriodTransfer`.

src/components/RichText/Parser/__tests__/Parser.spec.js
edoardo: Removed.

src/components/DimensionsPanel/__tests__/__snapshots__/DimensionsPanel.spec.js.snap
edoardo: Refactored to not rely on the snapshot.

jenniferarnesen (approved):

PR #1788, opened 2025-08-01
feat: event analytics response transformers (DHIS2-19818)

janhenrikoverland:
Implements https://dhis2.atlassian.net/browse/DHIS2-19818

- [x] Collect event values where needed
- [x] Prefix event values where needed
- [x] Produce prefixed metadata items, metadata dimensions and rows
- [x] Ensure that option sets with non-unique option codes always work
- [x] Display names for booleans instead of 0 and 1
- [x] Support "No value" dimension items for all value types and respect `hideNaData` option
- [x] Unit tests for all functions
- [x] Storybook examples for all supported types

**Supported value types**
- [x] Boolean: BOOLEAN
- [x] Coordinate: COORDINATE
- [x] Date: AGE, DATE (formatted)
- [x] Date & time: DATETIME (formatted)
- [x] Legend sets
- [x] Numeric: INTEGER, INTEGER_NEGATIVE, INTEGER_POSITIVE, INTEGER_ZERO_OR_POSITIVE, NUMBER, PERCENTAGE (formatted), UNIT_INTERVAL
- [x] Option sets
- [x] Organisation unit: ORGANISATION_UNIT
- [x] Text: EMAIL, LETTER, LONG_TEXT, PHONE_NUMBER, TEXT, URL, USERNAME 
- [x] Time: TIME
- [x] Yes only: TRUE_ONLY

**Unsupported value types**
- FILE_RESOURCE
- GEOJSON
- IMAGE
- MULTI_TEXT
- REFERENCE
- TRACKER_ASSOCIATE

**Storybook examples**

Boolean
Screenshot from 2025-09-02 11-20-03
Screenshot from 2025-09-02 11-20-06

Date
Screenshot from 2025-09-02 11-20-09
Screenshot from 2025-09-02 11-20-13

Datetime
Screenshot from 2025-09-02 11-20-16
Screenshot from 2025-09-02 11-20-19

Numeric
Screenshot from 2025-09-02 11-20-22
Screenshot from 2025-09-02 11-20-26

Numeric with legend set
Screenshot from 2025-09-02 11-20-29
Screenshot from 2025-09-02 11-20-32

Option sets with non-unique codes
Screenshot from 2025-09-02 11-20-34
Screenshot from 2025-09-02 11-20-37

Text
Screenshot from 2025-09-02 11-20-40
Screenshot from 2025-09-02 11-20-43

Time
Screenshot from 2025-09-03 10-46-37
Screenshot from 2025-09-03 10-46-44

Yes only
Screenshot from 2025-09-02 11-20-45
Screenshot from 2025-09-02 11-20-48


src/modules/response/response.js
```diff
@@ -0,0 +1,128 @@
...
+                ![DIMENSION_ID_PERIOD, DIMENSION_ID_ORGUNIT].includes(
+                    header.name
+                )
+        )
+
+    // Legendsets use uids and do not need transformation
+    // Skip unsupported value types
+    // Option set and Boolean have separate handlers
+    // All other types use default handler with specific item formatter
+    metaHeaders.forEach((header) => {
```
edoardo: We could use `for ... of` and `continue` here in order to remove the nested `if` and reduce cognitive complexity.

src/modules/response/response.js
```diff
@@ -0,0 +1,128 @@
...
+                            itemFormatterByValueType[header.valueType],
+                    }
+                )
+            }
+        }
+    })
+
+    // Add "No value" dimension item if "Hide NA data" option is disabled
+    // Only add if there is at least one empty value
+    if (!hideNaData) {
```
edoardo: Could this be integrated in the same loop above to avoid iterating over `metaHeaders` again?

src/modules/response/response.js
edoardo: If we were to combine the 2 loops on `metaHeaders`, the first `if` here would be moved in the `else` at the end.

src/modules/response/__tests__/boolean.spec.js
```diff
@@ -0,0 +1,27 @@
+import booleanResponseHideNa from '../../../__demo__/data/event/boolean.data.hidena.json'
+import booleanResponseOrg from '../../../__demo__/data/event/boolean.data.org.json'
+import yesOnlyResponseHideNa from '../../../__demo__/data/event/yesonly.data.hidena.json'
+import yesOnlyResponseOrg from '../../../__demo__/data/event/yesonly.data.org.json'
+import { applyBooleanHandler } from '../boolean.js'
+
+const headerIndex = 0
+
+describe('boolean', () => {
```
edoardo: @jenniferarnesen you have some opinions about how to structure these tests. I think all the nested `describe` can be simplified.

src/modules/response/boolean.js
```diff
@@ -0,0 +1,23 @@
+import i18n from '@dhis2/d2-i18n'
+import { applyDefaultHandler, getPrefixedValue } from './default.js'
+
+export const applyBooleanHandler = (response, headerIndex) => {
+    const dimensionId = response.headers[headerIndex].name
+    const defaultResponse = applyDefaultHandler(response, headerIndex)
```
edoardo: I'm confused about this.
Wouldn't it potentially override the `itemFormatter` [previously applied](https://github.com/dhis2/analytics/pull/1788/files#diff-c2996315871e34453c4f85cd433f682ca5b27041e75c097b612d4c353a9b1201R101)?

`response` is already the transformed response, shouldn't this function only override the boolean item identified by the `headerIndex`?

edoardo: Ok, I think I got it now, but it's still a bit confusing because without knowing what `defaultResponse` is at this point it might seem that some `metaData` and/or `items` can loose their transformed values.

edoardo (commented):
Added some comments about potential improvements.
Otherwise it looks good, except perhaps for the boolean "transformer" which confused me initially.

src/modules/response/response.js
```diff
@@ -0,0 +1,128 @@
...
+
+export const itemFormatterByValueType = {
+    [VALUE_TYPE_AGE]: (name) => name.replace(/ 00:00:00\.0$/, ''),
+    [VALUE_TYPE_DATETIME]: (name) => name.replace(/:00\.0$/, ''),
+    [VALUE_TYPE_DATE]: (name) => name.replace(/ 00:00:00\.0$/, ''),
+    [VALUE_TYPE_PERCENTAGE]: (name) =>
+        name.endsWith('.0') ? name.slice(0, -2) : name,
+}
+
+export const transformResponse = (response, { hideNaData = false } = {}) => {
```
edoardo: Does this apply to all responses (aggregate and events)?
If not perhaps rename the function to indicate it's events specific.

src/modules/response/__tests__/default.spec.js
edoardo, line 1: Should this have a test for boolean?

src/modules/response/event/__tests__/response.spec.js
```diff
@@ -22,9 +22,28 @@ import responseTimeOrg from '../../../../__demo__/data/event/time.data.org.json'
 import responseYesOnlyHideNa from '../../../../__demo__/data/event/yesonly.data.hidena.json'
 import responseYesOnly from '../../../../__demo__/data/event/yesonly.data.json'
 import responseYesOnlyOrg from '../../../../__demo__/data/event/yesonly.data.org.json'
-import { transformResponse } from '../response.js'
+import { getItemFormatterByValueType, transformResponse } from '../response.js'
 
 describe('response', () => {
+    describe('getItemFormatterByValueType', () => {
+        it('should return the correct formatter and format correctly', () => {
+            expect(getItemFormatterByValueType('BOOLEAN')('1')).toBe('Yes')
```
edoardo: Should these use the constants from `valueTypes`? 😁

edoardo (dismissed):

src/modules/response/event/__tests__/response.spec.js
janhenrikoverland: Yes

edoardo (approved):

PR #1791, opened 2025-09-04
feat: add InterpretationsProvider and refactor all components to use it

HendrikThePendric:
Implements [DHIS2-19822](https://dhis2.atlassian.net/browse/DHIS2-19822)

**Relates to**
- https://github.com/dhis2/line-listing-app/pull/716

---

### Description

The new `InterpretationsProvider` and its hooks keep track of the internal state. As a result, the components have been simplified significantly, they are now mostly presentational. Because the state is now managed in a Provider have been able to remove "prop-drilling" in general and passing "on-complete-handlers" down the component tree.

And since all the business logic is now encapsulated in the hooks it was possible to create a comprehensive test suite for these which actually covers virtually all of the important stuff.

---

### BREAKING CHANGE

Consumers must implement the provider, or the interpretations modal and unit will break.

---

### TODO

- [x] Tests added
- [x] Storybook updated

---

### Known issues

- ~I could potentially also address the date formatting issues in this PR. But not 100% what to address.~

---

[DHIS2-19822]: https://dhis2.atlassian.net/browse/DHIS2-19822?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/Interpretations/InterpretationsProvider/InterpretationsManager.js
```diff
@@ -0,0 +1,307 @@
...
+                    ],
+                    filter: `${this.currentType}.id:eq:${this.currentVisualizationId}`,
+                    paging: false,
+                },
+            },
+        })
+        return result.interpretations.interpretations
+    }
+
+    async loadActiveInterpretation(id) {
```
edoardo, line 153:
```suggestion
    async setActiveInterpretation(id) {
```
`load` is a bit ambiguous, I think `set` or `store` is more clear.

src/components/Interpretations/InterpretationsProvider/InterpretationsManager.js
```diff
@@ -0,0 +1,307 @@
...
+    async createInterpretation({ type, id, text, onComplete }) {
+        await this.mutate(
+            {
+                resource: `interpretations/${type}/${id}`,
+                type: 'create',
+                data: text,
+            },
+            { onComplete }
+        )
+        /* since the create request does not rerun an ID we must refetch the list
```
edoardo:
```suggestion
        /* since the create request does not return an ID we must refetch the list
```
I guess you meant `return` here.

src/components/Interpretations/InterpretationsProvider/InterpretationsManager.js
```diff
@@ -0,0 +1,307 @@
...
+            ...interpretation,
+            likedBy: isLikedByCurrentUser
+                ? interpretation.likedBy.concat({
+                      id: this.currentUser.id,
+                  })
+                : interpretation.likedBy.filter(
+                      (lb) => lb.id !== this.currentUser.id
+                  ),
+            likes: isLikedByCurrentUser
+                ? interpretation.likes + 1
```
edoardo, line 243: I understand why we are doing this and I suppose we were doing something similar before.
The potential issue is that you get an out of sync counter, but I guess it's not that important, it will be fixed whenever the interpretations panel is re-opened.

src/components/Interpretations/InterpretationsProvider/InterpretationsManager.js
```diff
@@ -0,0 +1,307 @@
...
+        this.interpretations.set(id, interpretation)
+        this.notifyInterpretationObservers(id)
+        return interpretation
+    }
+
+    async deleteCommentFromActiveInterpretation({ id }) {
+        const activeInterpretation = this.getActiveInterpretation()
+        await this.mutate({
+            resource: 'interpretations',
+            id: `${activeInterpretation.id}/comments/${id}`,
```
edoardo: Any reason for not putting the whole string in `resource` as done above?
This `id` parameter containing a path is a bit weird.
If `id` is required for the `delete` mutation I would put the comment id there and move the rest of the path in `resource`.

src/components/Interpretations/InterpretationModal/InterpretationModal.js
```diff
@@ -39,82 +42,30 @@ function getModalContentCSS(width) {
     return css.resolve`
         div {
             width: ${width}px;
-            overflow-y: visible;
```
edoardo, line 42: @HendrikThePendric can you look into this one?

HendrikThePendric: As discussed, this should be reverted and the fix needs to be implemented in the EVER app

edoardo: Change reverted.

src/components/Interpretations/InterpretationsProvider/InterpretationsManager.js
HendrikThePendric: Yeah I agree with your assesment. There is a potential of getting out-of-sync, but actually I double checked how this code is used, and because this function is exposed to the component via the `useLike` hook, which has an "in-porgress" state, which disables the button, I think the problem is mostly theoritical. Also, as you mention, any potential issues would fix themself when the panel is re-opened, or another event happens that refetches things. Doing things this way avoids requesting the entire interpretation, but changing the code to do a `query` after the `mutate` would be fairly trivial. I personally suggest we keep things as they are now.

edoardo: Agreed.

src/components/Interpretations/InterpretationsProvider/InterpretationsManager.js
HendrikThePendric: I've just moved everything to the resource. I agree having this whole path in the ID was weird.

src/__demo__/InterpretationsUnit.stories.js
```diff
@@ -37,15 +41,21 @@ export const WithNoTimeDimensionsWarning = () => {
...
                     name: 'Tom Wakiki',
                 }}
-                id="abcd"
-                onReplyIconClick={Function.prototype}
-                type="eventVisualization"
-                visualizationHasTimeDimension={false}
-            />
+            >
+                <InterpretationsUnit
+                    currentUser={{
```
edoardo: I think this should be removed.

edoardo (dismissed):

edoardo (dismissed):

edoardo (approved):

PR #1801, opened 2026-01-27
feat: use enabled period types (DHIS2-21001)

janhenrikoverland:
**Architecture**

I decided to modify the existing components in @dhis2/analytics to communicate with the server and toggle on server version. Apps like Data Visualizer can still use PeriodDimension and release with min version 2.40 without problems. For <=42 nothing has changed.

In 43+ the _enabled_ PTs are fetched from the server. These are used to populate the fixed period types. For relative periods it's a bit more complicated. The following considerations have been made:

- The "Financial year relative period start month" setting points to a FY period type. This PT must be enabled, otherwise hide FY period type. Settings app or the server might improve their handling and ensure valid combinations later. EDIT: Settings app will show warnings, but not block the actions, ref https://github.com/dhis2/settings-app/pull/1456.
- For the above, extra mapping was needed as the server (unfortunately) stores e.g. "FINANCIAL_YEAR_OCTOBER" instead of matching its own period type "FinancialOct".
- The corresponding settings for weekly and six-monthly are not implemented yet. So for now: if any weekly type is enable, show Weeks. If any six-monthly type is enabled, show Six-months.
- The remaining relative PTs have respective fixed PTs, so show based on their enabled status.
- The old "Hide*periods" settings are no longer regarded in the 43+ implementation
- The setting "Default relative period for analysis" (e.g. "LAST_12_MONTHS") but could be of any PT, so we need to parse and extract the PT and ensure that it's enabled. Otherwise pick a period from the closest enabled PT. EDIT: The fallback to the "closest" enabled period type code has been removed due to the Settings app PR mentioned above.

Claude Code suggested a few improvements:

**Data fetching fix**

The original code used useDataQuery({ skip: true }) to conditionally skip queries on pre-v43 servers. This doesn't work — useDataQuery has no skip mechanism on the query object, and useStaticInput freezes the first query reference. Replaced with a single combined query using { lazy: true } + useEffect, which is the established pattern in this codebase (see OpenFileDialog, AboutAOUnit, etc.). This also reduces three network requests to one.

**Period options filtering consolidated in PeriodTransfer**

FixedPeriodFilter and RelativePeriodFilter were computing their own filtered period lists internally, duplicating logic already in PeriodTransfer. Now PeriodTransfer always passes pre-filtered availableOptions to both filter components, making them pure presentation components. FixedPeriodSelect (the only other consumer of
FixedPeriodFilter) now computes availableOptions from its allowedPeriodTypes prop before passing them down.

**Tab toggle bug fix**

onIsRelativeClick was using the raw getRelativePeriodsOptionsById/getFixedPeriodsOptionsById helpers, bypassing the filtered options. On v43+ servers with restricted period types, switching between relative/fixed tabs could show periods from disabled types.

**Screenshots v42**

No changes:

Screenshot from 2026-01-27 12-24-15
Screenshot from 2026-01-27 12-24-25

**Screenshots v43**

Respects the enabled types:

Screenshot from 2026-01-27 12-25-27
Screenshot from 2026-01-27 12-25-02
Screenshot from 2026-01-27 12-25-10


src/components/PeriodDimension/utils/enabledPeriodTypes.js
```diff
@@ -0,0 +1,198 @@
...
+    BiWeekly: 'BIWEEKLY',
+    Monthly: 'MONTHLY',
+    BiMonthly: 'BIMONTHLY',
+    Quarterly: 'QUARTERLY',
+    QuarterlyNov: 'QUARTERLYNOV',
+    SixMonthly: 'SIXMONTHLY',
+    SixMonthlyApril: 'SIXMONTHLYAPR',
+    SixMonthlyNov: 'SIXMONTHLYNOV',
+    Yearly: 'YEARLY',
+    FinancialFeb: 'FYFEB',
```
edoardo, line 19: Are `FinancialJan` and `FinancialDec` also possible?

edoardo (dismissed):

src/components/PeriodDimension/utils/enabledPeriodTypes.js
janhenrikoverland: Nope

PR #1826, opened 2026-03-23
chore: config to satisfy node 24 + yarn upgrades

janhenrikoverland:
Spring cleaning:

- Upgrading to Node LTS
- yarn upgrade
- Fixed lint warnings

Screenshot_2026-03-23_11-22-10

### DIFF
--- a/.github/workflows/copy-build-to-d2-ci.yml
+++ b/.github/workflows/copy-build-to-d2-ci.yml
@@ -27,7 +27,7 @@ jobs:
             - name: Set up Node.js
               uses: actions/setup-node@v4
               with:
-                  node-version: '20.x'
+                  node-version: '24.x'
 
             - name: Install dependencies
               run: yarn install --frozen-lockfile
--- a/.github/workflows/copy-build-to-d2-ci.yml
+++ b/.github/workflows/copy-build-to-d2-ci.yml
@@ -27,7 +27,7 @@ jobs:
             - name: Set up Node.js
               uses: actions/setup-node@v4
               with:
-                  node-version: '24.x'
+                  node-version: 'lts/*'
 
             - name: Install dependencies
               run: yarn install --frozen-lockfile
--- a/config/setupJest.js
+++ b/config/setupJest.js
@@ -1,5 +1,6 @@
-if (global.CSS === undefined) {
-    global.CSS = { supports: () => false }
-} else if (typeof global.CSS.supports !== 'function') {
-    global.CSS.supports = () => false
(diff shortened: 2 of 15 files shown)


.github/workflows/copy-build-to-d2-ci.yml
```diff
@@ -27,7 +27,7 @@ jobs:
             - name: Set up Node.js
               uses: actions/setup-node@v4
               with:
-                  node-version: '20.x'
+                  node-version: '24.x'
```
edoardo:
```suggestion
                  node-version: 'lts/*'
```

Should we use this so we don't have to manually update the version?
In EVER app we have it this way.

config/setupJest.js
```diff
@@ -0,0 +1,6 @@
+/* eslint-disable no-undef */
+if (globalThis.CSS === undefined) {
+    globalThis.CSS = { supports: () => false }
```
edoardo: We have this in dashboard app but returning `true` to fix some Highcharts issues...

src/components/UserMention/useUserSearchResults.js
```diff
@@ -21,7 +21,7 @@ export const useUserSearchResults = ({ searchText }) => {
         lazy: true,
     })
 
-    const debouncedRefetch = useCallback(debounce(refetch, 250), [refetch])
+    const debouncedRefetch = useMemo(() => debounce(refetch, 250), [refetch])
```
edoardo, line 24: Does this fix some linting errors?

edoardo (commented):
Overall looks good.
I have some doubts about a few changes; I added comments on those.

src/components/UserMention/useUserSearchResults.js
janhenrikoverland: Yes, the third one here:

Screenshot_2026-03-23_11-22-10

config/setupJest.js
janhenrikoverland: Our tests pass either way as long as CSS.supports is callable and doesn't crash. But true would probably be good whenever a test actually tests the highcharts output as it would not fall back to some legacy rendering 👍

.github/workflows/copy-build-to-d2-ci.yml
janhenrikoverland: :+1:

config/setupJest.js
janhenrikoverland: Updated

edoardo (approved):

PR #1830, opened 2026-04-20
feat(interpretations): visual design and UX improvements

cooper-joe:
Implements [UX-198](https://dhis2.atlassian.net/browse/UX-198)

---

### Key features

1. Visual and UX refresh for the `Interpretation` and related components

---

### Description

This PR makes visual and ux-focused changes to the `Interpretation` and its related components. The changes remove some unnecessary elements. 

Summary of major changes to each component:

**Message**
- Removed card-style layout
- Improved typography and readability
- Adjusted date format
- Improve footer action buttons, removed 0 counts
- Added delete confirmation

**List**
- Removed date-grouped headers

**Thread/Modal**
- Removed visualization title from modal title. This was added when this modal was designed for use on Dashboards, but this functionality hasn't materialised. Repeating the visualization name in DV/Maps is poor use of space. If this functionality is needed in the future we can add it elsewhere.
- Adjusted content width calculation for the updated modal component, fixes spacing bugs.
- Improved thread rendering
- Moved interpretation date to the modal title.

**Rich text editor**
- Custom button usage, reduce the border-stacking poor visual appearance
- Improve visual design

---

Storybook stories were added to support testing, I kept these in as they seem useful.

---

### TODO

- [ ] PRs for all affected apps created
- [x] Storybook added

---

### Screenshots

image

image

image

image

image
image

[UX-198]: https://dhis2.atlassian.net/browse/UX-198?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Interpretations/common/Message/MessageIconButton.js
+++ b/src/components/Interpretations/common/Message/MessageIconButton.js
@@ -1,4 +1,4 @@
-import { Tooltip, colors, spacers, theme } from '@dhis2/ui'
+import { Tooltip, colors, theme } from '@dhis2/ui'
 import cx from 'classnames'
 import PropTypes from 'prop-types'
 import React from 'react'
@@ -8,7 +8,8 @@ const MessageIconButton = ({
     disabled,
     onClick,
     selected,
-    count,
+    confirming,
+    label,
     iconComponent: Icon,
     dataTest,
     viewOnly,
@@ -26,12 +27,18 @@ const MessageIconButton = ({
                         event.stopPropagation()
                         onClick()
                     }}
-                    className={cx('button', { selected, viewOnly })}
+                    className={cx('button', {
+                        selected,
+                        viewOnly,
+                        confirming,
+                    })}
                     disabled={disabled}
                     data-test={dataTest}
                 >
-                    {count && count}
                     <Icon />
+                    {label !== undefined && label !== null && (
+                        <span className="label">{label}</span>
+                    )}
                 </button>
                 <style jsx>{`
                     .wrapper {
@@ -43,11 +50,15 @@ const MessageIconButton = ({
                         cursor: pointer;
                         display: inline-flex;
                         flex-direction: row;
-                        gap: ${spacers.dp4};
+                        gap: 3px;
                         align-items: center;
-                        font-size: 12px;
-                        line-height: 14px;
+
+                        font-size: 13px;
+                        line-height: 1;
                         color: ${colors.grey700};
+                        border: 1px solid transparent;
+                        border-radius: 5px;
+                        padding: 3px 5px;
                     }
 
                     .viewOnly {
@@ -56,11 +67,13 @@ const MessageIconButton = ({
 
                     .button.selected {
                         color: ${colors.teal600};
-                        font-weight: 500;
+                        font-weight: 400;
+                        background-color: ${colors.teal050};
                     }
 
                     .button:hover {
                         color: ${colors.grey900};
+                        background-color: ${colors.grey200};
                     }
 
                     .button.selected:hover {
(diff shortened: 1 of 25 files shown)


HendrikThePendric (commented):
Most of the changes look good, but I have some doubts about this delete confirmation UX. Firstly, it has a small bug that can be triggered like so:
- Click delete button once
- Move mouse to the right so that the cursor is over the right end of the now-expanded button. You'll see the "Click again to delete" tooltip
- Wait until the timeout expires, now the cursor is no longer over the button. So this means there was a mouse-enter event but never a mouse-leave event and the tooltip remains visible

I think by applying different markup and keeping the delete button the same size you'd be able to avoid all of this.

From a UX perspective this button is quite nice, because it's quite subtle. But it's not a conventional solution (I think) and since the timeout is fairly short and because you don't see a counter, it's also a bit surprising/confusing. It'd be very nice if there would be some sort of visual clue to what is happening. Possibly you could have a circle with an animation or animate the fill of the background or sth?

cooper-joe:
> Most of the changes look good, but I have some doubts about this delete confirmation UX. Firstly, it has a small bug that can be triggered like so:
> 
>     * Click delete button once
> 
>     * Move mouse to the right so that the cursor is over the right end of the now-expanded button. You'll see the "Click again to delete" tooltip
> 
>     * Wait until the timeout expires, now the cursor is no longer over the button. So this means there was a mouse-enter event but never a mouse-leave event and the tooltip remains visible
> 
> 
> I think by applying different markup and keeping the delete button the same size you'd be able to avoid all of this.

Good catch on that bug. A markup change to avoid the issue completely sounds ideal. 

> From a UX perspective this button is quite nice, because it's quite subtle. But it's not a conventional solution (I think) and since the timeout is fairly short and because you don't see a counter, it's also a bit surprising/confusing. It'd be very nice if there would be some sort of visual clue to what is happening. Possibly you could have a circle with an animation or animate the fill of the background or sth?

Could you elaborate or reword this? I don't follow? Specifically, 

> because you don't see a counter, it's also a bit surprising/confusing. It'd be very nice if there would be some sort of visual clue to what is happening

What do you mean by _counter_? I do think a label appearing saying _Delete?_, plus the colour change, does bring the user's attention to the fact that "confirmation is required to perform that action". But I'm curious to know the solution you describe above.

src/components/Interpretations/common/Message/MessageIconButton.js
```diff
@@ -26,12 +27,18 @@ const MessageIconButton = ({
...
+                    })}
                     disabled={disabled}
                     data-test={dataTest}
                 >
-                    {count && count}
                     <Icon />
+                    {label !== undefined && label !== null && (
+                        <span className="label">{label}</span>
+                    )}
                 </button>
```
HendrikThePendric, line 42: I'm thinking that if you wrap `<Icon />` with the tooltip anchor span, then that anchor will have a constant size which will avoid the bug I reported earlier. Downside would be that if you hover the right end of the button now the tooltip will disappear, but I think this is less of an issue.

HendrikThePendric: I don't think a best-of-both-worlds solution is possible:
- If the anchor element changes size then the mouse can move out of the button without a mouseout event and the tooltips keeps showing. 
- If the additional content is outside the tooltip anchor, then hovering over the additional content will not keep the tooltip in view

HendrikThePendric:
I opened https://github.com/dhis2/analytics/pull/1831, a PR against this branch. It does the following:
1. The bin icon changes into a progress ring that lasts as long as the confirm timeout. This is an illustrated example of what I was trying to say in my review comment. The problem as I see it is that the button exits the confirmation state rather unexpectedly.  We need to communicate to the user that they need to click before the time runs out.
2. The tootlip anchor is now the icon so the bug is resolved. 

@cooper-joe let me know if you like it!

cooper-joe:
> The bin icon changes into a progress ring that lasts as long as the confirm timeout. This is an illustrated example of what I was trying to say in my review comment. The problem as I see it is that the button exits the confirmation state rather unexpectedly. We need to communicate to the user that they need to click before the time runs out.

@HendrikThePendric Visually, I think this is a nice addition. But I'm not sure it makes the interaction any clearer. To me, the progress ring could also be interpreted as "I need to wait until it deletes" or "it's processing the delete". It does a better job of informing the user "you have X seconds remaining to confirm", but it doesn't make the "Confirm?" step any clearer.

PR #1835, opened 2026-05-28
fix: lookup ID values in metadata items

HendrikThePendric:
Implements [DHIS2-21528](https://dhis2.atlassian.net/browse/DHIS2-21528)

**Relates to https://github.com/dhis2/event-visualizer-app/pull/240** this PR is technically unrelated but both bugs cause the app to show ID string instead of display names, so they are functionally related.

---

### Key features

1. Try to lookup ID values in `metaData.items`

---

### Description

This is a generic fix that assumes a text value could be an ID. If this is the case then the item name can be looked up in `metaData.items`. If the lookup fails then we assume it must be a genuine free text item.

#### Note about i18n changes

This PR contains significant changes to the POT file. Upon inspection, it turned out that even though these changes are unrelated to the current PR, these are in fact correct. So somehow the POT file on master was stale, which is a bit surprising because the i18n generation runs on pre-commit. I suppose if someone pushed to this repo with a broken git-hooks setup that could explain it.

---

### Known issues

- In theory it is possible that a text value is encountered which is supposed to be a regular "free text string" but it happens to exactly match the ID of one of the items in `metaData.items`. I think we can classify this as being a hypothetical case.

---

### Screenshots

_supporting text_

[DHIS2-21528]: https://dhis2.atlassian.net/browse/DHIS2-21528?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/modules/response/event/default.js
+++ b/src/modules/response/event/default.js
@@ -24,9 +24,9 @@ const resolveName = (value, itemFormatter, items) => {
     if (itemFormatter) {
         return itemFormatter(value)
     }
-    // Look up the value in metaData.items: id-valued columns (e.g. a
-    // CATEGORY data element) carry their label there keyed by the id.
-    // Free-text columns miss the lookup and fall back to the raw value.
+    /* Assume the value could be an ID, which means the name should
+     * be looked up in `metaData.items`. If that lookup fails the
+     * value is used directly. */
     return items?.[value]?.name ?? value
 }
 
--- a/src/modules/response/event/default.js
+++ b/src/modules/response/event/default.js
@@ -20,12 +20,22 @@ export const getValuesUniqueSortedAsc = (values, valueType = VALUE_TYPE_TEXT) =>
 export const getPrefixedValue = (value, prefix) =>
     `${prefix}${PREFIX_SEPARATOR}${value}`
 
-export const getItems = (values, dimensionId, { itemFormatter } = {}) =>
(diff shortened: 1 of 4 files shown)


src/modules/response/event/default.js
```diff
@@ -20,12 +20,24 @@ export const getValuesUniqueSortedAsc = (values, valueType = VALUE_TYPE_TEXT) =>
...
+const resolveName = (value, itemFormatter, items) => {
+    if (itemFormatter) {
+        return itemFormatter(value)
+    }
+    // For id-valued columns (e.g. a CATEGORY data element) the response
+    // already carries a proper label in metaData.items keyed by the id.
+    // Prefer it; for genuine free-text columns the lookup misses and we
+    // fall back to the raw value.
+    const existingName = items?.[value]?.name
+    return existingName && existingName !== value ? existingName : value
```
janhenrikoverland: Is `existingName !== value` actually doing anything here? If they are the same it doesn't matter which one is returned. So maybe just return existingName if it exists, otherwise value.

Suggestion: Remove 31 and 32, replace with:

`return items?.[value]?.name ?? value`

janhenrikoverland (dismissed):

janhenrikoverland (approved):

